PR #292, opened 2017-06-20
Fixes #287 - Avoid encoding JavaScript regular expressions in strings…

### LINKED ISSUE #287: [Proposal] Avoid encoding JavaScript regular expressions in strings
Encoding regular expressions as JavaScript strings is kind of awful because of the escaping rules. Maybe default to using glob patterns instead (perhaps use [minimatch](https://github.com/isaacs/minimatch))?

Instead of the following:

```json
"ignoredUrls": {
    ".*\\.domain1\\.com/.*": ["*"],
    "www.domain2.net": ["disallowed-headers"]
}
```

I propose the following:

```json
"ignoredUrls": {
    "*.domain1.com/**": ["*"],
    "www.domain2.net": ["disallowed-headers"]
}
```

In cases where `.sonarrc.js` is used, the developer could directly provide a JavaScript regular expression (assuming the regular expression is not a key in an object...as is the case above).

austinkelleher:
… in favor of glob patterns.

/cc @molant @patrick-steele-idem I'm putting up this PR to continue the conversation. After implementing this, I'm not convinced that it is the right solution. Glob patterns may not provide the flexibility necessary for this type of operation. Let me know your thoughts.

For example:

`*.domain1.com/**`
- Matches: `hello.domain1.com/world`
- Does not match: `http://www.domain1.com/world`

Let me know your thoughts, so we can decide whether to move forward with this approach, change the implementation, or simply close the PR and stick with the current approach.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "lockfile": "^1.0.3",
     "lodash": "^4.17.4",
     "mime-db": "^1.28.0",
+    "minimatch": "^3.0.4",
     "mkdirp": "^0.5.1",
     "node-ssllabs": "^0.5.0",
     "optionator": "^0.8.2",
--- a/src/lib/sonar.ts
+++ b/src/lib/sonar.ts
@@ -12,6 +12,7 @@ import * as url from 'url';
 import * as _ from 'lodash';
 import * as browserslist from 'browserslist';
 import * as chalk from 'chalk';
+import * as minimatch from 'minimatch';
 import { EventEmitter2 as EventEmitter } from 'eventemitter2';
 
 import { debug as d } from './utils/debug';
@@ -36,7 +37,7 @@ export class Sonar extends EventEmitter {
     private collectorConfig: object
     private messages: Array<IProblem>
     private browsersList: Array<String> = [];
-    private ignoredUrls: Map<string, RegExp[]>;
+    private ignoredUrls: Map<string, string[]>;
     private _formatter: string
 
     get pageDOM() {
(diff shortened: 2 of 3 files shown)


molant:
Maybe we are overcomplicating things and the common use case doesn't require that much flexibility?
Couple of edge cases:

* we can do something like `*\.js` and we will ignore all `js` files. This might be useful if we don't want to run any rule on JS, but then why not disable the rules related to that?
* We can ignore certain paths: `example.com/ignoredPath`? Why would I want to do that?

For the way I usually work I know I want to ignore domains: `domain1.com`, `sub.domain1.com` and I think that it covers everything. Maybe we just need a list of domains to ignore?

At the same time, removing features we already have might not be great.

austinkelleher:
Alternatively, we could support regex and glob pattern strings by checking the type that it's passed and handle it accordingly. We could also allow providing a function that would do the handling for complicated use cases. Perhaps something like:

Config:
```
{
    ...
    excludeUrl(url) {
        // Some really awesome logic here
        return true;
    }
}
```

sarvaje:
I'm with @molant, if we don't need so much flexibility maybe if enough if we support values like: `*.domain.com` and `sub.domain.com`

molant:
@austinkelleher I like the idea of providing a custom function and it seems to align well to what other tools do.
Now the question is:
* do we want more simple patterns + custom function
* or regex + glob pattern + custom function
?

molant:
This questions has been in the air for way too long.
My vote is:
* support regex, glob + custom function if @austinkelleher has the time to implement it
* regex and glob otherwise

@sonarwhal/contributors what do you think? I'd like for us to make a decission in the next  24h.

austinkelleher:
@molant My suggestion is to reject this PR and work on include/exclude instead of spending the time to improve this since it will be deprecated anyway.

molant:
@AustinKelleher you own this PR so if you'd rather do that go for it 😊

austinkelleher:
Closing this PR in favor of working on #343 instead as it will result in the deprecation of `ignoredUrls`.

PR #293, opened 2017-06-20
Improve requester

### LINKED ISSUE #260: Improvements to requester
There are a few things to improve in requester and when using `request` in general:
* [x] Limit the number of redirects (no limit at the moment)
* [x] Correctly implement a timeout. I had an issue with `favicon.ico`  not returning and the test failing. I forgot to add a `res.end()`, but the request to the favicon should had failed and the process continue normally.

qzhou1607:
Fix #260

* Limit the number of redirects.
* Implement timeout
* Tests

### DIFF
--- a/src/lib/collectors/utils/requester.ts
+++ b/src/lib/collectors/utils/requester.ts
@@ -31,7 +31,8 @@ const defaults = {
         'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Safari/537.36'
     },
     jar: true,
-    time: true
+    time: true,
+    timeout: 30000
 };
 
 export class Requester {
@@ -41,8 +42,14 @@ export class Requester {
     private _request: request;
     /** Internal `redirectManager`. */
     private _redirects = new RedirectManager();
+    /** Maximum number of redirects */
+    private _maxRedirects: number = 10;
 
-    constructor(customOptions = {}) {
+    constructor(customOptions?: { [key: string]: any }) {
+        if (customOptions) {
+            customOptions.followRedirect = false;
+            this._maxRedirects = customOptions.maxRedirects || this._maxRedirects;
+        }
         const options = Object.assign({}, defaults, customOptions);
 
         this._request = request.defaults(options);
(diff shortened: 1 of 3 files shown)


src/lib/collectors/utils/requester.ts
```diff
@@ -41,11 +42,14 @@ export class Requester {
     private _request: request;
     /** Internal `redirectManager`. */
     private _redirects = new RedirectManager();
+    /** Maximum number of redirects */
+    private _limit;
 
-    constructor(customOptions = {}) {
+    constructor(customOptions = {}, redirectsLimit = 3) {
```
molant: I think `redirectsLimit` should be part of `customOptions` and be named `maxRedirects` for consistency with the `request` module. Also the default value should be 10 to match `request` again.

src/lib/collectors/utils/requester.ts
```diff
@@ -31,7 +31,8 @@ const defaults = {
         'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Safari/537.36'
     },
     jar: true,
-    time: true
+    time: true,
+    timeout: 3000
```
molant: This value is way too small. Any specific reason why you are changing this value instead of relying in what the OS does?

src/lib/collectors/utils/requester.ts
molant: Also we should make sure that if the user sets `followRedirect: true` in `customOptions` we set it to `false`.

tests/lib/collectors/utils/requester.ts
```diff
@@ -17,7 +17,7 @@ const text = `This is a text
 
 test.beforeEach(async (t) => {
     const server = createServer();
-    const requester = new Requester();
+    const requester = new Requester({});
```
molant: Do we need to pass an empty object now? That doesn't look ok.

qzhou1607: @molant Done. I changed the requester constructor so we don't need the empty object.

src/lib/collectors/utils/requester.ts
```diff
@@ -81,6 +88,12 @@ export class Requester {
 
                     this._redirects.add(newUri, uri);
 
+                    const currentRedirectNumber = this._redirects.calculate(newUri).length;
+
+                    if (currentRedirectNumber > this._maxRedirects) {
+                        reject(`The number of redirects(${currentRedirectNumber}) exceeds the limit(${this._maxRedirects}).`);
```
austinkelleher: I think you should `return` here otherwise `resolve` or `reject` (again) could get called below.

molant: Good catch. Fixed in https://github.com/sonarwhal/sonar/commit/e8c79e8dadfc631459aa35ceb29a9ec30f782136

Issue #283, opened 2017-06-19
[Proposal] Improve rules configuration (include/exclude)


patrick-steele-idem:
Configuring rules is going to be one of the most common thing and I think there is room for improvement. It appears that the configuration structure was inspired by ESLint, but I feel like the following would be more intuitive:

```js
{
    "rules": [
        "rule1",        // Default to "error"
        "rule2:warn",   // Switch to warn
        "rule3:off",    // Quickly disable a rule
        ["rule4", {
            exclude: /.*\.foo\.com/.*/
        }],
        ["rule5", {
            include: [/.*\.domain1\.com/.*/],
            exclude: [/.*\.foo\.com/.*/]
        }],
        ["rule6", {
            include: ['*.domain1.com/**'] // Simple glob
        }],
        { // Additional rules for a specific domain
          include: ['*.foo.com/**'],
          rules: [
            'rule7',
            'rule8'
          ]
        }
    ],
    include: [ // Only run rules for the following URLs
      '*.ebay.com'
    ],
    exclude: [ // Don't run rules for any of the following URLs
      'tracking.ebay.com'
    ]
}
```

As shown above, each rule should support an include and exclude option. Value could either a be string (treated as a glob), a JS regular expression or an Array of either of those. The code above also shows how a global exclude/include could be supported.

austinkelleher:
@molant I can take this one

molant:
Go for it, I think you can assign yourself to this task.
I'm refactoring how we are loading the rules, collectors, etc. And I'm changing where the validation is done (`sonar.ts` instead of `cli.ts`) . I don't think this is going to impact you but just FYI.
Hopefully I'll have the PR done by the end of the day for review :)

austinkelleher:
@molant Okay thanks for the heads up. I cannot assign myself. I assume this has to do with the fact that the repo was moved to its own org.

molant:
@austinkelleher you should be now a contributor :)

austinkelleher:
Thanks!

PR #295, opened 2017-06-21
Change how sonar resources are loaded

### LINKED ISSUE #234: Test requiring rules, formatters, etc. defined in packages
We need to validate that we can use rules, formatters, etc. that are defined outside the core rules:

* Install sonar as if it were an npm package but without publishing it yet (@alrra has the right command)
* Create a new simple rule and do the same fake npm publish
* Make sure it runs in a project

molant:
The current order is:

1. Resources in core
1. Scoped packages (@sonarwhal/XXX)
1. Prefixed packages (sonar-resourceType-XXX)

The previous way of loading resources was weak and is better to rely
on `require` when dealing with external resources.

Some improvements of the current code over the previous:

* We load only the resources that we need instead of all, and on demand.
* `sonar.create()` is now sync and should load faster.
* Validation of rules is done now in `sonar` (when loading them individually
  and not the `cli`. This fixes a previously unkown issue on `rule-runner`
  where the schema of rules wasn't validated (it showed up in `ssllabs`.
* Removed all code related to `plugins` until we know how they look like.

Fix #234


molant:
There are a few things I still want to do but are relatively minor compare to the rest (and today is has been a really long day):

* [ ] now that `sonar.create` is sync, there is no point in having a `init` in there. Also that code is a bit messy and I'd like to refactor it a bit.
* [ ] need to add tests to actually validate that the order is core > @sonarwhal > prefixed
* [ ] also not sure about having `sonar-resourceType-resourceName`. Maybe just `sonar-resourceName`

If you can start reviewing it please do so. I'd like to merge this in the next 48h :)

austinkelleher:
👍  for making `sonar.create` sync. I thought it was a bit odd that it was synchronous. It looks like your PR steps on my changes for #283. In particular, I also removed the validation logic for the `cli` and let `sonar` be the one to do the validation. I think I'll wait for these changes to be more stable before I continue my work or perhaps I will branch off of your branch.

molant:
@austinkelleher things should be more stable with my latest push. I'm going to add some more tests and then address any feedback.

molant:
This is ready for feedback.

tests/lib/config/config-validator.ts
```diff
@@ -28,15 +28,15 @@ test('if config has an invalid schema, it should return false', (t) => {
     t.false(valid);
 });
 
-test(`if there is a rule in the config that doesn't exist, it should return false`, (t) => {
-    const config = _.cloneDeep(validConfig);
+// test(`if there is a rule in the config that doesn't exist, it should return false`, (t) => {
```
sarvaje: are we going to need it in the future? If so add some comment about when can we uncomment it, if not, remove the commented code

tests/lib/config/config-validator.ts
```diff
@@ -48,15 +48,15 @@ test(`if rule severity isn't valid, it should return false`, (t) => {
     t.false(valid);
 });
 
-test(`if rule schema isn't valid, it should return false`, (t) => {
-    const config = _.cloneDeep(validConfig);
+// test(`if rule schema isn't valid, it should return false`, (t) => {
```
sarvaje: are we going to need it in the future? If so add some comment about when can we uncomment it, if not, remove the commented code

src/lib/utils/resource-loader.ts
```diff
@@ -11,81 +11,148 @@
...
     formatter: 'formatter',
-    plugin: 'plugin',
     rule: 'rule'
 };
 
-/** Loads all the resources available for the given type */
-const loadOfType = (type: string): Map<string, Resource> => {
-    const PROJECT_ROOT = path.join(__dirname, '../../../../');
+const getCoreResources = (type: string): Array<string> => {
+    const resourcesFiles: string[] = globby.sync(`${PROJECT_ROOT}/dist/src/lib/${type}s/**/*.js`, { absolute: true });
```
sarvaje: I think it would be nice to have some kind of cache here instead of calculate every time the core resources

tests/lib/config/config-validator.ts
molant: I just moved this test to the `sonar.ts` ones because that's where we do the rule validation.

tests/lib/config/config-validator.ts
molant: I just moved this test to the `sonar.ts` ones because that's where we do the rule validation.

src/lib/utils/resource-loader.ts
molant: Done

tests/lib/ui/generator.ts
```diff
@@ -0,0 +1,150 @@
...
+const resourceLoader = {
+    getCoreCollectors() { },
+    getCoreFormatters() { },
+    getCoreRules() { },
+    loadRules() { }
+};
+
+const stubPifyMethodObject = { pifyMethod() { } };
+
+const stubPifyObject = {
```
qzhou1607: Since we don't use `pify` any more. I think we should rename this `stubUtilObject`.

src/lib/cli.ts
```diff
@@ -25,10 +25,10 @@ import { loadJSONFile } from './utils/misc';
 import * as logger from './utils/logging';
 import { cutString } from './utils/misc';
 import { options } from './ui/options';
+import { initSonarrc } from './ui/generator';
```
qzhou1607: It might be only me. We have both a `rule-generator` and a `generator`,  it seems a little confusing. Maybe rename the latter one as `sonarrc-generator`.

molant: I was thinking about moving `rule-generator` under `ui` as well. Maybe we should rename that folder to `cli` too.

tests/lib/ui/generator.ts
molant: Done

src/lib/cli.ts
molant: Done

qzhou1607:
Looks good to me. Just one minor thing that I should have changed in the ts rule template `rule-script-ts`: Instead of `import * from '../../utils/rule-helpers'`, it should use `import * as helper from '../../utils/rule-helpers`. And the corresponding line in `tests/lib/cli/fixtures/new.txt` needs to be changed as well. Would you mind adding that? :)

molant:
@qzhou1607 I don't quite follow what you mean. Seems like `rule-helpers` isn't used in `rule-script-ts` so maybe it will have to be removed. Anyways, let's get this merged and then you can do a PR with that change.

austinkelleher:
Why are these changes not showing up in the `master` branch? /cc @molant @sarvaje

sarvaje:
> Why are these changes not showing up in the master branch? /cc @molant @sarvaje

That's weird, Maybe something I mess up with the rebase, I can see the commit for this PR in in the PR #296

PR #298, opened 2017-06-22
Fix failing request test using hardcoded port.

austinkelleher:
This test is failing because of the hardcoded port.


molant:
wow, I wonder how the build passed in the first place when submitting the PR.
Thanks!

PR #306, opened 2017-06-22
Specify required node version in package.json

t-sauer:
With this field npm will show a warning and Yarn will prevent the installation if the package is installed with an unsupported node version.


dmethvin:
Was just about to submit this myself! :+1:

PR #309, opened 2017-06-22
Fix typo in disown-opener.md

dhritzkiv:
`typeof` -> `target`

https://sonarwhal.com/docs/user-guide/rules/disown-opener.html


austinkelleher:
Thanks again!

PR #296, opened 2017-06-21
Eliminate as many `any` as possible

sarvaje:

PR #312, opened 2017-06-22
Change how sonar resources are loaded

### LINKED ISSUE #234: Test requiring rules, formatters, etc. defined in packages

molant:
The current order is:

1. Resources in core
2. Scoped packages (@sonarwhal/XXX)
3. Prefixed packages (sonar-resourceType-XXX)

The previous way of loading resources was weak and is better to rely
on `require` when dealing with external resources.

Some improvements of the current code over the previous:

* We load only the resources that we need instead of all, and on demand.
* `sonar.create()` is now sync and should load faster.
* Validation of rules is done now in `sonar` (when loading them individually
  and not the `cli`. This fixes a previously unkown issue on `rule-runner`
  where the schema of rules wasn't validated (it showed up in `ssllabs`.
* Remove all code related to `plugins` until we know how they look like.
* Refactor where the code to generate a valid `.sonarrc` is.
* Remove `sonar.create()` in favor of `new Sonar()`.
* Clean up several tests and added @types/sinon.
* Refactor `ui` to `cli` and move all `generators` there.

Fix #234


molant:
Take 2, let's see if this one goes to the right place.
Can you take a look to double check the code is the same as the original one?

austinkelleher:
This looks the same as what I had seen in your last PR. Should be fine to merge. I'm not sure how this will affect #296 since your commit is showing up there too.

molant:
@sarvaje can probably rebase once this is merged into master and hopefully things will be back to normal.

PR #320, opened 2017-06-23
283 - Improve rules configuration. Allow rules to be an array.

austinkelleher:
**Do not merge** until the issue with the build not working on UNIX systems is fixed. The tests were passing, but I've made a few more changes without running the tests (since I can't right now). The Travis tests will obviously fail because the build is not passing at the moment. I just wanted to get any feedback from @molant before he leaves on vacation.

This change allows arrays to be passed for `rules` in the config:

```
{
    "rules": [
        "rule1",        // Default to "error"
        "rule2:warn",   // Switch to warn
        "rule3:off",    // Quickly disable a rule
        ["rule4", {
            something: true
        }]
    ]
}
```

It also allows for `warn` and `off` shorthands:

```
{
    "rules": [
    "rule1",          // Default to "error"
    "?rule2",        // Switch to warn
    "-rule3"        // Quickly disable a rule
}
```

My change is backwards compatible, so passing objects is still supported:

```
{
    "rules": {
        "rule1": "warning",
        "rule2": "off"
    }
}
```

### DIFF
--- a/tests/lib/sonar.ts
+++ b/tests/lib/sonar.ts
@@ -142,7 +142,6 @@ test.serial('If config.rules is an object with rules, we should create just thos
     t.context.rule = rule;
     sinon.stub(t.context.resourceLoader, 'loadRules').returns(new Map([
         ['disallowed-headers', rule],
-        ['lang-attribute', rule],
         ['manifest-exists', rule]
     ]));
     sinon.stub(rule, 'create')
@@ -180,7 +179,6 @@ test.serial(`If config.rules has some rules "off", we shouldn't create those rul
     t.context.rule = rule;
     sinon.stub(t.context.resourceLoader, 'loadRules').returns(new Map([
         ['disallowed-headers', rule],
-        ['lang-attribute', rule],
         ['manifest-exists', rule]
     ]));
     sinon.stub(rule, 'create').returns({ 'fetch::end': () => { } });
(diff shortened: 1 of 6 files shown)


austinkelleher:
@molant @sarvaje This is ready to be reviewed now. The build is back in working order.

molant:
@austinkelleher we've been using the label `status;work-in-progress` in the PRs to indicate that they shouldn't be merged.

austinkelleher:
@molant It is not a work in progress anymore now that the build id fixed.

austinkelleher:
@molant Oh I see you removed it. I will add it to future WIP PRs though. Thanks.

tests/lib/sonar.ts
```diff
@@ -198,6 +198,143 @@ test.serial(`If config.rules has some rules "off", we shouldn't create those rul
...
+        collector: 'collector',
+        rules: [
+            'disallowed-headers:warning',
+            'manifest-exists:warning'
+        ]
+    });
+
+    t.true(t.context.resourceLoader.loadRules.called);
+    t.true(t.context.rule.create.calledTwice);
+    t.true(t.context.eventemitter.prototype.on.calledTwice);
```
molant: I think we need to update this test and the previous one. `loadRules` should just return 2 rules now that we only load the ones we need. My mistake for not updating it in my PR :(

sarvaje:
LGTM. But I have another question:
Should this PR also include the `include/exclude` options?

>Add comments so they pop up in intellisense
Update the documentation in the user guide to say we support this format as well

+1

molant:
@alrra comments on this PR?

@austinkelleher do you want to add the include/exclude to this PR as well? If not, can you open a new issue to track that?
I think that once you add the jsdoc and documentation we should be ready to merge this (at least from my pocket).

austinkelleher:
@molant I was thinking that include/exclude should be tracked separately. Especially since there is some confusion with my other PR #292, which will be replaced by include/exclude. I will make the doc changes and fix conflicts tomorrow.

austinkelleher:
@molant @sarvaje Let me know if you think this is ready and I will squash and merge.

sarvaje:
@austinkelleher for me is OK. But I think you have a message from @molant pending.

Before merge, you have some conflicts.

PR #333, opened 2017-06-27
Add rule to check the markup validity

### LINKED ISSUE #28: Add rule to check the markup validity of web documents
`TODO:` Look into how the [`Nu HTML Checker`](https://validator.w3.org/nu/about.html) can be integrated.

qzhou1607:
Fix #28

### DIFF
--- a/.sonarrc
+++ b/.sonarrc
@@ -15,6 +15,7 @@
         "disallowed-headers": "warning",
         "disown-opener": "warning",
         "highest-available-document-mode": "warning",
+        "html-checker": "warning",
         "manifest-exists": "warning",
         "manifest-file-extension": "warning",
         "manifest-is-valid": "warning",
--- a/docs/user-guide/rules/html-checker.md
+++ b/docs/user-guide/rules/html-checker.md
@@ -1,0 +1,68 @@
+# The Nu HTML Test (`html-checker`)
+
+`html-checker` validates the markup of a website against the [Nu HTML checker](https://validator.github.io/validator/).
+
+## Why is this important?
+
+> Serving valid HTML nowadays have been commonly overlooked these days.
+By running the HTML documents through a checker, it's easier to catch
+unintended mistakes which might have otherwise been missed.
+Adhering to the W3C' standards has a lot to offer to both the
+developers and the web users: It provides better browser compatibility,
+helps to avoid potential problems with accessibility/usability, and makes it easier for future maintainance.
+>
+> The Nu Html Checker(v.Nu) serves as the backend of [checker.html5.org](https://checker.html5.org/),
+[html5.validator.nu](https://html5.validator.nu), and [validator.w3.org/nu](https://validator.w3.org/nu/).
+It also provides a [web service interface](https://github.com/validator/validator/wiki/Service-%C2%BB-HTTP-interface).
+This rule interacts with this service via [html-validator](https://www.npmjs.com/package/html-validator),
+and is able to test both remote websites and local server instances.
+
+## What does the rule check?
+
+According to the Nu Html checker [documentation](https://validator.w3.org/nu/about.html), the positive cases contain two sections:
+
+* Markup cases that are potential problems for accessibility, usability,
+  interoperability, security, or maintainability—or because they can result in poor performance,
+  or that might cause your scripts to fail in ways that are hard to troubleshoot.
+
+* Markup cases that are defined as errors because they can cause you to run into potential
+  problems in HTML parsing and error-handling behavior—so that, say, you’d end up with some unintuitive, unexpected result in the DOM.
+
+For explanation behind those requirements, please checkout:
+
+* [rationale for syntax-level errors](https://www.w3.org/TR/html/introduction.html#syntax-errors)
+* [rationale for restrictions on content models and on attribute values](https://www.w3.org/TR/html/introduction.html#restrictions-on-content-models-and-on-attribute-values)
+
+## Can the rule be configured?
+
+You can ignore certain error/warning by setting the `ignore` option for the `html-checker` rule.
+You can either pass in a string or an array that contains all the messages to be ignored.
+
+E.g. The following configuration will ignore the errors/warnings with the message of `Invalid attribute`:
+
+```json
+"html-checker": ["error", {
+    "ignore": "Invalid attribute"
+}]
+```
+
+Alternative, you can pass in an array if you have more than one type of messages to ignore:
+
+```json
+"html-checker": ["error", {
+    "ignore": ["Invalid attribute", "Invalid tag"]
+}]
+```
+
+You can also override the default validator by passing in the endpoint of an alternative validator. However, you need to make sure that this alternative validator exposes the same REST interface as the default one.
+
+```json
+"html-checker": ["error", {
+    "validator": "http://html5.validator.nu"
+}]
+```
+
+## Further Reading
+
+* [Why Validate Using the Nu Html Checker?](https://validator.w3.org/nu/about.html)
+* [The Nu Html Checker Wiki](https://github.com/validator/validator/wiki)
--- a/docs/user-guide/rules/index.md
+++ b/docs/user-guide/rules/index.md
@@ -16,6 +16,7 @@
 * [`content-type`](content-type.md)
 * [`highest-available-document-mode`](highest-available-document-mode.md)
 * [`no-friendly-error-pages`](no-friendly-error-pages.md)
+* [`html-checker`](html-checker.md)
 
 ## Performance
 
--- a/src/lib/rule-context.ts
+++ b/src/lib/rule-context.ts
@@ -88,7 +88,7 @@ export class RuleContext {
     }
 
     /** Reports a problem with the resource. */
-    public async report(resource: string, element: IAsyncHTMLElement, message: string, content?: string, location?: IProblemLocation, severity?: Severity): Promise<void> { //eslint-disable-line require-await
+    public async report(resource: string, element: IAsyncHTMLElement, message: string, content?: string, location?: IProblemLocation, severity?: Severity, codeSnippet?: string): Promise<void> { //eslint-disable-line require-await
         let position: IProblemLocation = location;
         let sourceCode: string = null;
 
@@ -107,7 +107,7 @@ export class RuleContext {
         this.sonar.report(
             this.id,
             severity || this.severity,
-            sourceCode,
+            codeSnippet || sourceCode,
             position,
             message,
             resource
--- a/src/lib/rules/html-checker/html-checker.ts
+++ b/src/lib/rules/html-checker/html-checker.ts
@@ -1,0 +1,155 @@
+/**
+ * @fileoverview Validating html using `the Nu html checker`;
+ * https://validator.w3.org/nu/
+ */
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+import { debug as d } from '../../utils/debug';
+import { RuleContext } from '../../rule-context'; // eslint-disable-line no-unused-vars
+import { IRule, IRuleBuilder, ITargetFetchEnd, IScanEnd, IProblemLocation, Severity } from '../../types'; // eslint-disable-line no-unused-vars
+
+const debug: debug.IDebugger = d(__filename);
+
+// ------------------------------------------------------------------------------
+// Public
+// ------------------------------------------------------------------------------
+
+const rule: IRuleBuilder = {
+    create(context: RuleContext): IRule {
+        /** The promise that represents the scan by html checker. */
+        let htmlCheckerPromise: Promise<any>;
+        /** Array of strings that needes to be ignored from the checker result. */
+        let ignoredMessages;
+        /** The options to pass to the html checker. */
+        const scanOptions = {
+            data: '',
+            format: 'json',
+            validator: ''
+        };
+
+        type HtmlError = { // eslint-disable-line no-unused-vars
+            extract: string, // code snippet
+            firstColumn: number,
+            lastLine: number,
+            hiliteStart: number,
+            message: string,
+            subType: string
+        };
+
+        const loadRuleConfig = () => {
+            const ignore = (context.ruleOptions && context.ruleOptions.ignore) || [];
+            const validator = (context.ruleOptions && context.ruleOptions.validator) || 'http://validator.w3.org/nu/';
+
+            scanOptions.validator = validator;
+
+            // Up to now, the `ignore` setting in `html-validator` only works if `format` is set to `text`
+            // So we implement `ignore` in our code rather than pass it to `scanOptions`
+            // TODO: Pass `ignore` once this issue (https://github.com/zrrrzzt/html-validator/issues/58) is solved.
+            ignoredMessages = Array.isArray(ignore) ? ignore : [ignore];
+        };
+
+        // Filter out ignored messages
+        const filter = (messages) => {
+            return messages.filter((message) => {
+                return !ignoredMessages.includes(message.message);
+            });
+        };
+
+        const locateAndReport = (resource: string) => {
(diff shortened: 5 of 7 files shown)


molant:
@qzhou1607 there are some conflicts. Can you please solve them?

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,107 @@
...
+        };
+    },
+    meta: {
+        docs: {
+            category: 'Interoperability',
+            description: 'Validating html using `the Nu html checker`'
+        },
+        fixable: 'code',
+        recommended: true,
+        schema: [],
```
molant: Doesn't the checker allow any configuration? I thought we could ignore some type of errors, etc.

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,107 @@
...
+                    elementColumn: error.hiliteStart + 1,
+                    elementLine: 1, // We will pass in the single-line code snippet generated from the html checker, so the elementLine is always 1
+                    line: error.lastLine
+                };
+
+                return context.report(resource, null, error.message, null, position, Severity[error.subType], error.extract);
+            });
+
+            try {
+                await Promise.all(reportPromises);
```
molant: Can we add support for async iterator? They should be working on typescript 2.3 (check the release notes, there might be some hack to do to make it work everywhere).

Also, I'd refactor the reporting and problem location code to another method and then call it in the iteration.

qzhou1607:
@molant In this PR, I use the `html-validator` dependency, which uses the web service interface of the validator ( I doubled checked the web service and didn't see a limit of requests anywhere.). An alternative is to use the [vnu-jar](https://www.npmjs.com/package/vnu-jar), but using java with node seems complicated according to @alrra . Do you have any opinion on that which one to use?

src/lib/rule-context.ts
```diff
@@ -107,7 +107,7 @@ export class RuleContext {
         this.sonar.report(
             this.id,
             severity || this.severity,
-            sourceCode,
+            sourceCode || codeSnippet,
```
sarvaje: IMO if the `codeSnippet` is present, maybe it should have preference over the `sourceCode`

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,107 @@
...
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+import { debug as d } from '../../utils/debug';
+import { IRule, IRuleBuilder } from '../../types'; // eslint-disable-line no-unused-vars
+import { RuleContext } from '../../rule-context'; // eslint-disable-line no-unused-vars
+
+import { ITargetFetchEnd, IScanEnd, IProblemLocation, Severity } from '../../types'; // eslint-disable-line no-unused-vars
```
sarvaje: the `import` for `../../types` is duplicate, merge everything in just one `import`

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,107 @@
...
+
+const debug = d(__filename);
```
sarvaje: add the type to debug `debug.IDebugger`

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,107 @@
...
+            if (!htmlCheckerPromise) {
+                return;
+            }
+
+            debug(`Waiting for Html Checker results for ${resource}`);
+
+            try {
+                result = await htmlCheckerPromise;
+            } catch (e) {
+                debug(`Error getting html checker result for ${resource}.`);
```
austinkelleher: I think you should log the actual error.

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,107 @@
...
+            firstColumn: number,
+            lastLine: number,
+            hiliteStart: number,
+            message: string,
+            subType: string
+        };
+
+        const start = (data: ITargetFetchEnd) => {
+            const { response } = data;
+            const htmlChecker = require('html-validator');
```
austinkelleher: Why is this `require` inside of the function instead of lifted to the top of the scope. It should also be an `import` right?

src/lib/rules/html-checker/html-checker.ts
austinkelleher: Should be included in the report?

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,107 @@
...
+                    line: error.lastLine
+                };
+
+                return context.report(resource, null, error.message, null, position, Severity[error.subType], error.extract);
+            });
+
+            try {
+                await Promise.all(reportPromises);
+            } catch (e) {
+                debug(`Error reporting the html checker results.`);
```
austinkelleher: Same here. I think we should log the error too.

.sonarrc
```diff
@@ -7,14 +7,15 @@
             "maxLoadWaitTime": 30000
         }
     },
-    "formatter": "stylish",
+    "formatter": "codeframe",
```
sarvaje: I think the formatter by default is `stylish`

molant:
@qzhou1607 if you are using https://www.npmjs.com/package/html-validator it had at least an option to ignore in the configuration object. That property should be exposed somehow.

qzhou1607:
@molant In `html-validator`, currently the setting of `ignore` is only supported if the output `format` is set to `text`. We have to use `json` here for further parsing. It was not difficult to implement `ignore` in our code though, so I added that. 

>@qzhou1607 if you are using https://www.npmjs.com/package/html-validator it had at least an option to ignore in the configuration object. That property should be exposed somehow.

I added documentation also.
>I don't see a page with the description of the rule. Remember to use the new-rule parameter you added to dogfood it and adjust if necessary!

I had issues with implementing async iterator. It looks like I need to convert iterables (array) into async iterables first using a generator, and then do the `for..await..of` loop. But for some reason, the rule can't be exported right whenever I drop the generator in the `create` function. Am I looking at the wrong direction?
>Can we add support for async iterator? They should be working on typescript 2.3 (check the release notes, there might be some hack to do to make it work everywhere).

src/lib/rules/html-checker/html-checker.ts
qzhou1607: @austinkelleher This is due to the mocking in tests. I added some comments there.

qzhou1607:
@molant @sarvaje @alrra I looked into `async iterator` more, still having trouble with async generators. The async generator function in typescript was directly copied when compiled and throws an error when running in Node. The `lib` setting I have in `tsconfig.json` is 
```
        "lib": [
            "es2017",
            "esnext",
            "esnext.asynciterable",
            "dom",
            "dom.iterable"
        ]
```
I also tried polyfill `Symbol.asyncIterator` according to [this post](https://stackoverflow.com/questions/43694281/ts2318-cannot-find-global-type-asynciterableiterator-async-generator) using `core-js`, still no luck, same error. So I'll keep the current code for now until I have more information about the async generators.

molant:
> I also tried polyfill `Symbol.asyncIterator` according to [this post](https://stackoverflow.com/questions/43694281/ts2318-cannot-find-global-type-asynciterableiterator-async-generator) using `core-js`, still no luck, same error. So I'll keep the current code for now until I have more information about the async generators.

Sounds good to me!

qzhou1607:
@molant Yep working on it. If we give user this option we should probably ask for the output interface (`firstColumn`, `subType` and etc.) as well, otherwise the output can't be consumed by `context.report` and our formatter.

molant:
> ce (firstColumn, subType and etc.) as well, otherwise the output can't be consumed by context.report and our formatter.

Not sure I understand why we need that. We are only changing the end point to use.

qzhou1607:
@molant A different service/endpoint may have a different REST interface (like using  `column` instead of `firstColumn` as the property name). This inconsistency could raise error when we process the output from different services. Or should we assume that other service endpoints always have the same output format as the default validator? If not, throws an error?
>Not sure I understand why we need that. We are only changing the end point to use.

molant:
We will be using the same code as the official one. There are some instructions on how to set it up.

qzhou1607:
@molant validator option was added.

molant:
@sarvaje @AustinKelleher this seems to be ready for another review.

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,155 @@
...
+        /** Array of strings that needes to be ignored from the checker result. */
+        let ignoredMessages;
+        /** The options to pass to the html checker. */
+        const scanOptions = {
+            data: '',
+            format: 'json',
+            validator: ''
+        };
+
+        type Error = { // eslint-disable-line no-unused-vars
```
qzhou1607: @sarvaje Done.

docs/user-guide/rules/html-checker.md
```diff
@@ -0,0 +1,68 @@
...
+"html-checker": ["error", {
+    "ignore": "Invalid attribute"
+}]
+```
+
+Alternative, you can pass in an array if you have more than one type of messages to ignore:
+
+```json
+"html-checker": ["error", {
+    "ignore": ["Invalid attribute", "Invalid tag"]
```
alrra: We should use more specific example here. These are too general, and it might seem like one has to ignore types/groups of errors, and cannot ignore very specific ones.

docs/user-guide/rules/html-checker.md
```diff
@@ -0,0 +1,68 @@
...
+}]
+```
+
+You can also override the default validator by passing in the endpoint of an alternative validator. However, you need to make sure that this alternative validator exposes the same REST interface as the default one.
```
alrra: `same REST interface` should link to the API.

docs/user-guide/rules/index.md
```diff
@@ -16,6 +16,7 @@
 * [`content-type`](content-type.md)
 * [`highest-available-document-mode`](highest-available-document-mode.md)
 * [`no-friendly-error-pages`](no-friendly-error-pages.md)
+* [`html-checker`](html-checker.md)
```
alrra: Always use alphabetical order. Should be before `no-friendly-error-pages`.

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,155 @@
...
+            subType: string
+        };
+
+        const loadRuleConfig = () => {
+            const ignore = (context.ruleOptions && context.ruleOptions.ignore) || [];
+            const validator = (context.ruleOptions && context.ruleOptions.validator) || 'http://validator.w3.org/nu/';
+
+            scanOptions.validator = validator;
+
+            // Up to now, the `ignore` setting in `html-validator` only works if `format` is set to `text`
```
alrra: If this is a simple change, maybe we should offer and contribute back the fix to it?

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,155 @@
...
+                        }, { type: 'string' }
+                    ]
+                },
+                validator: {
+                    pattern: '^(http|https)://',
+                    type: 'string'
+                }
+            }
+        }],
+        worksWithLocalFiles: false
```
alrra: It does not?

docs/user-guide/rules/html-checker.md
```diff
@@ -0,0 +1,68 @@
...
+`html-checker` validates the markup of a website against the [Nu HTML checker](https://validator.github.io/validator/).
+
+## Why is this important?
+
+> Serving valid HTML nowadays have been commonly overlooked these days.
+By running the HTML documents through a checker, it's easier to catch
+unintended mistakes which might have otherwise been missed.
+Adhering to the W3C' standards has a lot to offer to both the
+developers and the web users: It provides better browser compatibility,
+helps to avoid potential problems with accessibility/usability, and makes it easier for future maintainance.
```
alrra: Were possible, try to keep the text under 80 characters / line. The same goes for the lines below.

docs/user-guide/rules/html-checker.md
```diff
@@ -0,0 +1,68 @@
...
+"html-checker": ["error", {
+    "ignore": ["Invalid attribute", "Invalid tag"]
+}]
+```
+
+You can also override the default validator by passing in the endpoint of an alternative validator. However, you need to make sure that this alternative validator exposes the same REST interface as the default one.
+
+```json
+"html-checker": ["error", {
+    "validator": "http://html5.validator.nu"
```
alrra: `httpS`

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,155 @@
...
+            firstColumn: number,
+            lastLine: number,
+            hiliteStart: number,
+            message: string,
+            subType: string
+        };
+
+        const loadRuleConfig = () => {
+            const ignore = (context.ruleOptions && context.ruleOptions.ignore) || [];
+            const validator = (context.ruleOptions && context.ruleOptions.validator) || 'http://validator.w3.org/nu/';
```
alrra: `httpS`

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const debug: debug.IDebugger = d(__filename);
+
+// ------------------------------------------------------------------------------
+// Public
+// ------------------------------------------------------------------------------
+
+const rule: IRuleBuilder = {
+    create(context: RuleContext): IRule {
+        /** The promise that represents the scan by html checker. */
```
alrra: `html checker`, `Html Checker`, ... =>  Let's just use `HTML checker` consistently here, and throughout this file.

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -0,0 +1,155 @@
...
+
+        return {
+            'scan::end': end,
+            'targetfetch::end': start
+        };
+    },
+    meta: {
+        docs: {
+            category: 'Interoperability',
+            description: 'Validating html using `the Nu html checker`'
```
alrra: > Validate HTML using 'the Nu html checker'

tests/lib/rules/html-checker/tests.ts
```diff
@@ -0,0 +1,189 @@
...
+import * as mock from 'mock-require';
+
+import { RuleTest } from '../../../helpers/rule-test-type'; // eslint-disable-line no-unused-vars
+import * as ruleRunner from '../../../helpers/rule-runner';
+import { getRuleName } from '../../../../src/lib/utils/rule-helpers';
+
+const ruleName = getRuleName(__dirname);
+const exampleUrl = 'https://example.com/';
+const validatorError = 'error';
+const defaultValidator = 'http://validator.w3.org/nu/';
```
alrra: `httpS`

tests/lib/rules/html-checker/tests.ts
```diff
@@ -0,0 +1,189 @@
...
+const configValidator = 'http://html5.validator.nu';
```
alrra: `httpS`

tests/lib/rules/html-checker/tests.ts
```diff
@@ -0,0 +1,189 @@
...
+            hiliteLength: 4
+        }
+    ]
+};
+
+const htmlCheckerMock = (response) => {
+    const mockedChecker = (scanOptions) => {
+        let responseMessages;
+
+        if (response.pass) { // No errors/warnings are detected in the target html
```
alrra: `html` => Use `HTML` here, and everywhere else.

alrra:
Merged with some modifications in 9f73cc9b6071193a12933d756883738a0308a6b6.

PR #349, opened 2017-06-30
Docs: Emphasize more the full `User guide` in `README.md`

### LINKED ISSUE #326: sonar init, what are collectors?
Per the docs, I'm running `sonar --init`, and I'm being asked to select a collector. I have a choice of cdp and json. I've got no idea what a collector is, or cdp. I know what json is obviously, but... yeah, no idea. I'm going to select json, but I think there are two things here:

* the quick start should let us know about this prompt and give us some context as to what to pick
* the CLI should maybe add a bit more text?

alrra:
@cfjedimaster Do you think this change will make it more obvious to new users that they should look at the [full `user guide`](https://sonarwhal.com/docs/user-guide/) if they don't understand something from the `quick start user guide` from the `README.md`?

---

Fix #326

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@
 
 ## Quick start user guide
 
-Once you have node 8.x in your machine, install `sonar`:
+Once you have [`Node.js`](https://nodejs.org/en/download/current/)
+v8.x on your machine, install `sonar`:
 
 ```bash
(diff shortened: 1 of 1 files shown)


cfjedimaster:
Yep!

alrra:
@cfjedimaster Thanks for the feedback!

PR #356, opened 2017-07-04
Refactor debug and logging module

### LINKED ISSUE #342: Refactor debug and logger
Maybe we can do something similar to what we do to debug where we pass the __filename to add more data to logger and merge everything.

qzhou1607:
Fix #342

* Merge `debug` and `logging`
* Edit tests that involve the stubbing of `logger`

To stub `log` method returned by the `loggerInitiator`, the `logging` module has to be required later in the scripts rather than in the beginning. It is not ideal. Please let me know if you see a better way...


src/lib/utils/logging.ts
```diff
@@ -2,16 +2,69 @@
...
+        output = 'engine';
+    }
+
+    return `sonar:${output}`;
 };
 
-/** Cover for console.log */
-export const log = (message: any, ...optionalParams: Array<any>) => {
-    console.log(message, ...optionalParams);
+export const loggerInitiator = (filePath: string, includesModuleInfo: boolean = true) => {
```
molant: We are creating a bunch of this and each time a new `error`, `log` and `debug`. Let's create a class `Logger` and export it so we call it like this:
```js
import { Logger } from './utils/logger'; // Will have to rename the file
const logger = new Logger(__filename);
```

src/lib/cli/rule-generator.ts
```diff
@@ -80,7 +80,7 @@ const loadPopulatedTemplate = async (filePath: string, data: Object): Promise<st
     try {
         content = await util.promisify(fs.readFile)(filePath, 'utf8');
     } catch (e) {
-        debug(`Error reading file: ${filePath}`);
+        logger.debug(`Error reading file: ${filePath}`);
```
austinkelleher: I notice in several places you're using `logger.debug` to log errors. I think `logger.error` would be more appropriate.

qzhou1607:
@austinkelleher I like the idea of configuring log levels. But I'm a little confused about the implementation of `utils/logging`. Why do you return `logger` in `configure` and also export a function `logger`? 

Also, just to clarify, in the current code, `log`, `error` and `debug` are not exported individually. A `loggerInitiator` is exported and normally used in this way in each file:
```
const logger = loggerInitiator(__filename);

logger.log('...');
logger.error('...');
logger.debug('...');
```
`require('./utils/logging')(__filename)` were used in some scripts because of the mocking of `log` or `error` in the tests for these scripts. `sinon.spy` in the following test (See tests for `codeframe` as an example) doesn't work if logger is initiated in the beginning of the script (And I still don't know how to solve this problem) :
```
const loggerMethods = {
    debug() { },
    error() { },
    log() { }
};

const logging = {
    loggerInitiator: () => {
        return loggerMethods;
    }
};

proxyquire('../../../src/lib/formatters/codeframe/codeframe', { '../../utils/logging': logging.loggerInitiator });

test.beforeEach((t) => {
    sinon.spy(loggerMethods, 'log');

    t.context.logger = loggerMethods;
});
```
So I'm afraid we will still have the same issue in your API and will have to do `const logger = require('./utils/logging').logger(__filename)` here and there. Is my understanding correct?

alrra:
Are we doing anything with this, or should we close this?

molant:
This code is very old and has lots of conflicts. Also I'm not sure this was the best strategy so closing until we have a better idea.

PR #358, opened 2017-07-04
Switch to using `connector` instead of `collector`

### LINKED ISSUE #286: [Proposal] Move away from the term "Collectors" for end users?
I understand why you went with the name "Collectors", but I feel like "Browser Engines" would be more relatable to by end users. At first glance, I was confused by the usage of "Collectors". Maybe keep "collector" as more of an implementation detail? Maybe use "connector" instead ("Connectors are used to connect Sonar with browser engines")?

alrra:
Ref #286
Fix #286


src/lib/cli/sonarrc-generator.ts
```diff
@@ -59,9 +59,9 @@ export const initSonarrc = async () => {
 
     const questions = [
         {
-            choices: collectorKeys,
-            message: 'What collector do you want to use?',
-            name: 'collector',
+            choices: connectorKeys,
+            message: 'What connector do you want to use?',
```
molant: We could improve already the question here and briefly explain what a connector is. Maybe add a description for each one? Something similar to:
> What connector do you want to load the content of the website?
> * cdp (you will need to have chrome installed)
> * jsdom (will use jsdom, no additional installation required)

alrra: That should be handled separately from this PR.

I've opened: https://github.com/sonarwhal/sonar/issues/361.

PR #367, opened 2017-07-06
Improve documentations

qzhou1607:
Some updates made along the way of reading the rules.

One question - in the `event` documentation, under `manifest::missing`, it says "The event is the same for `targetfetch::start`". It doesn't look correct to me. Can you confirm? @alrra @molant


molant:
@qzhou1607 it should say it has the same properties as targetfetch::start, which should be just the resource property. We can probably make this section clearer. I believe @sarvaje did some refactoring with the events so we should update all these type of references to say they are the basic one.

alrra:
@qzhou1607 Let me know when this is ready for another review.

Issue #24, opened 2017-03-09
Add rule to check for `Secure` and `HttpOnly` cookies


alrra:
- [ ]  Check if the `Set-Cookie` header is sent with the `Secure` and `HttpOnly` values if the page is served over HTTPS.

---

See also:

* https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Secure_and_HttpOnly_cookies

qzhou1607:
A few more checks:
* [ ] In the same set-cookie-string, two attributes shouldn't have the same name.
* [ ] The name-value-pair string needs to have `=`, otherwise it will be ignored entirely.

PR #370, opened 2017-07-08
Add strict-transport-security rule

### LINKED ISSUE #23: Add rule(s) to check the usage of the `Strict-Transport-Security` HTTP header
- [x] Check if the header `Strict-Transport-Security` is sent for resources served over HTTPS.
- [x] Check if the `max-age` value  is small (less than 10886400 seconds should be an error, no configuration possible).

---

Also inform users about https://hstspreload.org/? However, make it clear that:

  * the preload list cannot easily be undone (domains can be removed, but it takes months for a change to reach users, basically until those users upgrade to a new version of the browser)
  * they should only do this if they are sure that they can support HTTPS for their entire site and all its `subdomains` in the long run.

---

See also:

* https://tools.ietf.org/html/rfc6797
* https://www.owasp.org/index.php/HTTP_Strict_Transport_Security

### LINKED ISSUE #369: Add support for HTTPS in testing
- [x] Generate self-assigned certificates
- [x] Edit `cdp` and `jsdom` settings to not throw errors/hang when using self-assigned certificates
- [x] Integrate both `http` and `https` options in the server

qzhou1607:
Fix #23 
Fix #369 

* Add rule
* Add test
* Add self signed certificate
* Adjust cdp and jsdom so that it doesn't get stuck on "Your connection is not private"
* Add the option to use a HTTPS server

### DIFF
--- a/docs/user-guide/rules/strict-transport-security.md
+++ b/docs/user-guide/rules/strict-transport-security.md
@@ -1,0 +1,147 @@
+# Require `Strict-Transport-Security` response header (`strict-transport-security`)
+
+`strict-transport-security` warns against serving resources over HTTPS without
+`strict-transport-security` header and validates the header directives
+and their corresponding values.
+
+## Why is this important?
+
+Web security has been a crutial concern for the developers nowadays. Unlike
+[cross-site scripting(XSS)][XSS] and [SQL injection](SQL injection), the
+exploit of insufficient protection over the transport layer can be harder to picture
+in practice. If a website accepts a connection through HTTP and then redirects to HTTPS,
+it opens up opportunites for a "man-in-the-middle" attack, when the redirect could be
+exploited and lead the user to a malicious site.
+
+By specifying the `Strict-Transport-Security` header along with a `max-age` value in
+the reponse, a website is able to declare themselves to be only accessible via the
+secure connections within the specified time period. So for future attempts to the
+same domain via the insecure connections, the browswer knows that it should never
+load the site using HTTP and automatically convert all attempts to HTTPS requests instead.
+
+Notably, to prevent the `Strict-Transport-Security` header from being stripped by
+the attacker on the user's first visit, major browsers include a "pre-loaded" list
+of sites that must be loaded via HTTPS. You can submit your domain name in the
+[online form][preload form] to be included in the list. After being included,
+browsers will never be able to connect to your domain using an insecure connection.
+So use with great caution: Before you decide to have your own domain included, make
+sure that you are able to support HTTPS for all the subdomains and will never again
+need the insecure scheme.
+
+More information about HTTP Strict Transport (HSTS), please see:
+
+* [HTTP Strict Transport Security wiki][HSTS wiki]
+* [HTTP Strict Transport Security Cheat Sheet][HSTS cheat sheat]
+
+## What does the rule check?
+
+For a site served over HTTPS, this rule checks the following:
+
+* If it has a `Strict-Transport-Security` header.
+* If the header has the required `max-age` directive.
+* If the `max-age` directive has a value that is longer than 18 weeks(10886400s).
+* If `Strict-Transport-Security` header has repetitive directives.
+* When a `Strict-Transport-Security` header contains `preload` directive, this rule
+
+will first check the domain name against the [HTTP Strict Transport Security (HSTS) preload list][preload list]
+for the preload status, and then check whether this domain has errors that would prevent
+preloading by calling the hstspreload api endpoint. This check is disabled by default.
+
+### Examples that **trigger** the rule
+
+`Strict-Transport-Security` response header was not sent over `HTTPS`:
+
+```text
+HTTP/... 200 OK
+
+...
+```
+
+`Strict-Transport-Security` response header is sent with a `max-age` value that is too short:
+
+```text
+HTTP/... 200 OK
+
+...
+Strict-Transport-Security: max-age=1
+```
+
+`Strict-Transport-Security` response header is sent without `max-age` directive:
+
+```text
+HTTP/... 200 OK
+
+...
+Strict-Transport-Security: maxage=31536000
+```
+
+`Strict-Transport-Security` response header is sent with duplicate `includeSubDomains` directives:
+
+```text
+HTTP/... 200 OK
+
+...
+Strict-Transport-Security: includeSubDomains; max-age=31536000; includeSubDomains
+```
+
+### Examples that **pass** the rule
+
+```text
+HTTP/... 200 OK
+
+...
+Strict-Transport-Security: max-age=31536000
+```
+
+```text
+HTTP/... 200 OK
+
+...
+Strict-Transport-Security: max-age=31536000; includeSubDomains
+```
+
+```text
+HTTP/... 200 OK
+
+...
+ Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
+```
+
+## Can the rule be configured?
+
+Yes, you can configure the value that `max-age` is checked against with. By default,
+this limit is set as 18 weeks (10886400s);
+
+E.g. The following configuration will change the `max-age` value limit to `123456`.
+
+```json
+"strict-transport-security": ["error", {
+    "minMaxAgeValue": 123456
+}]
+```
+
+Also, you can configure the rule so that if `preload` directive is included in the header,
+it will check whether this domain has errors that would prevent preloading by calling the
+hstspreload api endpoint. This validation is disabled by default.
+
+E.g. The following configuration will enable the `preload` validation.
+
+```json
+"strict-transport-security": ["error", {
+    "checkPreload": true
+}]
+```
+
+## Further Reading
+
+* [The beginners guide to breaking website security with nothing more than a Pineapple][pineapple]
+* [Understanding HTTP Strict Transport Security(HSTS) and preloading it into the browser][understading HSTS]
+
+[XSS]: https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)
+[SQL injection]: https://www.owasp.org/index.php/SQL_Injection
+[HSTS wiki]: https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security
+[HSTS cheat sheat]: https://www.owasp.org/index.php/HTTP_Strict_Transport_Security_Cheat_Sheet
+[pineapple]: https://www.troyhunt.com/the-beginners-guide-to-breaking-website/
+[understading HSTS]: https://www.troyhunt.com/understanding-http-strict-transport/
+[preload form]:https://hstspreload.org/
+[preload list]:https://cs.chromium.org/codesearch/f/chromium/src/net/http/transport_security_state_static.json
--- a/src/lib/rules/strict-transport-security/strict-transport-security.ts
+++ b/src/lib/rules/strict-transport-security/strict-transport-security.ts
@@ -1,0 +1,246 @@
+/**
+ * @fileoverview Check if responses served over HTTPS also have the Strict-Transport-Security header with a proper value max-age value.
+ */
+import * as url from 'url';
+
+import { RuleContext } from '../../rule-context'; // eslint-disable-line no-unused-vars
+import { debug as d } from '../../utils/debug';
+import { IAsyncHTMLElement, IResponse, IFetchEnd, IRule, IRuleBuilder, INetworkData } from '../../types'; // eslint-disable-line no-unused-vars
+
+const debug = d(__filename);
+
+// ------------------------------------------------------------------------------
+// Public
+// ------------------------------------------------------------------------------
+
+const rule: IRuleBuilder = {
+    create(context: RuleContext): IRule {
+        /** The minimum period (in seconds) allowed for `max-age`. */
+        let minMaxAgeValue: number;
+        /** Whether or not check the preload attribute */
+        let checkPreload: boolean;
+        /** Endpoint to verify that the domain name has already been included in the `preload list` */
+        const statusApiEndPoint = `https://hstspreload.org/api/v2/status?domain=`;
+        /** Endpoint to verify that the domain name is qualified to be preloaded */
+        const preloadableApiEndPoint = `https://hstspreload.org/api/v2/preloadable?domain=`;
+
+        /* HACK: Need to do a require here in order to be capable of mocking
+                when testing the rule and `import` doesn't work here. */
+        const { isHTTPS, isRegularProtocol, normalizeString, requestJSONAsync } = require('../../utils/misc');
+
+        const loadRuleConfigs = () => {
+            minMaxAgeValue = (context.ruleOptions && context.ruleOptions.minMaxAgeValue) || 10886400; // 18 weeks
+            checkPreload = (context.ruleOptions && context.ruleOptions.checkPreload);
+        };
+
+        // STS header Syntax:
+        // Strict-Transport-Security: max-age=<expire-time>
+        // Strict-Transport-Security: max-age=<expire-time>; includeSubDomains
+        // Strict-Transport-Security: max-age=<expire-time>; preload
+        // This function accomplishes the following:
+        // "max-age=31536000; includesubdomains; preload" => {"max-age":31536000,"includesubdomains":true,"preload":true}
+        const parse = (headerValue: string) => {
+            const parsedHeader = {};
+            const directives = headerValue.toLowerCase().split(';');
+            const nameValuePairRegex = /^ *([!#$%&'*+.^_`|~0-9A-Za-z-]+) *= *("(?:[~0-9])*"|[!#$%&'*+.^_`|~0-9]+) *$/;
+            // Regex for name-value pairs. E.g.: max-age=31536000
+            // Modified usage of https://github.com/jshttp/content-type/blob/64bde0d996ccb4334341662c0c7d25f7b370c4d9/index.js#L23
+            const tokenRegex = /^ *[!#$%&'*+.^_`|~0-9A-Za-z-]+$/; // Regex for single tokens. E.g.:  includesubdomains
+
+            directives.forEach((directive) => {
+                const match = tokenRegex.exec(directive) || nameValuePairRegex.exec(directive);
+
+                if (!match) {
+                    throw new Error(`'strict-transport-security' header has the wrong format: ${directive}`);
+                }
+
+                const [matchString, key, value] = match;
+                const name = key || matchString.trim();
+
+                if (parsedHeader[name]) {
+                    throw new Error(`'strict-transport-security' header contains more than one '${name}'`);
+                }
+
+                parsedHeader[name] = value || true;
+            });
+
+            return parsedHeader;
+        };
+
+        const isUnderAgeLimit = (maxAge: string, limit: number): boolean => {
+            return maxAge && parseInt(maxAge) < limit;
+        };
+
+        const isPreloaded = (hostname: string): Promise<{ [key: string]: any }> => {
+            debug(`Waiting to get preload status for ${hostname}`);
+
+            return requestJSONAsync(`${statusApiEndPoint}${hostname}`);
+        };
+
+        const issuesToPreload = (hostname: string): Promise<{ [key: string]: any }> => {
+            debug(`Waiting to get preload eligibility for ${hostname}`);
+
+            return requestJSONAsync(`${preloadableApiEndPoint}${hostname}`);
+        };
+
+        const verifyPreload = async (resource): Promise<{ [key: string]: any }> => {
+            const originalDomain = url.parse(resource).hostname;
+            const mainDomain = originalDomain.replace(/^www./, '');
+            // Some hostnames in the list include `www.`, e.g., `www.gov.uk`.
+            let status: string;
+            let issues: { [key: string]: any } = {};
+
+            try {
+                ({ status } = await isPreloaded(mainDomain) || await isPreloaded(originalDomain));
+            } catch (err) {
+                const message = `Error with getting preload status for ${resource}.`;
+
+                debug(message, err);
+                await context.report(resource, null, message);
+
+                return issues;
+            }
+
+            debug(`Received preload status for ${resource}.`);
+
+            if (!status) {
+                const message = `Error with getting preload status for ${resource}. There might be something wrong with the verification endpoint.`;
+
+                debug(message);
+                await context.report(resource, null, message);
+
+                return issues;
+            }
+
+            if (status !== 'preloaded') {
+                try {
+                    issues = await issuesToPreload(mainDomain);
+                } catch (err) {
+                    const message = `Error with getting preload eligibility for ${resource}.`;
+
+                    debug(message, err);
+                    await context.report(resource, null, message);
+                }
+
+                debug(`Received preload eligibility for ${resource}.`);
+            }
+
+            return issues;
+        };
+
+        const validate = async (fetchEnd: IFetchEnd) => {
+            const { element, resource, response }: { element: IAsyncHTMLElement, resource: string, response: IResponse } = fetchEnd;
+
+            // This check does not apply if URI starts with protocols others than http/https.
+            if (!isRegularProtocol(resource)) {
+                debug(`Check does not apply for URI: ${resource}`);
+
+                return;
+            }
+
+            const headerValue: string = normalizeString(response.headers && response.headers['strict-transport-security']);
+            let parsedHeader;
+
+            if (!isHTTPS(resource) && headerValue) {
+                await context.report(resource, element, `'strict-transport-security' header should't be specified in pages served over HTTP.`);
+
+                return;
+            }
+
+            if (!isHTTPS(resource) && !headerValue) {
+                const httpsResource = url.format(Object.assign(url.parse(resource), { protocol: `https` }));
+
+                try {
+                    const networkData: INetworkData = await context.fetchContent(httpsResource);
+
+                    if (!networkData || !networkData.response) {
+                        return;
+                    }
+
+                    if (networkData.response.statusCode === 200) {
+                        validate({
+                            element: null,
+                            request: networkData.request,
+                            resource: httpsResource,
+                            response: networkData.response
+                        });
+                    }
+                } catch (err) { // eslint-disable-line no-empty
+                    // HTTPS site can't be fetched, do nothing.
+                    debug(`${resource} doesn't support HTTPS`);
+                }
+
+                return;
+            }
+
+            // Check if the header `Strict-Transport-Security` is sent for resources served over HTTPS.
+            if (!headerValue) {
+                await context.report(resource, element, `'strict-transport-security' header was not specified`);
+
+                return;
+            }
+            // Parse header and report repetitive attributes
+            try {
+                parsedHeader = parse(headerValue);
+            } catch (err) {
+                await context.report(resource, element, err.message);
+
+                return;
+            }
+
+            // Verify preload attribute
+            if (checkPreload && parsedHeader.preload) {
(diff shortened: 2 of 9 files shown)


qzhou1607:
Still working on the documentation.

alrra:
@qzhou1607 Can you split the part related to #369 into a separate commit? Thanks!

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,70 @@
...
+    create(context: RuleContext): IRule {
+        /** The minimum period (in seconds) allowed for `max-age`. */
+        let minimum: number;
+
+        const loadRuleConfigs = () => {
+            minimum = (context.ruleOptions && context.ruleOptions.minimum) || 10886400;
+        };
+
+        const maxAgeLessThan = (headerValue: string, limit: number) => {
+            const maxAge = headerValue.match(/max-age=([^;]+);|max-age=([^;]+)$/).pop();
```
alrra: A header value parser would be great, but...

---

I think this will fail for something such as `Max-AGE=123456789`. Also, wouldn't it be more easy and clear to use `split`?

See also: https://tools.ietf.org/html/rfc6797#section-6.1

qzhou1607: @alrra You are right, I didn't know it could be uppercase. One question though - the doc says `Directive names are case-insensitive`. So that means `MAX-AGE=...` is not allowed and should **fail** the test, right? Also,  It mentioned `The order of appearance of directives is not significant`, so if we `split`, it's possible that `max-age` is not the first element. Am I understanding it correctly?

alrra: > One question though - the doc says Directive names are case-insensitive. So that means MAX-AGE=... is not allowed and should fail the test, right? 

No, it should be allowed. As you said they are case-**insensitive**, so it doesn't matter if it's `max-age`, `MAX-AGE`, or `mAx-AgE`, they are all valid.

Also, the above is the reason why I add [these types of tests](https://github.com/sonarwhal/sonar/blob/7cd550c353141a25f9f8530fabd78b3cecf9f97f/tests/lib/rules/content-type/tests.ts#L266-L273)), and you should too.

> Also, It mentioned `The order of appearance of directives is not significant`, so if we split, it's possible that max-age is not the first element. Am I understanding it correctly?

Yes.

---

Other things, the rule should check for other things, such as:

> [`The REQUIRED "max-age" directive`...](https://tools.ietf.org/html/rfc6797#section-6.1.1)

Check that the required directives are provided.

> [`All directives MUST appear only once in an STS header field. Directives are either optional or required, as stipulated in their definitions.`](https://tools.ietf.org/html/rfc6797#section-6.1)

Check that directives only appear once.

Question: What do browser do if you have multiple `max-age` directives? Do they follow the spec?

qzhou1607: @alrra I found it here: https://tools.ietf.org/html/rfc6797#section-8.1
`If a UA receives more than one STS header field in an HTTP
      response message over secure transport, then the UA MUST process
      only the first such header field.`
> Question: What do browser do if you have multiple max-age directives? Do they follow the spec?

qzhou1607:
@alrra This is ready for another round of review.

docs/user-guide/rules/strict-transport-security.md
```diff
@@ -0,0 +1,123 @@
...
+`strict-transport-security` header and validates the header directives
+and their corresponding values.
+
+## Why is this important?
+
+Web security has been a crutial concern for the developers nowadays. Unlike
+[cross-site scripting(XSS)][XSS] and [SQL injection](SQL injection), the
+exploit of insufficient protection over the transport layer can be harder to picture
+in practice. If you are a fan of the popular TV show
+"[Silicon Valley][silicon valley]", you might have heard about the intercept of
```
alrra: Let's not use the TV show as an example, and just explain the concept.

docs/user-guide/rules/strict-transport-security.md
```diff
@@ -0,0 +1,123 @@
...
+More information about HTTP Strict Transport (HSTS), please see:
+
+* [HTTP Strict Transport Security wiki][HSTS wiki]
+* [HTTP Strict Transport Security Cheat Sheet][HSTS cheat sheat]
+
+## What does the rule check?
+
+For a site served over HTTPS, this rule checks the following:
+
+* If it has a `Strict-Transport-Security` header
```
alrra: Add `.` at the end of this sentence, and the ones below it.

docs/user-guide/rules/strict-transport-security.md
```diff
@@ -0,0 +1,123 @@
...
+* If the header has the required `max-age` directive
+* If the `max-age` directive has a value that is longer than 18 weeks(10886400s)
```
alrra: Let's make the value a "round" 12 weeks?

docs/user-guide/rules/strict-transport-security.md
```diff
@@ -0,0 +1,123 @@
...
+* If `Strict-Transport-Security` header has repetitive directives
+
+### Examples that **trigger** the rule
+
+```text
```
alrra: Add short explanation for each example on why it triggers the rule (here is an [example](https://github.com/sonarwhal/sonar/blob/b013bc5dbf43e42acac34a4f6e621583bb0679a6/docs/user-guide/rules/content-type.md#examples-that-trigger-the-rule) on how it should look like).

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,70 @@
...
+            // `max-age=31536000; includeSubDomains; preload` => `31536000`
+            // `max-age=31536000` => `31536000`
+
+            return parseInt(maxAge) < limit;
+        };
+
+        const validate = async (fetchEnd: IFetchEnd) => {
+            const { element, resource, response }: { element: IAsyncHTMLElement, resource: string, response: IResponse } = fetchEnd;
+            const headerValue: string = normalizeString(response.headers && response.headers['strict-transport-security']);
+
```
alrra: ⬆️

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,124 @@
...
+                return;
+            }
+
+            // Check if the header `Strict-Transport-Security` is sent for resources served over HTTPS
+            if (!headerValue) {
+                await context.report(resource, element, `'strict-transport-security' header was not specified.`);
+
+                return;
+            }
+
```
alrra: Remove empty line.

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,124 @@
...
+        const loadRuleConfigs = () => {
+            minMaxAgeValue = (context.ruleOptions && context.ruleOptions.minMaxAgeValue) || 10886400; // 18 weeks
+        };
+
+        const parse = (headerValue: string): DirectiveMatch => {
+            const regexes = {
+                // `max-age=31536000; includeSubDomains; preload` => [`max-age=31536000`]
+                // `max-age=31536000; MAX-age=12345` => [`max-age=31536000`, `MAX-age=12345`]
+                includesubdomains: /includeSubDomains/gi,
+                'max-age': /max-age=([^;]+);|max-age=([^;]+)$/gi,
```
alrra: From the looks of the code, `max-age=10886401,includesubdomains` passes this rule, and it should not. 

---

Maybe look at how a similar parser does it (e.g.: for [`content-type`](https://github.com/jshttp/content-type/blob/64bde0d996ccb4334341662c0c7d25f7b370c4d9/index.js)).

alrra:
@qzhou1607 Also, can you change the commit to [follow the conventions](https://github.com/sonarwhal/sonar/blob/master/docs/developer-guide/contributing/pull-requests.md#commit-messages) (both commits can use the`New:` tag)?. Thanks!

qzhou1607:
@alrra Ready for review.

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,128 @@
...
+        const isInvalidValue = (maxAge: string, limit: number): boolean => {
+            return maxAge && parseInt(maxAge) < limit;
+        };
+
+        const validate = async (fetchEnd: IFetchEnd) => {
+            const { element, resource, response }: { element: IAsyncHTMLElement, resource: string, response: IResponse } = fetchEnd;
+            const headerValue: string = normalizeString(response.headers && response.headers['strict-transport-security']);
+            let parsedHeader;
+
+            if (!isHTTPS(resource)) {
```
alrra: We should do an extra check here. Even if the URL is HTTP, we should check if the site also supports `HTTPS`, and if so, still require this header.

molant: IIRC websites need to be https only for this to work. Here we should have the final url, and thus https if they redirect.

alrra: > websites need to be https only for this to work.

https only?  As in they automatically redirect you from http to https?

> Here we should have the final url, and thus https if they redirect.

Some sites don't redirect automatically.

alrra: > Some sites don't redirect automatically.

https://github.com/sonarwhal/sonarwhal.com/issues/4#issuecomment-300999687

molant: > As in they automatically redirect you from http to https?

Yes

molant: > This form is used to submit domains for inclusion in Chrome's HTTP Strict Transport Security (HSTS) preload list. This is a list of sites that are hardcoded into Chrome as being HTTPS only.

From https://hstspreload.org

alrra: > HTTP Strict Transport Security (HSTS) preload list.

@molant  Yes, that is for the HSTS preload list, not for the general usage of the`Strict-Transport-Security` header.

To be included in that list, one also need to have add the `includeSubDomains` and `preload` directives.

alrra: @molant @qzhou1607 Maybe we should have a config for this rule (something such as  `preloadList: <boolean>`) that makes the rule check if things follow the HSTS preload list requirements?

molant: > Maybe we should have a config for this rule (something such as preloadList: <boolean>) that makes the rule check if things follow the HSTS preload list requirements?

We are talking about HSTS in the docs so we should do something about it. Also IIRC max-age needs to be 2 years to be included in the list?

From the doc of this rule:

> same domain via the insecure connections, the browswer knows that it should never load the site using HTTP and automatically convert all attempts to HTTPS requests instead.

Does this means the browser will switch automatically to https? Then if we are in http at that point to we should trigger an error/warning. Also I'm not sure the jsdom collector does it.

molant:
The commit messages in this PR seem a bit weird. I'm seeing changes from other PRS and rules. Maybe it needs a rebase?

src/lib/rules/strict-transport-security/strict-transport-security.ts
qzhou1607: @molant @alrra Yes.  The HTTP requests will automatically be converted to HTTPS. Quote from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security:
>The first time your site is accessed using HTTPS and it returns the Strict-Transport-Security header, the browser records this information, so that future attempts to load the site using HTTP will automatically use HTTPS instead.

Shouldn't we just use the url after redirect? if it is `https`, check for the header, right? Does this mean jsdom doesn't do automatic redirect?
>Also I'm not sure the jsdom collector does it.

qzhou1607: @alrra @molant I tried with `http://developer.microsoft.com/en-us/microsoft-edge/`, using both `cdp` and `jsdom`, the url I got in `validate` is always `httpS://developer.microsoft.com/en-us/microsoft-edge/`.

Also the spec(https://tools.ietf.org/html/rfc6797#section-7.2) says:
>An HSTS Host MUST NOT include the STS header field in HTTP responses
   conveyed over non-secure transport.

So that means the header shouldn't be required if the url before redirect is `http`, right?

18 weeks. The "2 years" is the "recommended" length if the site is "committed" to `https`.
>Also IIRC max-age needs to be 2 years to be included in the list?

I think this is a good idea. Should I go ahead and add it?
>@molant @qzhou1607 Maybe we should have a config for this rule (something such as preloadList: <boolean>) that makes the rule check if things follow the HSTS preload list requirements?

molant: This is what I think this should rule should be checking for:

* Check that `http` doesn't send the STS header.
  * If initial url is http and redirects to https, this is fine
  * If initial url is https then should we check what happens with http? My vote is yes
*  Do the current checks you are doing

Add `preloadList` with a default to false to check for those values:

 * Should we download the list and see if the website has been already submitted?

We should also add a link to https://hstspreload.org/ in the documentation as it contains recommendations on how to add the domain to the list.

What do you think?

alrra: > So that means the header shouldn't be required if the url before redirect is http, right?

Yeap, and that is because over HTTP an attacker may intercept the connections and remove the header. 

Talking with @qzhou1607 it seems my comment was not clear. What I suggested was that even if the user provides a `HTTP` URL that, for some reason is not redirected to `HTTPS`,  we should still check if the site supports `HTTPS`, and if it does check for the header for `HTTPS`.

----

Basically: 

  * if it's `HTTP` - the presence of the header should be an error
  * if it's `HTTPS` - the lack of the header should be an error

----

> Add preloadList with a default to false to check for those values:

@qzhou1607 We should make it very clear in to docs that users should not do it if they are not 100% sure they can support `HTTPS` for their entire site and all its subdomains in the long run (see [my initial comment](https://github.com/sonarwhal/sonar/issues/23#issue-213013019)).

> Should we download the list and see if the website has been already submitted?

@molant Do you mean as in: do the tests as if `preloadList` was set to `true` if we encounter a site from that list?

molant: > Do you mean as in: do the tests as if preloadList was set to true if we encounter a site from that list?

That or flag an error if the preload option is true, passes everything but the site is not on the list.

qzhou1607: @molant @alrra  The preload list json file seems pretty large (https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json) and takes a while to load each time. How often should we download it?

alrra: > How often should we download it?

We can set up a cron job to fetch it, for example, every week.

qzhou1607:
@alrra @molant I add some more checks in the rule : 

- check `http` site for header, reports an error if found.
- check if a `http` site supports `https`, if yes, check for the header.
- check if the response header has `preload` and `includeSubdomains` attribute if it is included in the preload list.
- the `preload` check can be configured by the user.

The current `preload.json` will be replaced once https://github.com/sonarwhal/sonar/issues/386 is done.
I think this is all we have talked about to add if I remember it right?

qzhou1607:
Some more reading on `preload` today and our previous understanding about the preload submission process seems to be incorrect:
According to the following two blog posts:
https://scotthelme.co.uk/hsts-preload-test/
https://scotthelme.co.uk/hsts-preloading/

When a domain needs to be added to the preload list, two things need to happen:
* It includes `preload` attribute in the response header
* The domain name is submitted in https://hstspreload.org/

And once a domain is added to list, it CAN remove the `preload` attribute in the response header. So it is  acceptable if a domain is on the list, but doesn't have `preload` in the header. (e.g.,  stripe.com)

Additionally, there are two api endpoints that can be used to check a domain's `status(if it is on the preload list already)` and also `preloadable(if it is qualified to be added)` property:

https://hstspreload.org/api/v2/status?domain=DOMAIN
https://hstspreload.org/api/v2/preloadable?domain=DOMAIN

(The api endpoints are different from the blog posts, I think they were updated)

I think this is what we can do to the check the preload attribute :

* If a header doesn't have `preload` attribute => do nothing
* If a header has `preload` attribute => check if it is already on the list using the `status` endpoint:
   * If yes, do nothing
   * If no, check if the domain is qualified using the `preloadable` endpoint and report the errors

And we probably won't need to do #386 anymore. @alrra @molant , what do you think?

docs/user-guide/rules/strict-transport-security.md
```diff
@@ -0,0 +1,141 @@
...
+HTTP/... 200 OK
+
+...
+ Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
+```
+
+## Can the rule be configured?
+
+Yes, you can configure the value that `max-age` is checked against with. By default,
+this limit is set as 18 weeks(10886400s);
```
alrra: Add space before `(`.

docs/user-guide/rules/strict-transport-security.md
```diff
@@ -0,0 +1,141 @@
...
+
+```json
+"strict-transport-security": ["error", {
+    "minMaxAgeValue": 123456
+}]
+```
+
+Also, you can configure the rule so that the `preload` attribute in the header can be
+validated based on whether or not the domain being executed on was included in
+the preload list from chrome. This validation is disabled by default.
```
alrra: Use `Chrome` instead of `chrome`.

alrra: Also the list is used by other browsers, so we should specify them.

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,215 @@
...
+        // This function accomplishes the following:
+        // "max-age=31536000; includesubdomains; preload" => {"max-age":31536000,"includesubdomains":true,"preload":true}
+        const parse = (headerValue: string) => {
+            const parsedHeader = {};
+            const directives = headerValue.toLowerCase().split(';');
+            const nameValuePairRegex = /^ *([!#$%&'*+.^_`|~0-9A-Za-z-]+) *= *("(?:[~0-9])*"|[!#$%&'*+.^_`|~0-9]+) *$/;
+            // Regex for name-value pairs. E.g.: max-age=31536000
+            // Modified usage of https://github.com/jshttp/content-type/blob/64bde0d996ccb4334341662c0c7d25f7b370c4d9/index.js#L23
+            const tokenRegex = /^ *[!#$%&'*+.^_`|~0-9A-Za-z-]+$/;
+            // Regex for single tokens. E.g.:  includesubdomains
```
alrra: Leftover comment?

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,215 @@
...
+
+            if (!match && parsedHeader.preload) {
+                throw new Error(`This site is not in the preload list, but 'Strict-transport-security' header has 'preload' attribute`);
+            }
+        };
+
+        const validate = async (fetchEnd: IFetchEnd) => {
+            const { element, resource, response }: { element: IAsyncHTMLElement, resource: string, response: IResponse } = fetchEnd;
+
+            // This check does not make sense for data URI.
```
alrra: Also for `javascript:` and others (see: [this](https://github.com/sonarwhal/sonar/commit/28c782db16ca6140a083fae76f143a05f595694e#diff-3f3def15d620b7145d22d1a373aa59f0R87)).

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,215 @@
...
+                    if (networkData.response.statusCode === 200) {
+                        validate({
+                            element: null,
+                            request: networkData.request,
+                            resource: httpsResource,
+                            response: networkData.response
+                        });
+                    }
+                } catch (err) { // eslint-disable-line no-empty
+                    // HTTPS site can't be fetched, do nothing.
```
alrra: Add a `debug` statement here.

qzhou1607: @alrra This was added, not sure why it doesn't show.

qzhou1607:
The following check was done and tests were added.
> Additionally, there are two api endpoints that can be used to check a domain's `status(if it is on the preload list already)` and also `preloadable(if it is qualified to be added)` property:
> https://hstspreload.org/api/v2/status?domain=DOMAIN
>https://hstspreload.org/api/v2/preloadable?domain=DOMAIN

>* If a header doesn't have `preload` attribute => do nothing
>* If a header has `preload` attribute => check if it is already on the list using the `status` endpoint:
>    * If yes, do nothing
>    * If no, check if the domain is qualified using the `preloadable` endpoint and report the errors

qzhou1607:
I'm almost done with the new rule #24 to check `set-cookie` header in `https` pages. But the tests rely on the code in this PR (https test server) to work. Should I wait until this PR is merged? Or should I push it to this PR as a separate commit? @alrra @molant

molant:
@qzhou1607 if the code for the https server is in a separate commit maybe you can cherry pick it? If not maybe copy the code for now and we'll deal with it later. @alrra should be online on Monday and I'll have some time on Tuesday probably.

src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -597,7 +597,24 @@ export class Connector implements IConnector {
...
+                Security.certificateError(({ eventId }) => {
+                    Security.handleCertificateError({
+                        action: 'continue',
+                        eventId
+                    });
+                });
+
+                await Security.enable();
+
+                // enable the override
```
alrra: @qzhou1607 For consistency, start comment with uppercase letter and add `.` at the end (e.g. in this case: `Enable the override.`)

Note: I'll take care of these change when I merge this.

src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -597,7 +597,24 @@ export class Connector implements IConnector {
             }
```
alrra: @qzhou1607 Will change the commit message to use the [`Chore:`](https://sonarwhal.com/docs/developer-guide/contributing/pull-requests.html#commitmessages) tag as this is not a user-facing thing).

src/lib/connectors/shared/remote-debugging-connector.ts
qzhou1607: @alrra Sure. Do you need me to resolve the conflicts?

alrra: @qzhou1607 I already took care of that locally, so there is no need for this PR.

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,221 @@
...
+            return status === `preloaded`;
+        };
+
+        const issuesToPreload = (hostname: string): Promise<{ [key: string]: any }> => {
+            return requestJSONAsync(`${preloadableApiEndPoint}${hostname}`);
+        };
+
+        const verifyPreload = async (resource): Promise<{ [key: string]: any }> => {
+            const originalDomain = url.parse(resource).hostname;
+            const mainDomain = originalDomain.replace('www.', '');
```
alrra: @qzhou1607  `'testwww.com'.replace('www.', '');` => `testcom`

src/lib/utils/misc.ts
```diff
@@ -61,6 +67,26 @@ const normalizeString = (value: string, defaultValue?: string): string => {
     return value.toLowerCase().trim();
 };
 
+/** Convenience function to check if a uri's protocol is http/https if specified. */
+const hasRequiredProtocol = (uri: string): boolean => {
```
alrra: We should rename this to something else. Nothing comes to mind right now.

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,221 @@
...
+                    throw new Error(`'strict-transport-security' header contains more than one '${name}'`);
+                }
+
+                parsedHeader[name] = value || true;
+            });
+
+            return parsedHeader;
+        };
+
+        const isInvalidValue = (maxAge: string, limit: number): boolean => {
```
alrra: Maybe rename this to something such as `isUnderLimit`?

src/lib/rules/strict-transport-security/strict-transport-security.ts
```diff
@@ -0,0 +1,221 @@
...
+
+        const verifyPreload = async (resource): Promise<{ [key: string]: any }> => {
+            const originalDomain = url.parse(resource).hostname;
+            const mainDomain = originalDomain.replace('www.', '');
+            const hostInTheListAlready = await isPreloaded(mainDomain) || await isPreloaded(originalDomain);
+            // Some hostnames in the list include `www.`, e.g., `www.gov.uk`.
+
+            let issues = {};
+
+            if (!hostInTheListAlready) {
```
alrra: If the API cannot be reached (e.g.: the service is down, the user is not connected to the internet), should the test fail? 

Cc @sonarwhal/contributors

src/lib/rules/strict-transport-security/strict-transport-security.ts
molant: isUnderAge?

src/lib/rules/strict-transport-security/strict-transport-security.ts
molant: I think we do something similar with the ssllabs tests. IIRC we print a warning but it might be lost with all the other errors/warnings

src/lib/utils/misc.ts
molant: isWebProtocol?

alrra: `isRegularProtocol`?

src/lib/rules/strict-transport-security/strict-transport-security.ts
qzhou1607: @alrra How about `const mainDomain = originalDomain.replace(/^www./, '')`? Do you need me to make changes in this PR?

By the way, both the original and main domain will be tested against the list. See:
```
const hostInTheListAlready = await isPreloaded(mainDomain) || await isPreloaded(originalDomain);
```

src/lib/rules/strict-transport-security/strict-transport-security.ts
qzhou1607: +1 for printing warnings. @alrra Again, do you need me to make changes in this PR?

alrra: > @alrra Again, do you need me to make changes in this PR?

Yes, you can do them, at first I thought everything was ready, but then I realized we need to change some other things.

src/lib/rules/strict-transport-security/strict-transport-security.ts
alrra: > How about const mainDomain = originalDomain.replace(/^www./, '')?

I think that would work.

src/lib/rules/strict-transport-security/strict-transport-security.ts
qzhou1607: @alrra @molant Handling of service failure added. I wanted to add some tests with service failure mocked, but then realized that we don't have access to port number in tests, and thus the messages are never right. So I left those code commented out in case we will have access to port number in tests in the future.

PR #400, opened 2017-07-31
Fix: Fix a bug in `rule-generator`

### LINKED ISSUE #392: Rule-generator not rendering quotes
When using the rule-generator, if the `description` contains quotes, it doesn't display right in the generated files:
`This rule checks if the &#x60;Set-Cookie&#x60; header is sent with  &#x60;Secure&#x60; and &#x60;HttpOnly&#x60; Directive over  HTTPS.`

qzhou1607:
* Fix the bug that the rule generator can't take quoted strings.
* Add tests

Fix #392

### DIFF
--- a/src/lib/cli/rule-generator.ts
+++ b/src/lib/cli/rule-generator.ts
@@ -60,7 +60,7 @@ export type NewRule = {
     /** Category of the new rule */
     category: string,
     /** Description of the new rule */
-    description: string,
+    description: hbs.SafeString,
     /** Element type if `dom` is selected in useCase */
     elementType?: string,
     /** Extension of the new rule file*/
@@ -71,6 +71,20 @@ export type NewRule = {
     isCore: boolean,
     /**  Usage categories that the new rule applies to */
     useCase: UseCase
+};
+
+/**
+ * Use `escapeSafeString` function instead of triple curly brace in the templates
+ * to escape the backticks (`) in the user's input.
+ * Example:
+ * ```
+ * description: `This is a \`important\` rule that has 'single' and "double" quotes.`
+ * ```
+ */
+const escapeSafeString = (str: string): hbs.SafeString => {
+    const result = str.replace(/(`)/g, '\\$1');
+
+    return new Handlebars.SafeString(result);
 };
 
 /** Loads a template from a `filePath` and interpolates the values with the given `data`. */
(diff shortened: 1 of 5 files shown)


src/lib/cli/rule-generator.ts
```diff
@@ -60,7 +60,7 @@ export type NewRule = {
     /** Category of the new rule */
     category: string,
     /** Description of the new rule */
-    description: string,
+    description: object,
```
molant: object is too generic. What is the type returned by Handlebars.SafeString?

qzhou1607: @molant  It returns an object that has a property of `string`. I edited the commit and added a `SafeString` type to be more specific. But unless I make this `string` property optional in the type, an error of ` Type 'hbs.SafeString' is not assignable to type 'SafeString'` always shows. Any suggestion on better solving that?

qzhou1607: Just figured it out. Please ignore my last comment.

src/lib/cli/templates/rule-script-ts.hbs
```diff
@@ -70,8 +69,8 @@ const rule: IRuleBuilder = {
     },
     meta: {
         docs: {
-            category: '{{category}}',
-            description: '{{description}}'
+            category: `{{category}}`,
```
molant: @sarvaje shouldn't this be a triple curly brace?

sarvaje: The triple curly brace `{{{variable}}}` avoid to escape the value of the variable.
Using the triple curly brace, maybe you don't need to use the function `escapeSafeString`

qzhou1607: @sarvaje @molant  The reason of using `escapeSafeString` function instead of triple curly brace is that the backticks (`) need to be escaped by adding a backslash, otherwise the following might be generated if only using the triple curly braces:
```
...
description: `This is a `important` rule.`
...
```

sarvaje: ahh, ok!

molant: Then there should be a comment somewhere explaining this 😊

qzhou1607: @molant Done.

alrra:
Note: Changed the commit message a bit.

PR #420, opened 2017-08-11
New: Added Chrome Traces to cdp connector

axemclion:
Chrome Traces has much more information about performance and
rules can use the new events - tracing::data and tracing::end
for trace information.

This information can be used for new rules like firstPaint,
total time taken to parseHTML, long frames, rendering, etc.

### DIFF
--- a/src/lib/connectors/shared/remote-debugging-connector.ts
+++ b/src/lib/connectors/shared/remote-debugging-connector.ts
@@ -521,7 +521,7 @@ export class Connector implements IConnector {
     /** Initiates all the process */
     private onLoadEventFired(callback: Function): Function {
         return () => {
-            const { DOM } = this._client;
+            const { DOM, Tracing } = this._client;
             const event: IEvent = { resource: this._finalHref };
 
             setTimeout(async () => {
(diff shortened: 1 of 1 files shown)


alrra:
@axemclion Thanks for the PR! 

Can you [sign the CLA](https://github.com/sonarwhal/sonar/pull/420#issuecomment-321831363)? Thanks!

----

Cc: @molant @sarvaje

panarasi:
Sorry, I think I sent the PR using a wrong account. I work for Microsoft.

molant:
@panarasi you still need to sign the CLA. The code now belongs to the JS Foundation 😊

sarvaje:
@panarasi the build is failing because of some `eslint` errors

axemclion:
@sarvaje  Fixed eslint issues.

axemclion:
Also in [line 551](https://github.com/sonarwhal/sonar/pull/420/files#diff-367932743ebd6416c02c82d8f4249581R551), I am not using reject since the whole block is wrapped inside a `setTimeout` anyway. Is that ok ?

src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -548,6 +548,15 @@ export class Connector implements IConnector {
...
                     }
 
+                    await new Promise((resolve) => {
+                        Tracing.tracingComplete(() => {
+                            this._server.emitAsync('tracing::end', event);
+                            resolve();
+
+                        });
+                        debug(`Ending Collecting Chrome traces`);
+                        Tracing.end();
```
sarvaje: Can `Tracing.end` throw an exception? if so, you can wrap it in a `try/catch` and call to `reject` in the `catch`

axemclion: I think Tracing.end would throw just as Page.navigate or other chrome commands would. This just sends the command over to chrome over the protocol, so IMHO, it should be fine.

molant:
Closing this as there's more work that needs to happen to enable it in all connectors but definitely something we want to add in the future.

PR #423, opened 2017-08-14
New: Notify users if the current version of sonar is not up to date

### LINKED ISSUE #419: Notify users if there is a newer version of sonar
This should be the first task the cli does. Something similar to what npm does is fine. Ideally it should link to the changelog as well.

qzhou1607:
Fix #419

Currently the message looks as below:

new-version-notification

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -285,6 +285,12 @@
       "integrity": "sha512-Ei8DBGk7P2BNGMglIqnFJKYY//0pKV1ctXCbf+hjdVF/2wxYgfuUSZOonVrYu3sy8wToKtnDQWE5rSpNgWpmrw==",
       "dev": true
     },
+    "@types/update-notifier": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@types/update-notifier/-/update-notifier-1.0.1.tgz",
+      "integrity": "sha1-R+jwB8a1vK++QnMy47a0M83Bjdw=",
+      "dev": true
+    },
     "@types/user-home": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/@types/user-home/-/user-home-2.0.0.tgz",
@@ -385,7 +391,6 @@
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/ansi-align/-/ansi-align-2.0.0.tgz",
       "integrity": "sha1-w2rsy6VjuJzrVW82kPCx2eNUf38=",
-      "dev": true,
       "requires": {
         "string-width": "2.1.1"
       },
@@ -393,20 +398,17 @@
         "ansi-regex": {
           "version": "3.0.0",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.0.tgz",
-          "integrity": "sha1-7QMXwyIGT3lGbAKWa922Bas32Zg=",
-          "dev": true
+          "integrity": "sha1-7QMXwyIGT3lGbAKWa922Bas32Zg="
         },
         "is-fullwidth-code-point": {
           "version": "2.0.0",
           "resolved": "https://registry.npmjs.org/is-fullwidth-code-point/-/is-fullwidth-code-point-2.0.0.tgz",
-          "integrity": "sha1-o7MKXE8ZkYMWeqq5O+764937ZU8=",
-          "dev": true
+          "integrity": "sha1-o7MKXE8ZkYMWeqq5O+764937ZU8="
         },
         "string-width": {
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/string-width/-/string-width-2.1.1.tgz",
           "integrity": "sha512-nOqH59deCq9SRHlxq1Aw85Jnt4w6KvLKqWVik6oA9ZklXLNIOlqg4F2yrT1MVaTjAqvVwdfeZ7w7aCvJD7ugkw==",
-          "dev": true,
           "requires": {
             "is-fullwidth-code-point": "2.0.0",
             "strip-ansi": "4.0.0"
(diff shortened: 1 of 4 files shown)


src/lib/cli.ts
```diff
@@ -61,6 +69,22 @@ const setUpUserFeedback = (sonarInstance: Sonar, spinner: IORA) => {
...
+    current: string,
+    type: string,
+    name: string
+};
+
+const notifyNewVersion = (update: Update) => {
+    const changelogUrl: string = `https://sonarwhal.com/about/changelog.html`;
+    // No indentation due to the use of `\` to avoid new line.
+    // https://stackoverflow.com/a/35428171
+    const message: string = `Update available ${chalk.dim(update.current)}${chalk.reset(' → ')}${chalk.green(update.latest)}\nRun ${chalk.cyan(`npm i -g ${update.name}`)} to update\nSee ${chalk.red(changelogUrl)} for update details`;
```
alrra: `sonar` may be installed as a dependence and run via an `npm` script, so suggesting `npm i -g` is not accurate in that case.

package-lock.json
```diff
@@ -380,28 +380,24 @@
       "version": "2.0.0",
```
alrra: These `package-lock.json` changes fluctuate so much. :(

src/lib/cli.ts
qzhou1607: @alrra If we have a label in the code that indicates whether or not sonar is installed as a dependency, I can generate different messages based on different scenarios. Does such "label" exist? Otherwise I can include both `npm i -g` and `npm i -D` options in the message and ask the user to pick according to their own use.

package-lock.json
qzhou1607: @alrra Is there a good way to avoid fluctuation?

src/lib/cli.ts
alrra: > Otherwise I can include both npm i -g and npm i -D o

[`-P |  --save-prod`](https://docs.npmjs.com/cli/install) is also possible. 

> Otherwise I can include both `npm i -g` and `npm i -D` options in the message and ask the user to pick according to their own use.

Not all user will know the differences, so if we do that we will need to also explain the differences between the options, and IMHO that is not the way to go.

My opinion is that the less we asked from the user the better.

package-lock.json
alrra: @qzhou1607 I was just making a remark, nothing to do with this PR.

src/lib/cli.ts
qzhou1607: @alrra How about using [detect-installed](https://github.com/charlike/detect-installed) to see if sonar was installed globally first. If yes, then add the `-g` flag in there?

molant: I would just say "please update" or similar. Don't think it is worth pulling another dependency just for that.

qzhou1607:
@molant @alrra The message is changed as shown below:

update available

tests/lib/cli.ts
```diff
@@ -237,3 +251,33 @@ test.serial('Event scann::end should write a message in the spinner', async (t)
...
+    await cli.execute('http://localhost/');
+
+    t.true(t.context.notifier.notify.calledOnce);
+    t.is(t.context.notifier.notify.args[0][0].message, expectedMessage);
+});
+
+test.serial(`User shouldn't be notified if the current version is up to date`, async (t) => {
+    let undefinedUpdate;
+
+    t.context.notifier.update = undefinedUpdate;
```
sarvaje: I think you should do directly `t.context.notifier.update = void 0` instead of use a variable.

qzhou1607:
@sarvaje Done.

src/lib/cli.ts
```diff
@@ -61,6 +69,23 @@ const setUpUserFeedback = (sonarInstance: Sonar, spinner: IORA) => {
     });
 };
 
+type Update = { // eslint-disable-line no-unused-vars
```
molant: It's a bit weird to have this type here. Isn't there a type package for `update-notifier`?
If not I think the type should be declared elsewhere. @sarvaje thoughts?

sarvaje: It looks like there is a package `@types/update-notifier`.
Anyway, I think that all types should go to the folder `/types`

molant: Then we should use that package instead of doing this. @qzhou1607 can you please add it as a dev dependency?

qzhou1607: @molant @sarvaje Done.

package.json
```diff
@@ -12,6 +12,7 @@
   },
   "bin": "./dist/src/bin/sonar.js",
   "dependencies": {
+    "@types/update-notifier": "^1.0.1",
```
molant: This should be with the other types under dev dependencies.

src/lib/cli.ts
```diff
@@ -61,6 +69,16 @@ const setUpUserFeedback = (sonarInstance: Sonar, spinner: IORA) => {
     });
 };
 
+const notifyNewVersion = (update: updateNotifier.UpdateInfo) => {
+    const changelogUrl: string = `https://sonarwhal.com/about/changelog.html`;
+    // No indentation due to the use of `\` to avoid new line.
+    // https://stackoverflow.com/a/35428171
+    const message: string = `Update available ${chalk.red(update.current)}${chalk.reset(' → ')}${chalk.green(update.latest)}\
+\nPlease update to the newest version ${chalk.green(update.latest)}\nSee ${chalk.cyan(changelogUrl)} for details`;
```
alrra: Do we need both `Update available...` and `Please update to the newest version`? To me they basically specify the same thing.

molant:
@qzhou1607 or @alrra can you please take care of the conflicts?

qzhou1607:
@molant @alrra Done.

package-lock.json
alrra: From http://blog.npmjs.org/post/167963735925/v560-2017-11-27

> bc263c3fd  #19054  Fully cross-platform package-lock.json. Installing a failing optional  dependency on one platform no longer removes it from the dependency tree,  meaning that package-lock.json should now be generated consistently across  platforms! 🎉

molant: Update your npm install now 🎉

Issue #250, opened 2017-06-08
Allow connector opt-out for individual tests


molant:
Right now we can ignore a rule from being run and/or tested via the `ignoredCollectors` properties.
Sometimes the differences are more granular and there are just some scenarios when we are testing that do not work on a specific collector and maybe it is just a question of waiting to the next release. Instead of just commenting the test (and thus not test anywhere) we should add the option to opt-out for an specific collector.
An example will be the `content-type` tests. Currently Chrome doesn't download JavaScript modules (it's behind a flag in v60). Also it doesn't download anything if it doesn't like the [`type` attribute](https://github.com/MicrosoftEdge/Sonar/pull/245#discussion_r120083650)

We should also add a `//TODO` comment so we remove it later (hopefully once #210 is implemented so things are more automated).

I'm thinking about something similar to:
```js
 {
        name: `Script is served with 'Content-Type' header with the wrong media type (has 'type=text/plain' and 'js' file extension)`,
        reports: [{ message: generateIncorrectMediaTypeMessage('application/javascript', 'text/plain') }],
        serverConfig: {
            '/': generateHTMLPageData(generateHTMLPage(undefined, '<script type="text/plain" src="test.js"></script>')),
            '/test.js': { headers: { 'Content-Type': 'text/plain; charset=utf-8' } }
        },
        ignoredCollectors: ['cdp']
    },
```
@sarvaje I don't know how you are implementing the Edge Collector (#170) but if we want this to work it will need a different id and not just a config change.

@MicrosoftEdge/sonar-devs thoughts?

molant:
After thinking a bit more, I don't think we need this. It doesn't happen that often and when we have problems we should group the failing tests and just use the regular `ignoredConnectors` property of `testRule`:

```js

const tests = [...];
const testsNotRunningOnConnector = [...];

ruleRunner.testRule(ruleName, tests);
ruleRunner.testRule(ruleName, testsNotRunningOnConnector, { ignoredConnectors: ['connector']});
```

PR #425, opened 2017-08-15
Breaking: Make `content-type` rule use proper fonts types

alrra:
The `content-type` rule uses [`mime-db` which currently doesn't contain](https://github.com/jshttp/mime-db/issues/77) the [standard font media types]( https://www.iana.org/assignments/media-types/media-types.xhtml#font).

This commit fixes that by introducing code that overwrites the non-standard font media types with the standard ones.


alrra:
> Is there an issue on jsdom to download fonts?

@molant See https://github.com/sonarwhal/sonar/issues/71#issuecomment-306735295.

> Didn't we had a way to exclude connectors from tests? 

Note yet, see #250 (we can only [ignore connectors](https://sonarwhal.com/docs/developer-guide/rules/index.html#ignoreconnectors) as a whole).

PR #429, opened 2017-08-15
New: Propose to create a new config file if it doesn't exist

### LINKED ISSUE #354: If no .sonarrc file found, propose user to create a new one
Some user will most probably just install and do `sonar <url>`, and will get the following error: 

```sh
TypeError: Path must be a string. Received null
    at assertPath (path.js:28:11)
    at Object.resolve (path.js:1167:7)
    at Object.exports.load (/Users/alrra/.nvm/versions/node/v8.1.3/lib/node_modules/@sonarwhal/sonar/dist/src/lib/config.js:87:31)
    at Object.exports.execute (/Users/alrra/.nvm/versions/node/v8.1.3/lib/node_modules/@sonarwhal/sonar/dist/src/lib/cli.js:94:27)
    at run (/Users/alrra/.nvm/versions/node/v8.1.3/lib/node_modules/@sonarwhal/sonar/dist/src/bin/sonar.js:28:34)
    at Object.<anonymous> (/Users/alrra/.nvm/versions/node/v8.1.3/lib/node_modules/@sonarwhal/sonar/dist/src/bin/sonar.js:31:1)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
```

This constitutes a bad user experience, and I think we should change things so that, if no configurations are found, `sonar` will just use it's defaults (+ maybe also print a message, such as: `Configurations not found, using defaults...`).

----

Thoughts @sonarwhal/contributors?

qzhou1607:
Fix #354

### DIFF
--- a/src/lib/cli.ts
+++ b/src/lib/cli.ts
@@ -18,6 +18,7 @@ import * as path from 'path';
 import * as chalk from 'chalk';
 import * as ora from 'ora';
 import * as updateNotifier from 'update-notifier';
+import * as inquirer from 'inquirer';
 
 import * as Config from './config';
 import { debug as d } from './utils/debug';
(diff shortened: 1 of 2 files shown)


src/lib/cli.ts
```diff
@@ -61,6 +62,53 @@ const setUpUserFeedback = (sonarInstance: Sonar, spinner: IORA) => {
...
+const loadOrCreateIfNotExits = async () => {
+    let config: IConfig;
+    let configPath: string;
+
+    try {
+        configPath = Config.getFilenameForDirectory(process.cwd());
+        debug(`Loading configuration file from ${configPath}.`);
+        config = Config.load(configPath);
+    } catch (err) {
+        if (err.message !== `Couldn't find any valid configuration`) {
```
sarvaje: I don't like the idea to compare the `err.message`, if some day someone change the message, this code is not going to work as expected. However, I think we should offer them the option to create a new config file whatever the problem was. So this `if` is not necessary. (if so, maybe you need to change the `inquirer` message

qzhou1607:
@sarvaje Done.

src/lib/cli.ts
```diff
@@ -61,6 +62,41 @@ const setUpUserFeedback = (sonarInstance: Sonar, spinner: IORA) => {
     });
 };
 
+const confirmLaunchInit = () => {
+    debug(`Initiating launch init confirm.`);
+
+    const question: Array<object> = [{
+        message: `Configuration file doesn't exist or error in loading. Do you want to create a new config file?`,
```
sarvaje: Maybe we can have something like `Unable to find a valid configuration file` or something similar instead of `doesn't exist or error` .

What do you thing?

src/lib/cli.ts
```diff
@@ -136,6 +177,8 @@ export const execute = async (args: string | Array<string> | Object): Promise<nu
         }
     };
 
+    logger.log(`Start scanning...`);
```
molant: Don't think we need this output here.

qzhou1607: @molant Done.

qzhou1607:
@alrra Conflicts were resolved.

PR #430, opened 2017-08-16
Fix: Avoid analyzing `/favicon.ico` multiple times

### LINKED ISSUE #427: Don't analyze the `/favicon.ico` multiple times
Site: example.com

```text
http://example.com/favicon.ico

Warning  'content-type' header should have media type 'image/x-icon' (not 'text/html')  content-type
Warning  'server' header is disallowed                                                  no-disallowed-headers
Warning  'x-content-type-options' header was not specified                              x-content-type-options
Warning  'content-type' header should have 'charset=utf-8'                              content-type
Warning  'content-type' header should have media type 'image/x-icon' (not 'text/html')  content-type
Warning  'server' header is disallowed                                                  no-disallowed-headers
Warning  'x-content-type-options' header was not specified                              x-content-type-options
Warning  'content-type' header should have 'charset=utf-8'                              content-type
✖ Found 0 errors and 8 warnings
```

qzhou1607:
`CDP` already sends request to `/favicon.ico` automatically.
So when forcing to load favicon, we shouldn't send request
to `favicon.ico` again.

Fix #427


src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -488,7 +488,11 @@ export class Connector implements IConnector {
      * * uses `favicon.ico` and the final url after redirects.
      */
     private async getFavicon(element?: AsyncHTMLElement) {
-        const href = element ? element.getAttribute('href') : '/favicon.ico';
+        if (!element || !element.getAttribute('href')) {
```
alrra: There is also the case when `href` is empty.

alrra: Also, I don't think `getFavicon` should be responsible of checking these things, and instead, and at this point:

https://github.com/sonarwhal/sonar/blob/941d439affca16e3af1b0df90e739ee746df2313/src/lib/connectors/shared/remote-debugging-connector.ts#L329-L331

all the checks should be made to determine if valid `favicon` was specified or not.

qzhou1607: @alrra If a page doesn't have favicon in the root directory and `/favicon.ico` returns 404. Do we still want to list all the warnings/errors under this url?

alrra: > Do we still want to list all the warnings/errors under this url?

@qzhou1607 Sorry, I don't quite understand the question. Can you be more specific, thanks!

qzhou1607: @alrra Like http://www.example.com/favicon.ico returns 404. In the output of running sonar against http://www.example.com, we listed:
```
http://www.example.com: 1 issues
...
http://www.example.com/favicon.ico: 1 issues
...
```
But http://www.example.com/favicon.ico doesn't exist, it looks a little weird to still check its response.

alrra: > But http://www.example.com/favicon.ico doesn't exist, it looks a little weird to still check its response.

Yes. However, do make sure that `CDP` aligns with `jsdom`.

qzhou1607:
@alrra @sarvaje I pushed up some more changes. I've talked to you both about this issue and please let me know your thought about the current solution. Here is a summary:

The cause of the duplicate messages : `CDP` requests favicon from the root directory automatically if no favicon is present in the html. In our previous code we didn't set the `_faviconLoaded` flag to true in this scenario, so `getFavicon` got called and `fetch::end` event was emitted twice.

Changes in the current solution :
1. If `cdp` has sent request to `/favicon.ico` already, set the `_faviconLoaded` flag to true as well.
2. `getFavicon` doesn't need to request `/favicon.ico`. This function should have never been reached in the first place if favicon is in the root directory.
3. `getFavicon` shouldn't include the code to check if the element exists. If the element doesn't exist, it shouldn't reach `getFavicon`.

src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -326,7 +326,12 @@ export class Connector implements IConnector {
             };
         }
 
-        if (hasAttributeWithValue(data.element, 'link', 'rel', 'icon')) {
+        const hasFaviconLinkTag = hasAttributeWithValue(data.element, 'link', 'rel', 'icon');
+        // If favicon is not present in the html, CDP will send a request to get favicon from the root directory.
+        // We should set the tag to true in this case to avoid fetching favicon twice.
+        const hasFaviconInRootDir = data.request.url.endsWith('/favicon.ico');
```
molant: I don't think this is right. I could very well have the following url `https://sonarwhal.com/images/favicon.ico` and the favicon will not be in the root dir.
We should add a test for this case too.

sarvaje: @molant I think the check is ok, if we receive in the response the `favicon.ico` it doesn't matter the route, the favicon.ico is already download and we won't need to download it manually

molant: Then the name isn't appropriate.

sarvaje: I was talking with Qing, and you are right, if it isn't in the root, `hasFaviconLinkTag` should be true, she is going to change how to calculate `hasFaviconInRootDir`

qzhou1607: @molant I think your comment is correct. There is a scenario that `/images/favicon.ico` is included in the html as the resource a `img` instead of a `link` tag in the head (`rel` set as `icon`). In this scenario we shouldn't set the flag to true since the favicon image isn't included in the page in an appropriate way. We should still go into `getFavicon` trying to manually download the favicon.

qzhou1607: I will work on adding more tests too :)

qzhou1607:
@molant @alrra  I added some tests to make sure that there are no duplicate requests to favicon, but some of them are failing in `cdp` due to a difference in the `headless` mode - It seems that `cdp` in the headless mode doesn't send requests to `/favicon.ico` automatically. Should I skip these tests for `cdp` too?

molant:
> It seems that cdp in the headless mode doesn't send requests to /favicon.ico automatically
Then I think we should download it. If by the end of the traversing nothing has happened with should manually download it.

qzhou1607:
I modify the code and use an approach suggested by @sarvaje before: Ignore automatic requests(and responses) from `cdp` to `/favicon.ico` to avoid emitting duplicates events. This way favicon is downloaded only once from the root and the tests work in both the headless and regular `cdp`. Ready for review @sarvaje @molant @alrra  .

src/lib/connectors/jsdom/jsdom.ts
```diff
@@ -226,7 +226,7 @@ class JSDOMConnector implements IConnector {
      * * uses `favicon.ico` and the final url after redirects.
      */
     private async getFavicon(element?: HTMLElement) {
-        const href = element ? element.getAttribute('href') : '/favicon.ico';
+        const href = (element && element.getAttribute('href')) || '/favicon.ico';
```
sarvaje: Any reason to do assignment in this way instead of the old way?

In terms of readability I prefer the old way.

src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -487,8 +522,8 @@ export class Connector implements IConnector {
      * * uses the `src` attribute of `<link rel="icon">` if present.
      * * uses `favicon.ico` and the final url after redirects.
      */
-    private async getFavicon(element?: AsyncHTMLElement) {
-        const href = element ? element.getAttribute('href') : '/favicon.ico';
+    private async getFavicon(element: AsyncHTMLElement) {
+        const href = (element && element.getAttribute('href')) || '/favicon.ico';
```
sarvaje: same as before

tests/lib/connectors/collect.ts
```diff
@@ -0,0 +1,131 @@
...
+    test.serial(`[${name}] Favicon is present in a 'link' element with 'rel' attribute set to 'icon' `, async (t) => {
+        const faviconInLinkElementDir = `http://localhost:${t.context.server.port}/images/favicon-32x32.png`;
+        const serverConfig = {
+            '/': generateHTMLPage(`<link rel="icon" type="image/png" href="/images/favicon-32x32.png" sizes="32x32">`),
+            '/images/favicon-favicon-32x32.png': fs.readFileSync(pathToFaviconInLinkElement)
+        };
+
+        await runTest(t, connectorBuilder, serverConfig);
+
+        t.plan(2);
```
sarvaje: `t.plan` is not necessary here. Take a look to [this](https://github.com/avajs/ava/blob/master/docs/recipes/when-to-use-plan.md)

It makes sense when you need to be sure that a block of code with `t.is` (or another test) is run.
For example:
```
t.plan(2);
try {
    await runTest(t, connectorBuilder, serverConfig);
    t.is(t.context.sonar.emitAsync.withArgs('fetch::end').callCount, 1);
    t.is(t.context.sonar.emitAsync.withArgs('fetch::end').args[0][1].request.url, faviconInLinkElementDir);
}catch(e){}
```

src/lib/connectors/jsdom/jsdom.ts
qzhou1607: @sarvaje This is due to the scenario when `href` is an empty string. We still want to send out a request to `/favicon.ico` in this case.

sarvaje: Ahhh, ok!

tests/lib/connectors/collect.ts
qzhou1607: @sarvaje Done. Thanks for the information!

PR #434, opened 2017-08-18
Chore: Changes to support edge connector

### LINKED ISSUE #432: Need to add type definitions to npm package so we can create external rule/connector
If we try to create an external rule/connector, we don't have the type definitions, so typescript notifies some errors when they are not.

For example, if we try to create a new `Launcher` for a connector, I have to do something like:
`export class NewLauncher exports Launcher`

`Launcher` has a protected property `port`, but if I try to use that property inside `NewLauncher` I get a typescript error `error TS2339: Property 'port' does not exist on type 'NewLauncher'.`

sarvaje:
This PR has to be merged to make https://github.com/sonarwhal/edge15-connector works

Also, it would be great to have fixed #432 so we can change [this code](https://github.com/sonarwhal/edge15-connector/blob/master/src/lib/connectors/edge15-connector/edge15-connector-launcher.ts#L42)

### DIFF
--- a/src/lib/connectors/shared/async-html.ts
+++ b/src/lib/connectors/shared/async-html.ts
@@ -29,6 +29,12 @@ export class AsyncHTMLDocument implements IAsyncHTMLDocument {
         });
     }
 
+    private getHTMLChilden(children: Array<any>) {
+        return children.find((item) => {
+            return item.nodeType === 1 && item.nodeName === 'HTML';
+        });
+    }
+
     // ------------------------------------------------------------------------------
     // Public methods
     // ------------------------------------------------------------------------------
@@ -48,7 +54,15 @@ export class AsyncHTMLDocument implements IAsyncHTMLDocument {
     }
 
     public async pageHTML(): Promise<string> {
-        const { outerHTML } = await this._DOM.getOuterHTML({ nodeId: this._dom.nodeId });
+        let { outerHTML } = await this._DOM.getOuterHTML({ nodeId: this._dom.nodeId });
+
+        // Some browsers like Edge don't have the property outerHTML in the root element
+        // so we need to find the html element
+        if (!outerHTML) {
+            const htmlElement = this.getHTMLChilden(this._dom.children);
+
+            ({ outerHTML } = await this._DOM.getOuterHTML({ nodeId: htmlElement.nodeId }));
+        }
 
         return outerHTML;
     }
--- a/src/lib/connectors/shared/remote-debugging-connector.ts
+++ b/src/lib/connectors/shared/remote-debugging-connector.ts
@@ -19,6 +19,7 @@ import * as r from 'request';
 import { AsyncHTMLDocument, AsyncHTMLElement } from '../shared/async-html';
 import { debug as d } from '../../utils/debug';
 import { delay, hasAttributeWithValue } from '../../utils/misc';
+import { resolveUrl } from '../utils/resolver';
 
 /* eslint-disable no-unused-vars */
 import {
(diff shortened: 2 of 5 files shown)


src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -73,6 +74,7 @@ export class Connector implements IConnector {
         const defaultOptions = {
             loadCompleteRetryInterval: 250,
             maxLoadWaitTime: 30000,
+            newTabUrl: 'http://sonarwhal-empty-site.azurewebsites.net/',
```
alrra: @sarvaje Maybe add a bit of explanation why this URL is used, who owns it, etc. Thanks!

molant: It will actually be empty.sonarwhal.com once I'm back full time 😊

sarvaje: >It will actually be empty.sonarwhal.com once I'm back full time 😊

I'll change the url.

>@sarvaje Maybe add a bit of explanation why this URL is used, who owns it, etc. Thanks!

ok!

src/lib/connectors/shared/async-html.ts
```diff
@@ -48,7 +54,15 @@ export class AsyncHTMLDocument implements IAsyncHTMLDocument {
     }
 
     public async pageHTML(): Promise<string> {
-        const { outerHTML } = await this._DOM.getOuterHTML({ nodeId: this._dom.nodeId });
+        let { outerHTML } = await this._DOM.getOuterHTML({ nodeId: this._dom.nodeId });
+
+        // Some browsers like Edge doesn't have the property outerHTML in the root element
```
molant: > don't

src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -73,6 +74,10 @@ export class Connector implements IConnector {
         const defaultOptions = {
             loadCompleteRetryInterval: 250,
             maxLoadWaitTime: 30000,
+            // tabUrl is a empty html site used to avoid edge diagnostics adapter to receive unexpeted onLoadEventFired
+            // and onRequestWillBeSent events from the default url opened when you create a new tab in Edge.
+            tabUrl: 'http://empty.sonarwhal.com/',
```
molant: Let's use `https://empty.sonarwhal.com`
I'll configure the SSL certificate once the DNS records are working.

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -78,7 +78,7 @@ const rule: IRuleBuilder = {
                when testing the rule and `import` doesn't work here. */
             const htmlChecker = require('html-validator');
 
-            scanOptions.data = response.body.content;
+            scanOptions.data = response.body.rawContent.toString();
```
molant: I don't think this change is related to edge 15. Maybe put it as a different commit?

src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -338,6 +341,52 @@ export class Connector implements IConnector {
...
+            await this._server.emitAsync('manifestfetch::error', event);
+        }
+    }
+
+    private async getManifest(element: IAsyncHTMLElement) {
+        this._manifestIsSpecified = true;
+
+        try {
+            // CDP will not download the manifest on its own, so we have to "force" it
+            // This will trigger the `onRequestWillBeSent`, `OnResponseReceived`, and
```
molant: > `onResponseReceived`

src/lib/rules/html-checker/html-checker.ts
sarvaje: It is related to edge15. IIRC, rawContent in edge15 is in base64 and a string in chrome

molant: What? Then you should put a comment about this.

sarvaje:
@molant everything should be done!

molant:
@alrra do you want any other changes or can this be merged?

PR #448, opened 2017-08-22
Build: Generate files `d.ts`

sarvaje:
Related #432

### DIFF
--- a/tests/helpers/test-server.ts
+++ b/tests/helpers/test-server.ts
@@ -11,13 +11,13 @@ import * as express from 'express';
 import * as onHeaders from 'on-headers';
 import { promisify } from 'util';
 
-type ServerConfiguration = string | object; //eslint-disable-line
+export type ServerConfiguration = string | object; //eslint-disable-line
 
 const startPort = 3000;
 const maxPort = 65535;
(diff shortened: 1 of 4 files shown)


tests/helpers/test-server.ts
```diff
@@ -166,10 +166,8 @@ class Server {
 }
 
 /** Returns a testing server */
-const createServer = (isHTTPS?: boolean) => {
-    const server = new Server(isHTTPS);
+export const createServer = (isHTTPS?: boolean) => {
+    const a = new Server(isHTTPS);
```
molant: What about doing just a `return new Server(isHTTPS);`? Not a huge fan of `a`

molant:
@alrra, @sarvar is the `.d.ts` file going to be included directly on the npm package with this configuration or is there something else that needs to be done?

tests/helpers/test-server.ts
sarvaje: Ups, I forgot to restore the name to `server`... but yes, maybe is better just do a return.

sarvaje: @molant done!

PR #449, opened 2017-08-23
New: Add rule to check the viewport meta tag

### LINKED ISSUE #82: Add rule to check the usage of `<meta>` viewport
- [x] Check for if the <meta> tag is included, and has `width=device-width`.
   
    Note: For iOS 9+ `initial-scale=1` is no longer needed ([1](https://twitter.com/ppk/status/829329567219343360), [2](https://twitter.com/firt/status/829335091142918145))

- [x] Check for values that create bad user experience such as `user-scalable=no`.
- [x] Check for values that are ignored such as `user-scalable`, `min-scale`, and `max-scale`.

    From https://webkit.org/blog/7367/new-interaction-behaviors-in-ios-10/:

   >  "Now, we ignore the user-scalable, min-scale and max-scale settings. If you have content that disabled zoom, please test it on iOS 10, and understand that many users will be zooming now."

---

See also:

 * https://www.youtube.com/watch?v=8J6EdpXdzqc&t=23m31s

alrra:
Fix #82

### DIFF
--- a/src/lib/rules/meta-viewport/meta-viewport.ts
+++ b/src/lib/rules/meta-viewport/meta-viewport.ts
@@ -1,0 +1,201 @@
+/**
+ * @fileoverview Check if the viewport meta tag is specified in the
+ * `<head>` with the proper value.
+ */
+
+import { parseMetaViewPortContent } from 'metaviewport-parser';
+
+import { isHTMLDocument, normalizeString } from '../../utils/misc';
+import { IAsyncHTMLDocument, IAsyncHTMLElement, ITraverseEnd } from '../../types'; // eslint-disable-line no-unused-vars
+import { IRule, IRuleBuilder } from '../../types'; // eslint-disable-line no-unused-vars
+
+import { RuleContext } from '../../rule-context'; // eslint-disable-line no-unused-vars
+
+// ------------------------------------------------------------------------------
+// Public
+// ------------------------------------------------------------------------------
+
+const rule: IRuleBuilder = {
+    create(context: RuleContext): IRule {
+
+        // This function exists because not all connector (e.g.: jsdom)
+        // support matching attribute values case-insensitively.
+        //
+        // https://www.w3.org/TR/selectors4/#attribute-case
+
+        const getViewportMetaTags = (elements: Array<IAsyncHTMLElement>): Array<IAsyncHTMLElement> => {
+            return elements.filter((element) => {
+                return (element.getAttribute('name') !== null && normalizeString(element.getAttribute('name')) === 'viewport');
+            });
+        };
+
+        const listIncludesBrowsersWithOrientationChangeBug = (browsersList: string[]): boolean => {
+
+            // Old versions of Safari for iOS require `initial-scale=1`
+            // in order to react on orientation change.
+            //
+            //  * https://www.quirksmode.org/blog/archives/2013/10/initialscale1_m.html
+            //  * https://www.quirksmode.org/blog/archives/2013/10/more_about_scal.html
+            //  * https://www.quirksmode.org/mobile/metaviewport/quickdevicewidth.html
+            //
+            // As of Safari for iOS 9+, that is no longer the case.
+            //
+            //  * https://twitter.com/ppk/status/829329905942986752
+
+            return browsersList.some((browserVersion) => {
+
+                // The following is done because `browsersList` returns
+                // values such as: 'ios_saf 10.3', 'ios_saf 10.0-10.2'.
+
+                const version = (/ios_saf (\d+)\.?.*/).exec(browserVersion);
+
+                return version ? parseInt(version[1]) < 9 : false;
+            });
+        };
+
+        const checkContentValue = async (contentValue: string|null, resource: string, viewportMetaTag: IAsyncHTMLElement) => {
+
+            if (!contentValue) {
+                await context.report(resource, viewportMetaTag, `Meta tag should have non-empty 'content' attribute`);
+
+                return;
+            }
+
+            const content = parseMetaViewPortContent(contentValue);
+
+            // Check for unknown properties and invalid values.
+
+            for (const key of Object.keys(content.unknownProperties)) {
+                await context.report(resource, viewportMetaTag, `Meta tag has unknown property: '${key}'`);
+            }
+
+            for (const key of Object.keys(content.invalidValues)) {
+                await context.report(resource, viewportMetaTag, `Meta tag has invalid value '${content.invalidValues[key]}' for property '${key}'`);
+            }
+
+            // Disallow certain properties.
+
+            for (const key of Object.keys(content.validProperties)) {
+
+                // The following properties allow to block the user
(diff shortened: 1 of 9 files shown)


alrra:
Note: This is still `work in progress` as I need to do some minor changes (mostly documentation related).

src/lib/rules/meta-viewport/meta-viewport.ts
```diff
@@ -0,0 +1,208 @@
...
+            // are also supported.
+
+            if (browserslist.length === 0) {
+                return true;
+            }
+
+            return browsersList.every((e) => {
+
+                // The following is done because `browserslist` returns
+                // valudes such as: 'ios_saf 10.3', 'ios_saf 10.0-10.2'.
```
qzhou1607: returns `values`

src/lib/rules/meta-viewport/meta-viewport.ts
```diff
@@ -0,0 +1,208 @@
...
+
+                const version = (/ios_saf (\d+)\.?.*/).exec(e);
+
+                return version ? parseInt(version[1]) < 9 : true;
```
qzhou1607: If `browsersList` includes both `ios_saf 8` and `ios_saf 10`, `listIncludesBrowsersWithOrientationChangeBug` will be set to `false`. But it should actually be `true`. Right?

alrra: Good catch! Should be fixed.

src/lib/rules/meta-viewport/meta-viewport.ts
```diff
@@ -0,0 +1,208 @@
...
+            //  * https://twitter.com/ppk/status/829329905942986752
+
+            // If no browsers are targeted, assume Safari for iOS < 9
+            // are also supported.
+
+            if (browsersList.length === 0) {
+                return true;
+            }
+
+            return browsersList.some((e) => {
```
sarvaje: Call me picky but, `e` sound like an error or event to me. Maybe you can call it `browser` or something similar.

alrra: > Call me picky 

You are not, and this is a good call, thanks!

Note: I generate the `some` and similar things automatically in my editor, and `e` comes from `element`, so if you see it in the future, that is why.

> `browser`

Let's go with `browserVersion`.

molant:
@alrra any things that is pending here code wise or just documentation? And what do you want to change in the documentation?

alrra:
> any things that is pending here code wise or just documentation? 

Only documentation, otherwise it's done for quite some time.

> And what do you want to change in the documentation?

The TODOs: [1](https://github.com/sonarwhal/sonar/pull/449/files#diff-93a797307aaba7f38f327913586bbc4dR12) + [2](https://github.com/sonarwhal/sonar/pull/449/files#diff-93a797307aaba7f38f327913586bbc4dR19).

molant:
@alrra what about this:

## Why is this important? 
 
> Mobile browsers render pages in a virtual "window" (the viewport), usually wider than the screen, so they don't need to squeeze every page layout into a tiny window (which would break many non-mobile-optimized sites). Users can pan and zoom to see different areas of the page.

> Mobile Safari introduced the "viewport meta tag" to let web developers control the viewport's size and scale. Many other mobile browsers now support this tag, although it is not part of any web standard.

> In recent years, screen resolutions have risen to the size that individual pixels are hard to distinguish with the human eye. For example, recent smartphones generally have a 5-inch screens with resolutions upwards of 1920—1080 pixels (~400 dpi). Because of this, many browsers can display their pages in a smaller physical size by translating multiple hardware pixels for each CSS "pixel". Initially this caused usability and readability problems on many touch-optimized web sites.

*[Using the viewport meta tag to control layout on mobile devices (MDN)](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag)*

The viewport topic is complex. If you want to dig deeper, 
read Peter-Paul Koch's "A tale of two viewports" [part one][ppk 
article 1] and [part two][ppk article 2], or watch his talk 
['The Mobile Viewports'][ppk talk]. 

This rule checks if you have a viewport meta tag in your head, and if it is in the right format depending on the browsers you are targeting. E.g.: Old versions of Safari for iOS require `initial-scale=1` in order to react on orientation change. 

**NOTE:** If your website is not responsive or will not work on mobile devices, then you might not need this meta tag.

Ideally the following meta `viewport` tag should be used: 

[...]

src/lib/rules/meta-viewport/meta-viewport.ts
sarvaje: >Note: I generate the some and similar things automatically in my editor, and e comes from element, so if you see it in the future, that is why.

ahhh hehehe.

>Let's go with `browserVersion`.

Perfect!

src/lib/rules/meta-viewport/meta-viewport.ts
```diff
@@ -0,0 +1,208 @@
...
+            //  * https://www.quirksmode.org/mobile/metaviewport/quickdevicewidth.html
+            //
+            // As of Safari for iOS 9+, that is no longer the case.
+            //
+            //  * https://twitter.com/ppk/status/829329905942986752
+
+            // If no browsers are targeted, assume Safari for iOS < 9
+            // are also supported.
+
+            if (browsersList.length === 0) {
```
molant: With the work on #452, if `browserslist` is empty it will use the defaults so this should never happen.

alrra: Fixed.

alrra:
> Using the viewport meta tag to control layout on mobile devices (MDN)

I wanted to avoid copy-pasting content. But let's go with that for the time being, and if needed, update things later.

> If your website is not responsive or will not work on mobile devices

Let's just stick with _"not responsive"_, since the web works on all kind of weird devices and places.  Plus, I personally think we should not give people the idea that they should build some kind of special version of their site just for mobile devices.

> This rule checks if you have a viewport meta tag in your head, and if it is in the right format depending on the browsers you are targeting. E.g.: Old versions of Safari for iOS require initial-scale=1 in order to react on orientation change.

That part is already included under [`What does the rule check?`](https://github.com/sonarwhal/sonar/pull/449/files#diff-93a797307aaba7f38f327913586bbc4dR75) section.

molant:
> That part is already included under What does the rule check? section.

I think we should have some kind of TL;DR; in the top section so users don't have to dig searching what the rule does exactly.

alrra:
> I think we should have some kind of TL;DR; in the top section so users don't have to dig searching what the rule does exactly.

The [first thing is also basically that](https://github.com/sonarwhal/sonar/pull/449/files#diff-93a797307aaba7f38f327913586bbc4dR3).

alrra:
Updated: The [first thing is also basically that](https://github.com/sonarwhal/sonar/pull/449/files#diff-93a797307aaba7f38f327913586bbc4dR3).

Issue #446, opened 2017-08-22
Ask about browserlist when running --init and update default value


qzhou1607:
When `browserlist` in the config file is set as an empty string, the `targetedBrowsers` in the `context` of a rule will output an empty array. According to @alrra currently this output means all browsers are supported. @molant suggested maybe we have a default of latest 2 versions in the browserlist. And also I think we should document the browerlist in more details.

alrra:
> When browserlist in the config file is set as an empty string, 
>
> According to @alrra currently this output means all browsers are supported. 

Yes, my opinion is that, if the user didn't targeted any specific browsers, we should offer full support.

> @molant suggested maybe we have a default of latest 2 versions in the browserlist. 

👍

molant:
@alrra so what is your final opinion? Your 👍 to my opinion confuses me 😵

alrra:
> so what is your final opinion? Your 👍 to my opinion confuses me 😵

Sorry about that. What I meant by 👍 is that I agree with you that we should have/provide as the default in the `.sonarrc` file  `"last 2 versions"`.

molant:
OK, so the plan for now is:

* Have a default `"browserlist: ["last 2 versions"]` in our default json
* Add another question to the `--init` option to ask about browser support
* `browserlist == []` implies that all browsers are targeted

@alrra did I miss something?

alrra:
> did I miss something?

No.

>  browserlist == [] implies that all browsers are targeted

...all browsers that are somewhat still in use.

alrra:
> Have a default "browserlist: ["last 2 versions"] in our default json

Thinking a bit more about it, I think that should actually be `[  '> 1%',  'last 2 versions' ]`.

---

Note:  [`browserlist` default to](https://github.com/ai/browserslist/blob/630c3f8a1513ee263158259b43f5029b494f32fb/index.js#L303-L307): `[  '> 1%',  'last 2 versions',  'Firefox ESR' ]`.

molant:
> Note: browserlist defaults to: [ '> 1%', 'last 2 versions', 'Firefox ESR' ].

Do we have something wrong in our code? If we don't specify anything in our sonarrc we don't get any list at all.

qzhou1607:
[...]

PR #451, opened 2017-08-23
Breaking: Change`highest-available-document-mode` rule removal suggestions

### LINKED ISSUE #447: Configure `suggestRemoval` correctly when `targetBrowsers` is empty
https://github.com/sonarwhal/sonar/blob/ac7a3c31d333c21bc44cb4e0037453d0a1aa79bf/src/lib/rules/highest-available-document-mode/highest-available-document-mode.ts#L162-L164

See #446 for details.

alrra:
If no browsers are targeted (i.e. `browserlist` is empty), make `highest-available-document-mode` rule assume Internet Explorer 8, 9, and 10 are supported.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Ref #446
Fix #447

### DIFF
--- a/tests/lib/rules/highest-available-document-mode/tests.ts
+++ b/tests/lib/rules/highest-available-document-mode/tests.ts
@@ -91,25 +91,46 @@ const testsForRequireMetaTagConfig: Array<IRuleTest> = [
     }
 ];
 
+const testsForHeaderWithBlankTargetBrowsersConfig: Array<IRuleTest> = [
+    {
+        name: `HTML page is served without the 'X-UA-Compatible' header, but it's required by the targeted browsers`,
+        reports: [{ message: `'x-ua-compatible' header was not specified` }],
+        serverConfig: { '/': '' }
+    }
+];
+
+const testsForMetaTagWithBlankTargetBrowsersConfig: Array<IRuleTest> = [
+    {
+        name: `'X-UA-Compatible' meta tag is not specified, but it's required by the targeted browsers`,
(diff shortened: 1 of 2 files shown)


tests/lib/rules/highest-available-document-mode/tests.ts
```diff
@@ -91,25 +91,46 @@ const testsForRequireMetaTagConfig: Array<IRuleTest> = [
...
+            }
+        }
     }
 ];
 
 ruleRunner.testRule(ruleName, testsForDefaults);
 ruleRunner.testRule(ruleName, testsForRequireMetaTagConfig, { ruleOptions: { requireMetaTag: true } });
+ruleRunner.testRule(ruleName, testsForHeaderWithBlankTargetBrowsersConfig, { browserslist: [] });
+ruleRunner.testRule(ruleName, testsForMetaTagWithBlankTargetBrowsersConfig, {
+    browserslist: [],
```
qzhou1607: The behavior when `browserslist` is not set is different from `browserslist` is set as empty. See https://github.com/sonarwhal/sonar/issues/446#issuecomment-324388133. It doesn't affect the result of this test though.

molant: @alrra I think we should decide first what to do with what @qzhou1607 has pointed out in that comment. I think we should use browserlist's default behavior and don't make any assumption in our rules. The reason is that different rules could implement different "default behaviors", tracking them and making sure all are consistent is more complicated than just having a single place and update it there in the future.

molant:
Closing this in favor of #453

PR #455, opened 2017-08-25
New: Ask about browserslist when generating config

### LINKED ISSUE #446: Ask about browserlist when running --init and update default value

qzhou1607:
Fix #446 
I'm currently working on adding tests, but feedback on the options are welcome:

I checked out a few dependents of `browserslist` to see what queries are most often used. `last 2 versions` show up frequently, but a lot of them simply provide an interface to pass their users' queries to `browserslist`, which means they use `browserslist`'s default if not defined. Here in this PR, we have 4 options in the beginning:
* Default list from browserslist
* Last 2 versions of all browsers
* Add by percentage of usage statistics. (e.g., `>= 5%`, 5% by default)
* Add by specific browser type and version number range. (e.g., `IE 6-8`)

The users can define the value of `usage percentage`, `browser type` and the `version numbers`. Also the dialogue will be prompted in a recursive manner so if a user wants to add additional query he/she will be able to do so.

### DIFF
--- a/src/lib/cli/browserslist-generator.ts
+++ b/src/lib/cli/browserslist-generator.ts
@@ -1,0 +1,65 @@
+/**
+ * @fileoverview Generates a valid browserslist config.
+ */
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+import * as inquirer from 'inquirer';
+import * as browserslist from 'browserslist';
+import * as logger from '../utils/logging';
+import { debug as d } from '../utils/debug';
+
+const debug: debug.IDebugger = d(__filename);
+
+export const generateBrowserslistConfig = (): Promise<Array<string>> => {
+    debug('Initiating browserslist config generator');
+
+    const addBrowsersListOptions: Array<inquirer.ChoiceType> = [
+        { name: 'Default (last 2 versions of each browser, and browsers with globaly usage over 1%, plus Firefox ESR)', value: 'default' },
+        { name: 'Custom (use browserslist format:  https://github.com/ai/browserslist#queries)', value: 'custom' }
+    ];
+
+    const browsersListQuestions: inquirer.Questions = [
+        {
+            choices: addBrowsersListOptions,
+            message: 'What browsers are you targeting?',
+            name: 'targetBy',
+            type: 'list'
+        },
+        {
+            message: 'Please enter the queries you want to specify (Use comma as the separator if you have more than one query):',
+            name: 'customQueries',
+            when: (answers) => {
+                return answers.targetBy === 'custom';
+            }
+        }
+    ];
+
+    const askAndValidate = async (): Promise<Array<string>> => {
+        const results: inquirer.Answers = await inquirer.prompt(browsersListQuestions);
+
+        if (results.targetBy === 'default') {
+            return [];
+        }
+
+        const customQueries: Array<string> = results.customQueries.split(',').map((query: string) => {
+            return query.trim();
+        });
+
+        try {
+            browserslist(customQueries);
+
+            return customQueries;
+        } catch (err) {
+            // The query format is invalid.
+            logger.log(`${err.message}.`);
+            logger.log('Please try again.');
+
+            return askAndValidate();
+        }
+    };
+
+    return askAndValidate();
+};
--- a/package-lock.json
+++ b/package-lock.json
@@ -161,6 +161,16 @@
         "@types/node": "8.0.14"
       }
     },
+    "@types/inquirer": {
+      "version": "0.0.35",
+      "resolved": "https://registry.npmjs.org/@types/inquirer/-/inquirer-0.0.35.tgz",
+      "integrity": "sha512-/6WyyGROIw6qCmS8KmQtyt9Tj1OX7xIrGloAwMMOaNb3W/+QrrlL4Be4TDvLO4dlvw8gFrHfeXiWxS/k82jgEQ==",
+      "dev": true,
+      "requires": {
+        "@types/rx": "4.1.1",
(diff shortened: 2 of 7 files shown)


src/lib/cli/browserslist-generator.ts
```diff
@@ -0,0 +1,157 @@
...
+    debug('Initiating browserslist config generator');
+
+    const browserslistConfig = [];
+    const addBrowsersListOptions = [
+        { name: 'Use Default browsers', value: 'default' },
+        { name: 'Add the last 2 versions of each browser', value: 'last2' },
+        { name: 'Add by percentage of usage statistics. (e.g., `>= 5%`)', value: 'usage' },
+        { name: 'Add by specific browser type and version number range. (e.g., `IE 6-8`)', value: 'type' }
+    ];
+    const browsers = ['Chrome', 'Edge', 'Explorer', 'Firefox', 'Opera', 'Safari'];
```
alrra: Providing only these browser kinda hints that we only care about desktop browsers. Mobile is a big thing nowadays, and [mobile browsers such as Safari for iOS, UC, Samsung Internet, etc.](http://gs.statcounter.com/), have quite a significant market share. 

Also, in general I personally don't think this is the way to go, we are over complicating things here.

---

I think we should have 2 options: 

* Use default query.
* Specify custom query (+ a link to the [browserslist docs](https://github.com/ai/browserslist#queries))                              

@sonarwhal/contributors :arrow_up:

molant:
> Specify custom query (+ a link to the browserslist docs)

Problem with that is that I think we should validate it and I really don't want to. It would be the only field where we would accept any input. If we decide to go forward, maybe just add the property in the file and let the users know they have to update it?
Also, what happens if the formation not valid? If browserslist doesn't break then I'm OK with just those 2.

alrra:
> Problem with that is that I think we should validate it and I really don't want to.
> Also, what happens if the formation not valid? If browserslist doesn't break then I'm OK with just those 2.

`browserslist` does the validation automatically, for example for `eddge 15`:

```
{ BrowserslistError: Unknown browser eddge
    at checkName (/Users/alrra/projects/sonar/node_modules/browserslist/index.js:339:20)
    at Function.select (/Users/alrra/projects/sonar/node_modules/browserslist/index.js:680:18)
    at /Users/alrra/projects/sonar/node_modules/browserslist/index.js:226:33
    at Array.forEach (native)
    at browserslist (/Users/alrra/projects/sonar/node_modules/browserslist/index.js:208:11)
    at process (/Users/alrra/projects/sonar/a.js:6:24)
    at Object.<anonymous> (/Users/alrra/projects/sonar/a.js:14:13)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
  name: 'BrowserslistError',
  message: 'Unknown browser eddge',
  browserslist: true }
```

qzhou1607:
@molant @alrra so it looks like `browserslist` does break on invalid queries. What do you think if we have the two options as @alrra suggested, but catch the error in the code if it breaks and provides the error message along with the browserslist doc link so that the user can fix it in `.sonarrc`?

molant:
> What do you think if we have the two options as @alrra suggested,
OK

Let's put something like: 

> What browsers are you targeting?
* Default: last 2 versions with more than 1% of market share (is this exactly what the defaults are?)
* Custom (use browserslist format: https://github.com/ai/browserslist#queries)

We catch an exception as @qzhou1607 proposes and if it breaks then we ask the question again with a message saying `Invalid format, please try again`.

qzhou1607:
@molant The default of `browserslist` is `last 2 versions`, ` >1 %`, and plus `Firefox ESR` . See [Ref](https://github.com/ai/browserslist/blob/cb9abd4ae061a256157ff93c1bd7dc64a6997a2e/index.js#L325-L327).
> * Default: last 2 versions with more than 1% of market share (is this exactly what the defaults are?)

Just to clarify: It won't break when we set the value when running `init`. The error will only be thrown when we run sonar with the invalid queries. So technically we can't `try again` immediately. The user will have to go to `.sonarrc` to fix it with the correct query and then run sonar again.
> if it breaks then we ask the question again with a message saying `Invalid format, please try again`.

alrra:
> We catch an exception as @qzhou1607 proposes and if it breaks then we ask the question again with a message saying Invalid format, please try again.

Also, say exactly what is invalid, `browserslist` provides that info.

alrra:
> Just to clarify: It won't break when we set the value when running init

You can do something like to test things out:

```js

try { 
    browserslist(theCustomQueryTheUserEntered) `
} catch (e) {
    // handle failure, and re ask question
}
```

src/lib/cli/browserslist-generator.ts
```diff
@@ -0,0 +1,65 @@
...
+// ------------------------------------------------------------------------------
+
+import * as inquirer from 'inquirer';
+import * as browserslist from 'browserslist';
+import * as logger from '../utils/logging';
+import { debug as d } from '../utils/debug';
+
+const debug: debug.IDebugger = d(__filename);
+
+export const generateBrowserslistConfig = (): Promise<string[]> => {
```
sarvaje: `Promise<Array<string>>`

It is just to be consistent with the style. we are using that style everywhere in the code. There are more places where you have this (or similar).

qzhou1607: @sarvaje Fixed.

PR #456, opened 2017-08-25
New: Add rule to check for vulnerable libraries

### LINKED ISSUE #125: Add rule to check for old libraries and/or vulnerabilities
We could integrate with [snyk.io vulnerability database](https://snyk.io/vuln/).
[This post by them](https://snyk.io/blog/77-percent-of-sites-use-vulnerable-js-libraries/) is really interesting.

Things to do:
* [x] Find a way to execute JS in the page to detect what libraries they are running
* [x] Create a list of libraries to detect and how to do it
* [x] Find a way for libraries to easily report newer versions and ways to detect them (maybe an online service and we have some patterns already in the rule?)

molant:
Fix #125


molant:
Still need to add tests but you can start reviewing the code and documentation.

molant:
This is how it looks like at the moment:
image

src/lib/rules/no-vulnerable-libraries/no-vulnerable-libraries.ts
```diff
@@ -0,0 +1,203 @@
...
+            }
+        };
+
+        /** Given a list of libraries, reports the ones that have known vulnerabilities. */
+        const detectAndReportVulnerableLibraries = async (libraries: Array<Library>, resource: string) => {
+            // TODO: Check if snykDB is older than 24h and download if so. If not, or if itfails, use local version
+            const snykDB = await loadSnyk();
+
+            for (const lib of libraries) {
+                const snykInfo = snykDB.npm[lib.npmPkgName] as Vulnerability[];
```
sarvaje: `Array<Vulnerability>`

molant: We have a rule configured to for `as Something` over the other way of casting. Or are you talking about adding the type to `snykInfo`? In that case it is assigned directly because of the `as Vulnerability[]`.

alrra: > We have a rule configured to for as Something over the other way of casting

That's because of [this](https://github.com/sonarwhal/sonar/blob/bbf1e6eaa401a14733a623a9416292966d8e64e0/.eslintrc.json#L259) (see also: [`no-angle-bracket-type-assertion`](https://github.com/nzakas/eslint-plugin-typescript/blob/f860b4bf3b43e6129fc5ca28564b8ef75a79df5c/docs/rules/no-angle-bracket-type-assertion.md)).

sarvaje: I was talking about to change `Vulnerbility[]` to `Array<Vulnerability>` because we are using that way in all the code.
The rule `no-angle-bracket-type-assertion` should be used if our code include `tsx` files [see here](https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/no-angle-bracket-type-assertion.md) if not, we can remove that rule

molant:
@sarvaje changes are in

sarvaje:
>@sarvaje changes are in

I think you miss the review message hehehe
>Can you review the variable types? there are a lot of missed types

molant:
@sarvaje most of the variables are assigned directly so the type is there and will be maintained. I'll review the ones that don't like:
>   let snykStat

sarvaje:
ok!

molant:
@sarvaje I've added the type for a couple ones. Merge if you thing this is ready.

Issue #361, opened 2017-07-05
Provide more information when running `--init`


alrra:
From https://github.com/sonarwhal/sonar/pull/358#discussion_r125529863:

> We could improve already the question here and briefly explain what a connector is. Maybe add a description for each one? Something similar to:
> > What connector do you want to load the content of the website?
> > * cdp (you will need to have chrome installed)
> > * jsdom (will use jsdom, no additional installation required)

----

See also: https://github.com/sonarwhal/sonar/issues/326#issue-238851829

molant:
We should use the shortlinks of https://github.com/sonarwhal/sonarwhal.com/issues/185 for this.

PR #466, opened 2017-08-30
Breaking: Rename cdp connector to chrome

molant:
Work on #361


molant:
Renaming cdp to chrome because it is more meaningful and we are using `cdp` in the `shared` part for those connectors that rely in the protocol.

This should help with #361 and the first run experience for users.

PR #468, opened 2017-08-30
Fix: Filter out duplicate fetch requests in jsdom

### LINKED ISSUE #460: Same issue reported multiple times for the same resource on jsdom
**URL:** https://capitalone.com
**Connector:** jsdom
**Rule**: no-disallowed-headers

image

This doesn't reproduce with `cdp`

qzhou1607:
Fix #460

### DIFF
--- a/src/lib/connectors/jsdom/jsdom.ts
+++ b/src/lib/connectors/jsdom/jsdom.ts
@@ -77,6 +77,7 @@ class JSDOMConnector implements IConnector {
     private _manifestIsSpecified: boolean = false;
     private _window: Window;
     private _document: JSDOMAsyncHTMLDocument;
+    private _fetchedHrefs: Set<string>;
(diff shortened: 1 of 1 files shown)


src/lib/connectors/jsdom/jsdom.ts
```diff
@@ -77,6 +77,7 @@ class JSDOMConnector implements IConnector {
     private _manifestIsSpecified: boolean = false;
     private _window: Window;
     private _document: JSDOMAsyncHTMLDocument;
+    private _fetchedHrefs: Array<string>;
```
sarvaje: I think this is perfect for a `Set` instead of an array.

qzhou1607: @sarvaje Fixed.

PR #473, opened 2017-08-30
Fix: Fix error in `onLoadingFailed`

### LINKED ISSUE #469: Error in `onLoadingFailed`
Connector: chrome
Url: betterment.com

```
PS C:\github\Sonar.git> cd 'c:\github\Sonar.git'; node --inspect=27766 --debug-brk dist\src\bin\sonar.js betterment.com
Debugger listening on ws://127.0.0.1:27766/e1a11ceb-3bc1-423a-9dd6-fbf6a5dfeca9
For help see https://nodejs.org/en/docs/inspector
Debugger attached.
(node:12540) [DEP0062] DeprecationWarning: `node --inspect --debug-brk` is deprecated. Please use `node --inspect-brk` instead.
- Finishing...TypeError: Cannot match against 'undefined' or 'null'.
    at ChromeConnector.onLoadingFailed (C:\github\Sonar.git\dist\src\lib\connectors\shared\remote-debugging-connector.js:173:63)
    at <anonymous>
\ https://betterment.repo … r/default/csp/reportOnly downloaded
```

qzhou1607:
Fix #469

The reason that this issue happened is as below:

The page has many failed requests to https://betterment.report-uri.io/r/default/csp/reportOnly. Both `onResponseReceived` and `onLoadingFailed` were fired upon the failed requests. In the beginning, this request exists in `this._requests`, so  the check below in `onLoadingFailed` passes. But soon after this check the request is removed from `this._requests` in the `onResponseReceived` since it is running cocurrently. So later on in `onLoadingFailed`, the destructuring `const { request: { url: resource } } = this._requests.get(params.requestId)` fails, because at this stage, the request has already been removed from `this._requests` and `this._requests.get(params.requestId)` returns nothing.

The check in `onLoadingFailed` that gets bypassed:
https://github.com/sonarwhal/sonar/blob/0cc1f05e515755e6f25542c6c4d0362b48e3ba4e/src/lib/connectors/shared/remote-debugging-connector.ts#L216-L224

Solution:
Instead of retrieving request from `this._requests` every time, use the value previously saved in variable `request` for destructuring assignment. However, this means for a failed request, both `fetch::end` and ` fetch::error` might be emitted. I'm not sure if this is the intended behavior. Maybe we should consider adding conditions for emitting `fetch::end`  like only emit if the status is 200?

### DIFF
--- a/src/lib/connectors/jsdom/jsdom.ts
+++ b/src/lib/connectors/jsdom/jsdom.ts
@@ -210,6 +210,7 @@ class JSDOMConnector implements IConnector {
 
             // TODO: Replace `null` with `resource` once it
             // can be converted to `JSDOMAsyncHTMLElement`.
+            // Event is also emitted when status code in response is not 200.
             await this._server.emitAsync('fetch::end', fetchEndEvent);
 
             return callback(null, resourceNetworkData.response.body.content);
@@ -368,6 +369,7 @@ class JSDOMConnector implements IConnector {
                 response: this._targetNetworkData.response
             };
 
+            // Event is also emitted when status code in response is not 200.
             await this._server.emitAsync('targetfetch::end', fetchEnd);
 
             jsdom.env({
--- a/src/lib/connectors/shared/remote-debugging-connector.ts
+++ b/src/lib/connectors/shared/remote-debugging-connector.ts
@@ -233,7 +233,7 @@ export class Connector implements IConnector {
         }
 
         if (params.type === 'Manifest') {
-            const { request: { url: resource } } = this._requests.get(params.requestId);
+            const { request: { url: resource } } = request;
             const event: IManifestFetchError = {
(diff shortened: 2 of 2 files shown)


molant:
> Maybe we should consider adding conditions for emitting fetch::end like only emit if the status is 200? 

Yes, let's do that.

Also, can you add a tests for this?

qzhou1607:
@molant Does `fetch::end` need to be emitted on status 404?
>Yes, let's do that.
> Maybe we should consider adding conditions for emitting fetch::end like only emit if the status is 200?

molant:
>  Does fetch::end need to be emitted on status 404?

IIRC we only send end on success. We should continue doing that.

molant:
> IIRC we only send `end` on success. We should continue doing that.

Looks like my previous statement is wrong, so we don't have to change the behavior.

PR #479, opened 2017-08-31
Fix: Remove `null` locations in the error message

### LINKED ISSUE #478: Issue with error location and duplicate page evaluation
connector: chrome
url: https://www.w3schools.com/js/

Two issues:
* location shows as null when not available
* duplicate messages

location

qzhou1607:
Fix #478

The page loads some scripts multiple times, which results in the duplicate messages. It makes sense but looks like duplicate evaluation, should we add a filter similar to what we did to `jsdom`  in https://github.com/sonarwhal/sonar/pull/468?

### DIFF
--- a/src/lib/rule-context.ts
+++ b/src/lib/rule-context.ts
@@ -97,12 +97,8 @@ export class RuleContext {
             sourceCode = (await element.outerHTML()).replace(/[\t]/g, '    ');
         }
 
-        if (position === null) {
-            position = {
-                column: null,
-                line: null
-            };
-        }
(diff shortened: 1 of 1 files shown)


molant:
>  It makes sense but looks like duplicate evaluation, should we add a filter similar to what we did to jsdom in #468?

I don't think we should. If it is a script or a css then the browser will evaluate it again (that's what seems to happen here). Maybe we should update the behavior of `jsdom` to only cache resources that are not scripts or css? This will match more closely the behavior of `chrome`. If we go down that path we need to add tests for duplicate resources on a page and how they should be handled.

PR #480, opened 2017-08-31
Chore: Use compact html checker reports

### LINKED ISSUE #474: Redundant errors in html-checker
connector: `chrome` and `jsdom`
url: www.gmail.com
rule: html-checker

About 366 errors were reported from `html-checker`.
html-checker

But running `https://www.gmail.com` in https://validator.w3.org/nu/ returns only 4 errors. For this particular page, the html returned from our connector seems to be different from the content fetched directly by the validator using the url.

qzhou1607:
Fix #474

366 errors/warnings from `html-checker` are reduced to ~ 20:

gmail-html-checker

### DIFF
--- a/tests/lib/rules/html-checker/tests.ts
+++ b/tests/lib/rules/html-checker/tests.ts
@@ -43,6 +43,17 @@ const defaultCheckerMessages = {
             extract: '-section"><h1>example<',
             hiliteStart: 10,
             hiliteLength: 4
+        },
+        {
+            type: 'info',
+            lastLine: 1,
+            lastColumn: 3000,
+            firstColumn: 2000,
+            subType: 'warning',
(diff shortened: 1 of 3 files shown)


tests/lib/rules/html-checker/tests.ts
```diff
@@ -185,5 +232,9 @@ ruleRunner.testRule(ruleName, testsForValidatorConfig, {
     ruleOptions: { validator: configValidator },
     serial: true
 });
+ruleRunner.testRule(ruleName, testsForDetailsConfig, {
+    ruleOptions: { details: true },
+    serial: true
```
molant: Do we need it to be `serial`?

qzhou1607: @molant Yes. I'm not sure if it's because of the use of `mock`, running tests in parallel doesn't work.

PR #482, opened 2017-09-01
Fix: Error with `jsdom` and attribute names with .

molant:


molant:
Lots of Microsoft properties have attributes such as `ms.title` for analytics. These names will break `jsdom`'s `querySelectorAll` with a `SyntaxError`.

This fixes that.

PR #488, opened 2017-09-02
New: Add rule to check the usage of the `apple-touch-icon`s

### LINKED ISSUE #33: Add rule to check the usage of `apple-touch-icon`s
- [x] Check if the `<link>` tag (e.g. `<link rel="apple-touch-icon" href="apple-touch-icon.png">`) is specified

     Note: In the past, people usually just had the `apple-touch-icon` in the root of the site, but that is no longer consider a good practice and can create a lot of issues (see: https://github.com/h5bp/html5-boilerplate/pull/1622).

- [x] Check if a `180×180px` image is not used

- [x] Check if multiple multiple images of different sizes are used   

     Note: Usually iOS devices get upgraded pretty quickly, [most people being on the latest 2 version of the iOS](https://developer.apple.com/support/app-store/), so specifying multiple sizes of the `apple-touch-icon` just adds to the weight of the page, without no real benefit. One `180×180px` `apple-touch-icon` is nowadays enough for all cases, as [Safari will scale it down automatically if needed](https://realfavicongenerator.net/blog/how-ios-scales-the-apple-touch-icon/) (see also: https://github.com/h5bp/html5-boilerplate/issues/1367).

- [ ] Check if the image [has transparent background](https://realfavicongenerator.net/blog/apple-touch-icon-turns-black/)

- [x] Other?

alrra:
Fix #33


src/lib/rules/apple-touch-icons/apple-touch-icons.ts
```diff
@@ -0,0 +1,268 @@
...
+        return { 'traverse::end': validate };
+    },
+
+    meta: {
+        docs: {
+            category: `PWAs`,
+            description: `Require an 'apple-touch-icon'`
+        },
+        fixable: 'code',
+        ignoredConnectors: ['chrome'],
```
alrra: @molant The _"is this a PNG?", "is it 180x180px?"_, etc, will not work with the `chrome` connector as they require the raw bytes of the image. :(

molant: Why can't you use [`rawContent`](https://github.com/sonarwhal/sonar/blob/master/src/lib/connectors/shared/remote-debugging-connector.ts#L294)?

alrra: [I am](https://github.com/sonarwhal/sonar/pull/488/files#diff-046ee208e76bb5dffb66c554f2aac000R123). Now that you mention it, `rawContent` used to work for `chrome`, something must be wrong.

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
```diff
@@ -0,0 +1,268 @@
...
+
+            // Check if the image is a PNG.
+
+            if (image.type !== 'png') {
+                await context.report(resource, appleTouchIcon, `'${appleTouchIconHref}' is not a PNG`);
+            }
+
+            // Check if the image is 180x180px.
+
+            if (image.width !== 180 && image.height !== 180) {
```
sarvaje: I think this should be `image.width !== 180 || image.height !== 180` just in case someone use an image with size 180x160 for example.

alrra: Fixed and added a test, thanks!

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
```diff
@@ -0,0 +1,268 @@
...
+            }
+
+            // - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+            // Note: Async version of `image-size` doesn't work if the
+            // input is a Buffer.
+            //
+            // https://github.com/image-size/image-size/tree/4c527ba608d742fbb29f6d9b3c77b831b069cbb2#asynchronous
+
+            const image = getImageData(response.body.rawContent);
```
austinkelleher: `image-size` exposes an [async API](https://github.com/image-size/image-size#asynchronous) that can be promisified. I think we should use that:

```js
import { promisify } from 'util';
import * as getImageData from 'image-size';

const getImageDataAsync = promisify(getImageData);

...

const image = await getImageData(response.body.rawContent);
```

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
```diff
@@ -0,0 +1,268 @@
...
+            // so what this function will try to do is select the
+            // icon that will most likely generate the fewest errors.
+
+            let bestIcon;
+
+            for (const icon of icons) {
+                const sizes = normalizeString(icon.getAttribute('sizes'));
+
+                if (!sizes) {
+                    return icon;
```
austinkelleher: What do you think about `continue`ing here instead of returning. It could be possible that one of the other ones has `180x180`. Also, should we warn if they don't have sizes? I'm not too familiar with what is a best practice here, but just throwing out some thoughts.

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
```diff
@@ -0,0 +1,268 @@
...
+                } else if (sizes === '180x180') {
+                    bestIcon = icon;
+                }
+            }
+
+            return bestIcon ? bestIcon : icons[0];
```
austinkelleher: The ternary operator is not necessary here: 

```js
return bestIcon || icons[0];
```

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
```diff
@@ -0,0 +1,268 @@
...
+            // pass most of the following tests.
+
+            const appleTouchIcon: IAsyncHTMLElement = chooseBestIcon(appleTouchIcons);
+
+            // - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+            // Check if `rel='apple-touch-icon'`.
+            // See `getAppleTouchIcons` function for more details.
+
+            if (normalizeString(appleTouchIcon.getAttribute('rel')) !== 'apple-touch-icon') {
```
austinkelleher: It looks like in `getAppleTouchIcons` you're also checking if `normalizeString(appleTouchIcon.getAttribute('rel'))` includes `apple-touch-icon-precomposed`. Is it necessary to do that here too?

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
alrra: @austinkelleher See [my comment](https://github.com/sonarwhal/sonar/pull/488/files/9b9900abee8adc44d6956f876b0e7a0014c5f516#diff-046ee208e76bb5dffb66c554f2aac000R118).

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
```diff
@@ -0,0 +1,268 @@
...
+                //  or
+                //
+                //      <link rel="apple-touch-icon-precomposed" href="...">
+                //
+                //  or, since the `rel` attribute accepts a space
+                //  separated list of values in HTML, theoretically:
+                //
+                //      <link rel="apple-touch-icon-precomposed apple-touch-icon" href="...">
+                //
+                //  but that donesn't work in practice.
```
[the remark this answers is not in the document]
alrra: Fixed.

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
alrra: Fixed.

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
alrra: >  Is it necessary to do that here too?

Yes since you can have something such as: `REL="APPLE-TOUCH-ICON"`

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
alrra: > What do you think about continueing here instead of returning. It could be possible that one of the other ones has 180x180

That was intentional. I want the one without `sizes` to be chosen over the `180x180`. But thinking again about it, yes, maybe it will be better to choose the one with `sizes="180x180"`since in most cases, it will probably actually be `180x180px`, and that will require less work.

> Also, should we warn for those tags that do not have the sizes attribute? I'm not too familiar with what is a best practice here, but just throwing out some thoughts.

Since we are guiding the user to just one icon, using [`sizes` doesn't make much sense](https://github.com/sonarwhal/sonar/pull/488/files#diff-75ae57f8862939f10ef6a8409a4b777cR69).

alrra: > maybe it will be better to choose the one with `sizes="180x180"`

Done.

alrra:
@austinkelleher, @sarvaje This PR is ready for another review.

docs/user-guide/rules/apple-touch-icons.md
```diff
@@ -0,0 +1,223 @@
...
+
+    </head>
+    <body>
+        <link rel="apple-touch-icon" href="apple-touch-icon.png">
+        ...
+    </body>
+</html>
+```
+
+The `apple-touch-icon` has a `rel` attribute different then
```
molant: > different than

tests/lib/rules/apple-touch-icons/tests.ts
```diff
@@ -0,0 +1,157 @@
...
+import { IRuleTest } from '../../../helpers/rule-test-type'; // eslint-disable-line no-unused-vars
+import * as ruleRunner from '../../../helpers/rule-runner';
+
+const ruleName = getRuleName(__dirname);
+
+const defaultImage = fs.readFileSync(`${__dirname}/fixtures/apple-touch-icon.png`); // eslint-disable-line no-sync
+const imageWithIncorrectDimensions = fs.readFileSync(`${__dirname}/fixtures/incorrect-dimensions.png`); // eslint-disable-line no-sync
+const imageThatIsNotSquare = fs.readFileSync(`${__dirname}/fixtures/not-square.png`); // eslint-disable-line no-sync
+const imageWithIncorrectFileFormat = fs.readFileSync(`${__dirname}/fixtures/incorrect-file-format.png`); // eslint-disable-line no-sync
+// const imageWithTransparentBackground = fs.readFileSync(`${__dirname}/fixtures/transparent-background.png`); // eslint-disable-line no-sync
```
molant: Do you plan on adding the transparency part in this PR?

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
molant: The problem with `rawContent` is [here](https://github.com/sonarwhal/sonar/blob/master/src/lib/connectors/shared/remote-debugging-connector.ts#L798). Nothing is connected when using `fetchContent`.

I'll create an issue for that and work on it tomorrow.

alrra: OK.

docs/user-guide/rules/apple-touch-icons.md
alrra: Fixed.

tests/lib/rules/apple-touch-icons/tests.ts
alrra: > Do you plan on adding the transparency part in this PR?

No. That part is a bit tricky, so I'll try to do it once we have more time.

src/lib/rules/apple-touch-icons/apple-touch-icons.ts
molant: This is the issue https://github.com/sonarwhal/sonar/issues/495

tests/lib/rules/apple-touch-icons/tests.ts
molant: Then we should add an issue for that. What is the problem when using transparency?

alrra: > What is the problem when using transparency?

1) I want to void introducing a big image manipulation library just to check that.
2) I also want to figure out what should and should not pass (e.g.: you have just a few transparent pixels, it should be fine).

alrra: > Then we should add an issue for that

[Done.](https://github.com/sonarwhal/sonar/issues/496)

PR #494, opened 2017-09-06
Chore: Move `evaluate` to child process

### LINKED ISSUE #485: `jsdom` gets stuck at `traverse::end`
connector: jsdom
url: www.facebook.com

This happens when `axe` is activated and is back to normal when `axe` is deactivated. This delay seems to be a extra long time taken in the line below:
https://github.com/sonarwhal/sonar/blob/79b8da5d337ef915fe2895d424872bdd21ef4111/src/lib/connectors/jsdom/jsdom.ts#L467

qzhou1607:
* Move `evaluate` to its child process
* Start `evaluate` on `traverse::start`

This solves the stuck spinner issue.

Fix #485

### DIFF
--- a/src/lib/connectors/jsdom/evaluate-runner.ts
+++ b/src/lib/connectors/jsdom/evaluate-runner.ts
@@ -1,0 +1,47 @@
+import * as vm from 'vm';
+
+import * as jsdom from 'jsdom/lib/old-api';
+import * as jsdomutils from 'jsdom/lib/jsdom/living/generated/utils';
+
+const run = (data) => {
+    const { source } = data;
+    const result = {
+        error: null,
+        evaluate: 'result'
+    };
+    const url = process.argv[2];
+    const waitFor = process.argv[3];
+
+    jsdom.env({
+        done: (error, window) => {
+            if (error) {
+                result.error = error;
+
+                return process.send(result);
+            }
+
+            /* Even though `done()` is called after window.onload (so all resoruces and scripts executed),
+                      we might want to wait a few seconds if the site is lazy loading something. */
+            return setTimeout(async () => {
+                try {
+                    const script: vm.Script = new vm.Script(source);
+                    const evaluteResult = await script.runInContext(jsdomutils.implForWrapper(window.document)._global);
+
+                    result.evaluate = evaluteResult;
+                } catch (err) {
+                    result.error = err;
+                }
+
+                process.send(result);
+            }, waitFor);
+        },
+        features: {
+            FetchExternalResources: ['script', 'link', 'img'],
+            ProcessExternalResources: ['script'],
+            SkipExternalResources: false
+        },
+        url
+    });
+};
+
+process.on('message', run);
--- a/src/lib/rules/axe/axe.ts
+++ b/src/lib/rules/axe/axe.ts
@@ -11,7 +11,7 @@ import { AxeResults, Result as AxeResult, NodeResult as AxeNodeResult } from 'ax
 
 import { readFileAsync } from '../../utils/misc';
 import { debug as d } from '../../utils/debug';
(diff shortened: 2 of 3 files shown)


src/lib/connectors/jsdom/evaluate-runner.ts
```diff
@@ -0,0 +1,41 @@
...
+        done: async (error, window) => {
+            if (error) {
+                result.error = error;
+
+                return process.send(result);
+            }
+
+            try {
+                const script: vm.Script = new vm.Script(source);
+                const evaluteResult = await script.runInContext(jsdomutils.implForWrapper(window.document)._global);
```
sarvaje: After navigate you should wait for `this._options.waitFor` as it does in `jsdom.ts`. In that way, you will have a version of the site as much similar to the parent one as possible

src/lib/rules/axe/axe.ts
```diff
@@ -11,7 +11,7 @@ import { AxeResults, Result as AxeResult, NodeResult as AxeNodeResult } from 'ax
 
 import { readFileAsync } from '../../utils/misc';
 import { debug as d } from '../../utils/debug';
-import { IAsyncHTMLElement, ITraverseEnd, IRule, IRuleBuilder, Severity } from '../../types'; // eslint-disable-line no-unused-vars
+import { IAsyncHTMLElement, IRule, IRuleBuilder, Severity, ITraverseStart } from '../../types'; // eslint-disable-line no-unused-vars
```
molant: Now that I think about it, we have a problem switching to `traverse::start` with `chrome`.
We might want to create a custom event for evaluating scripts and each `connector` can fire it when it is more appropriate.
Maybe something like `canevaluate`?

@sonarwhal/contributors what do you think?

qzhou1607: @molant I haven't encounter any issue with `axe` using `chrome` so far. What's the issue with using `traverse::start` in chrome?
> Now that I think about it, we have a problem switching to traverse::start with chrome.

That way maybe we can have a more clear message to show when evaluate is running, like `Running scripts...`, rather than hanging at `Traversing the DOM` when it's actually evaluating.
> Maybe something like canevaluate?

molant: >  I haven't encounter any issue with axe using chrome so far.

The problem is that we will be telling people to `evaluate` scripts on `traverse::start`. Those scripts could modify the DOM and alter the results. That's why I'd rather do all of that at the end of the process for `chrome`. With `jsdom` we don't have that problem because it is a separate process.

src/lib/connectors/jsdom/evaluate-runner.ts
qzhou1607: @sarvaje Done.

src/lib/rules/axe/axe.ts
qzhou1607: @molant I put it back as `traverse::end` for now before we decide whether or not we should have a custom event.

molant:
@alrra what do you think about the tagging of this PR? Maybe use a `Fix` instead of `Chore`?

alrra:
> @alrra what do you think about the tagging of this PR? Maybe use a Fix instead of Chore?

It's a user facing thing _("This solves the stuck spinner issue")_ so yes, use `Fix:`.

molant:
> It's a user facing thing ("This solves the stuck spinner issue") so yes, use `Fix:`.

I'll update the commit message when squashing. There is a huge queue on travis an don't want to wait even more to merge this.

src/lib/rules/axe/axe.ts
molant: For reference, the discussion about this is now in https://github.com/sonarwhal/sonar/issues/497

PR #498, opened 2017-09-07
New: Add rule to check the usage of web app manifest's `short_name` & `name`

### LINKED ISSUE #136: Add rule to check the usage of web app manifest's `short_name` & `name`
See also: 

* [Steps for processing the `name` member](https://w3c.github.io/manifest/#name-member)
* [Steps for processing the `short_name` member](https://w3c.github.io/manifest/#short_name-member)

Ref:  #32 

---

Android:

  * There are no differences between what browsers do (tested with Chrome, Opera, and Samsung Internet)
  * Where the text will be truncated depends on:

      * on the length of the characters sed
           
        

      * on the font used
 
         

      * on the size of the app icon + name, or where it's included

            

         

          

Note: The recommendation from the Chrome team [was](https://developer.chrome.com/extensions/manifest/name), and still is less then or 12 characters.  

----

  - [x] While on iOS there is no support for the manifest file, look into how the app text behaves (it's very similar)
  - [x] Edge/Windows?

alrra:
Fix #136


docs/user-guide/rules/manifest-app-name.md
```diff
@@ -0,0 +1,115 @@
...
+{
+    "name": "Baldwin Museum of Science",
+    "short_name": "Baldwin Museum"
+    ...
+}
+```
+
+### Examples that **pass** the rule
+
+Manifest is specified with a `name` shorter than 30 characters
```
molant: Does not having a manifest pass the rule?
If so we should probably tell so.

src/lib/rules/manifest-app-name/manifest-app-name.ts
```diff
@@ -0,0 +1,131 @@
+/**
+ * @fileoverview Check if the web app manifest file includes the `name`
+ * and `short_name` member with appropriate values.
+ */
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const { ucs2 } = require('punycode');
```
molant: `punycode` [is deprecated](https://nodejs.org/dist/latest-v8.x/docs/api/punycode.html)
We should use [Punycode.js](https://github.com/bestiejs/punycode.js)

src/lib/rules/manifest-app-name/manifest-app-name.ts
```diff
@@ -0,0 +1,131 @@
...
+
+import { debug as d } from '../../utils/debug';
+import { IManifestFetchEnd, IResponse, IRule, IRuleBuilder } from '../../types'; // eslint-disable-line no-unused-vars
```
molant: With the latest TS parsers you probably don't need the `//eslint-disable-line no-unused-vars`. I'm doing a PR that removes most of them.

src/lib/rules/manifest-app-name/manifest-app-name.ts
```diff
@@ -0,0 +1,131 @@
...
+
+            const shortName = jsonContent.short_name;
+            const shortNameIsRequired = name && (name.trim() !== '') && (ucs2.decode(name).length > shortNameLengthLimit);
+
+            // Validate 'short_name' if:
+            //
+            //  * it's specified
+            //  * if `name` is over the `short_name` length limit
+
+            if (shortName || shortNameIsRequired) {
```
molant: I'm not sure this condition is right. Maybe something like this to make it clearer?

```js
if (!shortName && !shortNameIsRequired) {
  return;
}

// There is something wrong with `shortName` so we need to check what it is
await checkIfDefined(resource, shortName, 'short_name'); 
await checkIfNonEmpty(resource, shortName, 'short_name');
await checkIfUnderLimit(resource, shortName, 'short_name', shortNameLengthLimit); 
```

tests/lib/rules/manifest-app-name/tests.ts
```diff
@@ -0,0 +1,90 @@
...
+        name: `Manifest is specified with long 'name' and 'short_name'`,
+        reports: [{ message: `Manifest should have 'name' member under 30 characters` }],
+        serverConfig: {
+            '/': htmlWithManifestSpecified,
+            '/site.webmanifest': { content: `{ "name": "1234567890123456789012345678901", "short_name": "test" }` }
+        }
+    },
+    {
+        name: `Manifest is specified with 'name' and 'short_name'`,
+        serverConfig: {
```
molant: Shouldn't this test fail? `name.length` is <= 12

I don't see any `reports`

tests/lib/rules/manifest-app-name/tests.ts
```diff
@@ -0,0 +1,90 @@
...
+            '/': htmlWithManifestSpecified,
+            '/site.webmanifest': { content: `{ "name": "12345678901", "short_name": "🐛🐛🐛🐛🐛🐛🐛🐛🐛🐛🐛🐛" }` }
+        }
+    },
+    {
+        name: `Manifest is specified with 'name' and empty 'short_name'`,
+        reports: [{ message: `Manifest should contain non-empty 'short_name' member` }],
```
molant: Also shouldn't it fail because `short_name` isn't needed?

tests/lib/rules/manifest-exists/tests.ts
```diff
@@ -9,62 +9,62 @@ const htmlWithManifestSpecified = generateHTMLPage(`<link rel="manifest" href="s
...
+        name: `Manifest is specified and the file exists`,
         serverConfig: {
             '/': htmlWithManifestSpecified,
             '/site.webmanifest': ''
         }
     },
     {
-        name: `Web app manifest is specified and request for file fails`,
-        reports: [{ message: `Web app manifest file request failed` }],
+        name: `Manifest is specified and request for file fails`,
```
molant: we should also say in the test name that we expect to pass in this case (and for similar ones where fetching the manifest fails).

src/lib/rules/manifest-app-name/manifest-app-name.ts
alrra: [I'm using Punycode.js](https://github.com/alrra/sonar/blob/c6a5c54fec29410934db53968d1fdfa43d22ad3f/package.json#L110), just that it should be a dependency not a devDependency.

src/lib/rules/manifest-app-name/manifest-app-name.ts
alrra: I'll remove it.

src/lib/rules/manifest-app-name/manifest-app-name.ts
molant: For whatever reason I though we weren't using it before, I didn't see an update in `package.json` and I assumed the worst :P

alrra: > For whatever reason I though we weren't using it before,

We used to use it in a rule, but now it's only used in making sure we can use the [correct amount of emoji in commit messages](https://github.com/sonarwhal/sonar/blob/815dd24072d14ff286031dfe089fd358c676c65d/scripts/check-commit-message.js#L3).

> just that it should be a dependency not a devDependency.

Fixed.

src/lib/rules/manifest-app-name/manifest-app-name.ts
alrra: > remove

Done.

tests/lib/rules/manifest-app-name/tests.ts
alrra: No. If `name` is up to 12, then it can be used instead of`short_name. However, is `short_name` is also specified, it's perfectly valid, and not a problem.

tests/lib/rules/manifest-exists/tests.ts
alrra: Done.

docs/user-guide/rules/manifest-app-name.md
alrra: > If so we should probably tell so.

Done.

alrra:
@molant This PR is ready for another review.

docs/user-guide/rules/manifest-app-name.md
```diff
@@ -0,0 +1,119 @@
...
+```json
+{
+    "name": "Baldwin Museum of Science",
+    "short_name": "Baldwin"
+    ...
+}
+```
+
+Note: [Not specifying a manifest file](manifest-exists.md), or having
+an invalid one are covered by other rules, so those cases won't make i
```
sarvaje: `won't make this rule fail`

alrra: Fixed, thanks!

tests/lib/rules/manifest-app-name/tests.ts
molant: This point of the documentation makes me think it should fail:

> If the `name` is under or 12 characters, there isn't a need to specify `short_name` as browsers can and will just use `name`.

If they will just use `name` if less than 12 characters, then `short_name` shouldn't be there. If they can decide what to use depending on the context and the space (list of apps, title below and icon) then I think this phrase should be reformulated.

alrra: > then I think this phrase should be reformulated.

Any suggestions?

alrra: Done.

PR #499, opened 2017-09-07
Fix: Fix `rawXXX` in `fetchContent` in `chrome`

### LINKED ISSUE #495: `fetchContent` doesn't return any values for `rawResponse` and `rawContent` for Chrome
We are returning `null` for them in [here](https://github.com/sonarwhal/sonar/blob/master/src/lib/connectors/shared/remote-debugging-connector.ts#L798).

We should also add more tests for those properties in `/tests/lib/connectors/fetchContent.ts`

molant:
Fix #495

### DIFF
--- a/src/lib/connectors/shared/remote-debugging-connector.ts
+++ b/src/lib/connectors/shared/remote-debugging-connector.ts
@@ -21,16 +21,16 @@ import { debug as d } from '../../utils/debug';
 import { delay, hasAttributeWithValue } from '../../utils/misc';
 import { resolveUrl } from '../utils/resolver';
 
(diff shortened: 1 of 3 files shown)


src/lib/connectors/shared/remote-debugging-connector.ts
```diff
@@ -21,16 +21,16 @@ import { debug as d } from '../../utils/debug';
 import { delay, hasAttributeWithValue } from '../../utils/misc';
```
alrra: @molant I've fixed `17:13  error  'r' is defined but never used  no-unused-vars` before merging.

Issue #383, opened 2017-07-17
Scan multiple pages at once by pointing to sitemap


dstorey:
When talking to the a11y tools team they mentioned the requirement of being able to scan multiple pages, such as the entire site for issues.

Sites should ideally include a sitemap.xml (or other format, e.g. see https://support.google.com/webmasters/answer/183668?hl=en) for SEO, so it would be great if a user could just point to that file to be able to scan the entire site, rather than having to manually create a list of URLs. 

We could even auto-detect the sitemaps file if it is mentioned in the robots.txt file and offer to scan that rather than just homepage.

molant:
This should go in the `sonar-tools` repo. Closing in favor of https://github.com/sonarwhal/sonar-tools/issues/6

meleyal:
The above link is broken. Is this feature still planned?

alrra:
> The above link is broken.

@meleyal Yes, that repository wasn't very useful. 

>  Is this feature still planned?

There are no plans for now.

For web pages, the discussion was to keep `webhint` as it is and have other tools gather the URLs, pass them to `webhint`, and aggregate the results. As for local files, we have [`connector-local`](https://github.com/webhintio/hint/tree/master/packages/connector-local) that can be used to watch and check local files.

meleyal:
@alrra thanks for the info. 

> For web pages, the discussion was to keep webhint as it is and have other tools gather the URLs, pass them to webhint, and aggregate the results.

Is there any existing work on this somewhere?

alrra:
> Is there any existing work on this somewhere?

@meleyal Not that I know of.

PR #526, opened 2017-09-18
Chore: Add tool to crawl multiple sites with sonar

### LINKED ISSUE #525: Script to run sonar in multiple URLs
It will be good to have a script that allows you to run sonar on multiple websites, and by multiple I mean 1000s. We should use this script to check the reliability of sonar and also bulk find errors on it.

Because this is for reliability, probably it makes more sense to be a separate script and not integrated directly on the cli.

molant:
Fix #525 

@alrra will be nice to get this into Travis and run it periodically somehow. If we end up doing that, we probably want to know how many cores and adapt the number of workers on that (10 might be a bit too much for Travis). It runs a bit slow, so we probably want to trim down the list.

Ideally, this script should be used as followed (from Windows) `node crawler.js urls.txt 2>>errors.txt`

I can also look into adding the errors directly into a file if we think it makes more sense.
Another option to not pollute the output will be to support not having a `formatter` or maybe a `noop formatter`. Don't have any strong opinion about this.


tools/crawler/urls.txt
```diff
@@ -0,0 +1,517 @@
+http://google.com
```
alrra: Maybe drop the protocol from the URLs?

tools/crawler/.sonarrc
```diff
@@ -0,0 +1,31 @@
...
+        "name": "jsdom",
+        "options": {
+            "waitFor": 1000
+        }
+    },
+    "formatters": "summary",
+    "rulesTimeout": 120000,
+    "rules": {
+        "apple-touch-icons": "warning",
+        "axe": "off",
```
alrra: "off" ?

tools/crawler/readme.md
```diff
@@ -0,0 +1,15 @@
+# Description
+
+`crawler` is meant to be used to check the stability and performance
+of `sonar`, mainly on a CI environment.
+
+## Usage
+
+1. Update the `.sonarrc` file to the desired configuration
+1. Update `urs.txt` with the urls to analyze
```
alrra: > urLs

alrra: Note: `urls.txt` is not copied into `dist/`

tools/crawler/crawler.ts
```diff
@@ -0,0 +1,136 @@
+/* eslint-disable no-process-exit */
```
alrra: Since this is for testing, maybe have the `crawler` under [`tests/helpers/`](https://github.com/sonarwhal/sonar/tree/5c3eef7e8da391d6a363e19d507d41c44c8cd149/tests/helpers)?

tools/crawler/readme.md
```diff
@@ -0,0 +1,15 @@
...
+1. Run the following command (assumes code is compiled and in the dist)
+   folder: `node crawler.js urls.txt`
+
+By default `crawler` will scan 10 urls simultaneously. You can change
```
alrra: urls => URLs

tools/crawler/readme.md
```diff
@@ -0,0 +1,15 @@
...
+
+`crawler` is meant to be used to check the stability and performance
+of `sonar`, mainly on a CI environment.
+
+## Usage
+
+1. Update the `.sonarrc` file to the desired configuration
+1. Update `urs.txt` with the urls to analyze
+1. Run the following command (assumes code is compiled and in the dist)
+   folder: `node crawler.js urls.txt`
```
alrra: Maybe have a [`npm` script](https://docs.npmjs.com/misc/scripts)?

alrra:
> It runs a bit slow, so we probably want to trim down the list.

Yes, [Travis CI has time limits](https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts).

> I can also look into adding the errors directly into a file if we think it makes more sense.

For the Travis scenario, we should make @NellieTheNarwhal open an issue if sonar breaks.

tools/crawler/crawler.ts
```diff
@@ -0,0 +1,136 @@
...
+            } else if (code !== 0) {
+                logger.error(`Error runing url: ${url}. Exit code: ${code}`);
+            }
+
+            // We check if there are more urls to process
+            if (urls.length === 0) {
+                resolve();
+            } else {
+
+                await runSonar();
```
sarvaje: Can you try if it is necessary to add a `resolve()` after `await runSonar()`.

tools/crawler/urls.txt
molant: OK, if it doesn't start with `https?` it will add `http`

tools/crawler/.sonarrc
molant: Yes, `axe` timesout a lot with `jsdom` and it slows down the process a lot. The idea of the crawler is mainly te fetch websites, resources, and do the traversing to find reliability problems so this should be ok.

tools/crawler/crawler.ts
molant: It is mainly for testing, but maybe someone will use it for something else. We've had teams inside Microsoft asking for similar things and there is #383 that is somehow related. I was planning on putting those things in there although maybe it's better to create a separate repo for that?

molant:
> For the Travis scenario, we should make @NellieTheNarwhal open an issue if sonar breaks.

Right now it will not break, I process all the URLs no matter what the result is. I should update the exit code then. Is there an easy way to open the issues?

Probably this should be run every time we do a new release and in the same process do a `npm install` to make sure there isn't any problem during the installation?

molant:
@alrra @sarvaje PR updated and ready for another review.

tools/crawler/crawler.ts
alrra: > maybe it's better to create a separate repo for that?

Yes!

molant: Name of the repo?

alrra: > Name of the repo?

Nothing interesting comes to mind, maybe `sonar-testing-ground`?

molant: OK, I'll create a new repo when I finish the current refactor 😊

molant:
Closing this in favor of https://github.com/sonarwhal/sonar-tools/pull/5

PR #531, opened 2017-09-21
Chore: Refactor CLI infrastructure

molant:
This refactor cleans up the current code, making easier to add new
features to the CLI. It also homogenizes how the different actions
are tested.


src/lib/cli.ts
```diff
@@ -150,90 +61,13 @@ export const execute = async (args: string | Array<string> | Object): Promise<nu
...
-    if (configPathFromInput) {
-        config = Config.load(configPathFromInput);
-    } else {
-        config = await loadOrCreateIfNotExits();
-    }
+    let handled = false;
 
-    if (!config) {
-        logger.log(`Unable to find a valid configuration file. Please add a .sonarrc file by running 'sonar --init'. `);
+    while (cliActions.length > 0 && !handled) {
```
sarvaje: Some of the actions can throw exceptions. It think it would be nice to have a `try/catch` wrapping the `while`

molant: Change is pushed. Wrapping individual actions instead of the `while` though.

src/lib/cli/rules/delete-core-rule.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    text = text.replace(`\n${itemToDelete}`, '');
+
+    return util.promisify(fs.writeFile)(indexPath, text, 'utf-8');
+};
+
+/** List of questions to ask the user. */
+const questions = [
+    {
+        message: 'What\'s the file name of this rule to be removed? (No file extension needed)',
```
alrra: Use \`What's the file name...\`.

src/lib/cli/version.ts
```diff
@@ -0,0 +1,19 @@
+import * as path from 'path';
+
+
```
alrra: Remove empty lines.

(I'll updated all these minor things before I merge)

tests/lib/cli/analyze.ts
```diff
@@ -0,0 +1,278 @@
...
+    t.context.config.getFilenameForDirectory.restore();
+    t.context.config.load.restore();
+    t.context.generator.initSonarrc.restore();
+    t.context.spinner.start.restore();
+    t.context.spinner.fail.restore();
+    t.context.spinner.succeed.restore();
+    t.context.inquirer.prompt.restore();
+});
+
+test.serial('if config is not defined, it should get the config file from the directory process.cwd()', async (t) => {
```
alrra: For consistency use `If` instead of `if`, here and everywhere else.

PR #534, opened 2017-09-22
Breaking: Make `rawResponse` a Promise<Buffer>

### LINKED ISSUE #164: Return the right value for `rawResponse` in CDP
Right now ` rawContent`  returns a ` Buffer`  with the right size but no content, and `rawResponse` doesn't return anything.
* [x] `rawContent`
* [x] `rawResponse`

molant:
With `debugging protocol` connectors we need to do another request to
get the raw bytes of the response. With this change `sonar` will only
do the extra request when is needed, caching the result for later use.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Fix #164

### DIFF
--- a/src/lib/connectors/debugging-protocol-common/debugging-protocol-connector.ts
+++ b/src/lib/connectors/debugging-protocol-common/debugging-protocol-connector.ts
@@ -298,13 +298,19 @@ export class Connector implements IConnector {
         }
     }
 
-    private async getResponseBody(cdpResponse) {
+    private async getResponseBody(cdpResponse): Promise<{ content: string, rawContent: Buffer, rawResponse(): Promise<Buffer> }> {
         let content: string = '';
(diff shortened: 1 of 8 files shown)


src/lib/connectors/debugging-protocol-common/debugging-protocol-connector.ts
```diff
@@ -314,22 +320,51 @@ export class Connector implements IConnector {
...
+                rawContent,
+                rawResponse: () => {
+                    const self = (this as any);
+                    const cached = self._rawResponse;
+
+                    if (cached) {
+                        return Promise.resolve(cached);
+                    }
+
+                    if (rawContent.length.toString() === cdpResponse.response.headers['Content-Length']) {
```
alrra: @molant Maybe explain things a bit here? Thanks!

PR #541, opened 2017-09-26
New: Add option to create external rule

### LINKED ISSUE #528: Improve rule/connector developer experience
There are a few things we can do to make developing external sonar resources such as rules or connectors easier for developers.

What we should do:

* [x] Update templates and `--init`: It should create new projects with the right files already in places and the right `requires`. `@sonarwhal/sonar` should be a `devDependency`
* [x] Update the documentation
* [x] We should probably create an example rule that doesn't do anything (maybe just calls `debug`?). We could potentially use it for the template? We probably will have to simplify a bit the `--init` process and just ask for the name in this case and leave the current flow if we detect that we are in the main repo.

molant:
Fix #528

Before merging this:

1. [x] https://github.com/sonarwhal/rule-template/pull/1 needs to be merged
1. [x] project needs to be published to `npm` 
1. [x] need to update the dependencies in here so it works
1. [x] update the documentation


alrra:
> sonarwhal/rule-template#1 needs to be merged

Done.

Will do the others tomorrow (today for me).

alrra:
>  project needs to be published to npm

Done, see: https://www.npmjs.com/package/@sonarwhal/rule-template.

molant:
@alrra @sarvaje this is ready to review!

docs/developer-guide/rules/index.md
```diff
@@ -11,12 +11,113 @@ flexible enough to allow you to implement anything you want easily:
...
+
+There are 2 types of `rule`s a user can develop:
+
+* `external`: These are `rule`s that are published independently. When a
+  `rule` is specific of a domain or use case, it should be external.
+* `core`: These are the `rule`s that are shipped with `sonar` directly.
+  Before starting to develop a `core rule`, please make sure there is
+  an open issue and talk with the maintainers about it.
+
+Both types of `rule`s [work exactly the same](#howruleswork). It is just
```
alrra: Maybe use:

> Both types of `rule`s work exactly the same, the only difference being where they are located.

docs/developer-guide/rules/index.md
```diff
@@ -11,12 +11,113 @@ flexible enough to allow you to implement anything you want easily:
...
+where they are located that changes.
+
+### Creating an external rule
+
+To create a new rule that will be distributed outside `sonar` you will
```
alrra: Maybe use:

> The easiest way to create a new rule...

As users can do it manually if they want to, or what to do it in pure JavaScript. 

---

We should look into suggesting [npx](http://blog.npmjs.org/post/162869356040/introducing-npx-an-npm-package-runner).

molant: > We should look into suggesting npx.

Added, unfortunately [Windows users have problems])(https://github.com/npm/npm/issues/17869) with this option :(

molant:
@alrra all feedback is in

PR #542, opened 2017-09-27
Chore: Create an `enum` for rule category.

sarvaje:
Chore: Remove `fixable` field.

## Pull request checklist

Make sure you:

- [x] Signed the [Contributor License Agreement](https://cla.js.foundation/sonarwhal/sonar)
- [x] Followed the [commit message guidelines](https://sonarwhal.com/docs/developer-guide/contributing/pull-requests.html#commitmessages)

This PR create an `enum` to the category, so all the rules will have the same category with the same capitalization.

This PR also remove the rule property fixable because we are not using it and we are not going to use it shortly.


src/lib/types/rules.ts
```diff
@@ -1,8 +1,13 @@
+import { Category } from '../enums/category';
+
+export type MetadataDocs = {
+    category: Category;
```
molant: Maybe this field should be optional? I'm thinking about the rules that are outside sonar that might have a different category.
Another option will be to have `Category | string` and manually enforce a type when a rule is in core?

sarvaje: How do you enforce a type when a rule is in the core?

sarvaje:
@molant `category` is now optional and conflicts resolved

alrra:
> FIxing typo

I've fixed commit message. :P

molant:
> I've fixed the commit message. :P

I was planning in squashing...

PR #546, opened 2017-09-28
Fix: Make tests more reliable and faster

### LINKED ISSUE #502: strict-transport-security fails randomly when running the tests
image

@alrra and I have seen it happen in both connectors.

molant:
* Remove `concurrency` from `package.json` and let `ava` decide how
  many should be run depending on the CPU
* Pick random port for `test-server` so it doesn't have to look for an
  available one as many times as before
* Fix issue with `static-transport-protocol` where the response could
  be empty and thus producing an exception
* Change test server for `html-checker` and how the html text is
  accessed

Fix #502

### DIFF
--- a/src/lib/rules/html-checker/html-checker.ts
+++ b/src/lib/rules/html-checker/html-checker.ts
@@ -89,10 +89,7 @@ const rule: IRuleBuilder = {
                when testing the rule and `import` doesn't work here. */
             const htmlChecker = require('html-validator');
 
-            /* Property `rawContent` can contain the information with
-             * different encodings depending on the connector we are using on sonar
-             * i.e. In Chrome it is utf8 but in Edge it is base64 */
-            scanOptions.data = response.body.rawContent.toString();
+            scanOptions.data = response.body.content;
             htmlCheckerPromise = scanOptions.data ? htmlChecker(scanOptions) : Promise.resolve({messages: []});
         };
 
--- a/tests/lib/rules/strict-transport-security/tests.ts
+++ b/tests/lib/rules/strict-transport-security/tests.ts
@@ -24,8 +24,8 @@ const requestJSONAsyncMock = (responseObject) => {
         isDataURI() {
             return false;
         },
-        normalizeString(string) {
-            return string.toLowerCase();
+        normalizeString(str = '') {
+            return str.toLowerCase();
         },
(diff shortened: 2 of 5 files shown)


molant:
After removing `concurrency` In my PC the whole `npm test` command runs in under 3minutes. Don't know what the impact will be in Travis and Appveyor. If it isn't good, then we should add `concurrency=2` back for these environments.

molant:
It looks like travis doesn't like the automatic concurrency. Tests were passing before so I'll add it back.

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -89,10 +89,7 @@ const rule: IRuleBuilder = {
                when testing the rule and `import` doesn't work here. */
             const htmlChecker = require('html-validator');
 
-            /* Property `rawContent` can contain the information with
-             * different encodings depending on the connector we are using on sonar
-             * i.e. In Chrome it is utf8 but in Edge it is base64 */
-            scanOptions.data = response.body.rawContent.toString();
+            scanOptions.data = response.body.content;
```
sarvaje: `response.body.content` is always a string?

molant: Yes, that's the type. It could be empty but then it's because it's not HTML and we shouldn't analyze that.

molant:
All lights are green! Anything else I need to address or can this be merged?

sarvaje:
For me is ok!

tests/lib/rules/strict-transport-security/tests.ts
```diff
@@ -24,7 +24,7 @@ const requestJSONAsyncMock = (responseObject) => {
         isDataURI() {
             return false;
         },
-        normalizeString(string) {
+        normalizeString(string = '') {
```
sarvaje: can you change `string` to `str` or anything else, we are using TypeScript and in a first look `string` can look as a type instead of a variable

sarvaje: @molant just this detail. the rest for me is ok to merge

molant: Done, you will have to squash as I've done the change directly on GitHub (in the middle of something else on my laptop)

PR #547, opened 2017-09-28
New: Add `amp-validator` rule

### LINKED ISSUE #545: Integrate AMP validation
Would be great and probably fairly straightforward to integrate AMP validation into sonar, using the [amphtml-validator](https://www.npmjs.com/package/amphtml-validator) on npm.

molant:
Fix #545


molant:
I still need to figure out what's going on in https://github.com/sonarwhal/sonar/issues/545#issuecomment-332929566 and then decide if we remove the custom configuration or not, but @sonarwhal/core can take a look :)

src/lib/rules/amp-validator/amp-validator.ts
```diff
@@ -0,0 +1,91 @@
...
+                if (errorsOnly && error.severity !== 'ERROR') {
+                    debug(`AMP error doesn't meet threshold for reporting`);
+                } else {
+                    await context.report(resource, null, message, null, { column: error.column, line: error.line });
+                }
+            }
+        };
+
+        return {
+            'scan::end': onScanEnd,
```
sarvaje: Why is this event necessary? You already have the html in the event `targetfetch::end`

molant: My idea is to not block the other events and let them fire asap, similar to what we do in `html-checker`, `ssl-labs`, etc. In this case, the creation of a validator via `amphtmlValidator` returns a `Promise` in:

```js
const validPromise = amphtmlValidator.getInstance(); 
```
so I'm waiting for it in the `scan::end` event. I should probably move the creation to `targetfetch::end` or just combine everything on the same handler if we think it's best (personally I prefer to wait until the end to do the check).

sarvaje: Ok! I think is better too do the check at the end

molant: I'll move the promise creation to the `targetfetch::end`. No point in creating it if there is an error in the process.

src/lib/rules/amp-validator/amp-validator.ts
```diff
@@ -0,0 +1,86 @@
...
+            category: Category.performance,
+            description: `Validates the HTML of a page is AMP valid`
+        },
+        recommended: false,
+        schema: [{
+            additionalProperties: false,
+            properties: { 'errors-only': { type: 'boolean' } },
+            type: 'object'
+        }],
+        worksWithLocalFiles: false
```
alrra: Why not?

src/lib/rules/amp-validator/amp-validator.ts
sarvaje: ok!

src/lib/rules/amp-validator/amp-validator.ts
molant: Because that's the option that came by default with the generator and I didn't change it. Will update.

src/lib/rules/amp-validator/amp-validator.ts
```diff
@@ -0,0 +1,86 @@
...
+        const errorsOnly = context.ruleOptions && context.ruleOptions['errors-only'] || false;
+        let html;
+
+        const onTargetFetchEnd = (fetchEnd: ITargetFetchEnd) => {
+            const { response: { body: { content }, statusCode } } = fetchEnd;
+
+            if (statusCode !== 200 || !content) {
+                return;
+            }
+
```
alrra: @molant This rule don't make sense for non-HTML documents, so you should [check if it's a HTML document](https://github.com/sonarwhal/sonar/blob/ebb51bb8b61cecc6a04f3ed80f30ecb7599e7c6a/src/lib/utils/misc.ts#L55-L76) or not.

molant: In my mind everything that comes via the `targetfetch::end` should be `HTML` or at least a string. It doesn't make too much sense, at least for now. Once we start working on the online editor I think that `fetch::end` should be the way to go for the other type of files.
If we agree that anything could come in this event we should do the validation only once somewhere and not in each rule.

docs/user-guide/rules/amp-validator.md
```diff
@@ -0,0 +1,43 @@
...
+## Can the rule be configured?
+
+Yes, you can decide if you want to receive errors only or
+also warnings found by [amphtml-validator][amphtml-validator]. By default
+all warnings and errors are reported. If you prefer to see only the errors
+use the following configuration:
+
+```json
+"amp-validator": ["error", {
+    "errorsOnly": true
```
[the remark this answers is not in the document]
molant: Not sure about the name. We are not `showing`, but rather detecting or something else, hence why I didn't use a verb.

tests/lib/rules/amp-validator/tests.ts
```diff
@@ -0,0 +1,47 @@
+/* eslint sort-keys: 0, no-undefined: 0 */
```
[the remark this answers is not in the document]
molant: Yes it is needed in the list of tests:
```typescript
{
        name: 'Invalid AMP HTML fails',
        serverConfig: invalidAMPHTML,
        reports: [
            { message: `The mandatory attribute '⚡' is missing in tag 'html ⚡ for top-level html'. (https://www.ampproject.org/docs/reference/spec#required-markup)` },
            { message: `The mandatory tag 'html ⚡ for top-level html' is missing or incorrect. (https://www.ampproject.org/docs/reference/spec#required-markup)` }
        ]
    },
```
`reports` should be before `serverConfig`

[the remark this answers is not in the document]
alrra: @molant `no-undefined: 0` can be removed. 

(Sorry I was from my phone when I've written that comment).

src/lib/rules/amp-validator/amp-validator.ts
alrra: > If we agree that anything could come in this event 

My opinion is that should be the case. Users should be allow to provide sonar with any URL, and sonar should handle is appropriately.

I think having this kinda of behavior contributes to a better user experience (same as we don't require users to type the `http(s)://` part and sonar will figure it out).

> we should do the validation only once somewhere and not in each rule.

Can you open an issue with your proposal so that we can discuss it? Thanks!

molant: > My opinion is that should be the case. Users should be allow to provide sonar with any URL, and sonar should handle is appropriately.

Maybe, but in all the documentation we point to websites when using sonar and not individual files. We never talk about analyzing an image for example. IIRC (some?) rules assume that static assets like css, js, and images come in fetch::end and html in targetfetch::end. Connectors definitely assume that, and there isn't a single test for the case you point out.

I agree we should clarify the behavior sooner rather than later, and I'll open an issue for that, but I don't think that should block merging this PR.

alrra: > I agree we should clarify the behavior sooner rather than later, and I'll open an issue for that, but I don't think that should block merging this PR.

:+1:

tests/lib/rules/amp-validator/tests.ts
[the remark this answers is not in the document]
sarvaje: @molant why don't you move the `serverConfig` under `reports`, instead of overwrite the eslint rule?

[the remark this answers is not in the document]
molant: That's how the generator created the rule 🙄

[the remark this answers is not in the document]
sarvaje: and the generator cannot be changed? :P

src/lib/rules/amp-validator/amp-validator.ts
molant: Issue opened: https://github.com/sonarwhal/sonar/issues/550

tests/lib/rules/amp-validator/tests.ts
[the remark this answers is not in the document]
molant: Yes, but we need an issue for that 😜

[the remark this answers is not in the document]
molant: Also, virtually all the test files have the property sorting disabled. IIRC correctly we did that to improve the reading.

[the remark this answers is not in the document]
molant: Here you have the issue: https://github.com/sonarwhal/sonar/issues/551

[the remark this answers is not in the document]
sarvaje: Perfect!

PR #549, opened 2017-09-29
Fix: Make `meta-viewport` rule work with `viewport-fit` property

alrra:
From https://webkit.org/blog/7929/designing-websites-for-iphone-x/

> The first new feature is an extension to the existing viewport meta tag called `viewport-fit`, which provides control over the insetting behavior. `viewport-fit` is available in iOS 11.
>
> The default value of `viewport-fit` is `auto`, which results in the automatic insetting behavior  seen above. In order to disable that behavior and cause the page to lay out to the full size of the screen, you can set `viewport-fit` to `cover`. After doing so, our viewport meta tag now looks like this:
>
> `<meta name='viewport' content='initial-scale=1, viewport-fit=cover'>`

(See also: https://www.w3.org/TR/css-round-display-1/#viewport-fit-descriptor)

Using that `viewport-fit` property currently generates an error because it needs to be added to `metaviewport-parser` (see: https://github.com/dontcallmedom/metaviewport-parser/issues/5).

---

Related: In a discussion with @molant  we agreed that the `meta-viewport` rule should not enforce the usage of  the `viewport-fit` property, as blindly using it [without any other adjustments](https://webkit.org/blog/7929/designing-websites-for-iphone-x/) can potentially create a bad user experience (i.e. part of the website may be obstructed by the sensor housing).

### DIFF
--- a/docs/user-guide/rules/meta-viewport.md
+++ b/docs/user-guide/rules/meta-viewport.md
@@ -193,7 +193,7 @@ If versions of Safari for iOS &lt; 9 are targeted:
     <head>
         <meta charset="utf-8">
         <title>example</title>
-        <meta name="viewport" content="width=device-width, initial-scale=1.0, shrink-to-fit=no">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0, shrink-to-fit=no, viewport-fit=cover">
         ...
     </head>
     <body>...</body>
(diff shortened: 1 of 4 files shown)


alrra:
See also: https://www.w3.org/TR/css-round-display-1/#viewport-fit-descriptor

alrra:
> Using that viewport-fit property currently generates an error because it needs to be added to metaviewport-parser (see: dontcallmedom/metaviewport-parser#5).

I've open a pull request: https://github.com/dontcallmedom/metaviewport-parser/pull/6.

PR #556, opened 2017-10-01
Chore: Update jsdom to 11.3.0

molant:


molant:
Well, it looks like there are still issues with jsdom 11.3.0 :(
Will try to find out what's going on and do a PR if I find a fix

molant:
I think this is a different issue than before. When running the failing tests individually, everything passes. If I run the whole `test` suite then some tests fail randomly with

> TypeError: Cannot read property 'outerHTML' of undefined

Gonna keep trying to find out what's wrong.

PR #560, opened 2017-10-02
Fix: Update `new-core-rule` templates

### LINKED ISSUE #551: Update new core rule template
Things have changed a bit since the template was created. Things to update:

* [x] Remove unnecessary `eslint` comments
* [x] Add missing `;` at the end of `rule`
* [x] Add `debug`
* [x] `category` should use the new `enum`
* [x] `isCore` should be a new option and rename it to `isRecommended`

molant:
Fix #551

### DIFF
--- a/src/lib/cli/rules/templates/core-rule/rule-test.hbs
+++ b/src/lib/cli/rules/templates/core-rule/rule-test.hbs
@@ -1,6 +1,6 @@
-/* eslint sort-keys: 0, no-undefined: 0 */
+/* eslint sort-keys: 0 */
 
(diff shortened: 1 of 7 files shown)


src/lib/cli/rules/templates/core-rule/rule-test.hbs
```diff
@@ -1,6 +1,6 @@
-/* eslint sort-keys: 0, no-undefined: 0 */
+/* eslint sort-keys: 0 */
```
alrra: @molant Can we remove this also and change the order of the properties below? Thanks!

alrra: @molant I did the change before merging.

PR #580, opened 2017-10-17
Docs: Update FAQ with scanner and history

molant:


docs/about/FAQ.md
```diff
@@ -22,6 +22,15 @@ We have a few:
...
   could not be known by everybody.
 
+## Who started sonar?
+
+`sonar`'s development started inside the Microsoft Edge team. Early on,
+the team realized that not only the project had to be open source, but
+also be community driven. The best way to achieve that was by donating
+the code to the JS Foundation, have a governance model that welcomes
+input for anyone in the web community (browser vendors, web experts,
+and developer), and continue the work there.
```
alrra: > developers

docs/about/FAQ.md
```diff
@@ -46,8 +55,17 @@ list.
...
+you can scan an online website in [here][scanner].
+
+We are working on tweaking the list of rules we run and its
+configuration, but if it is a recommended one it will probably be run.
+Also, it uses the `jsdom connector` so there might be unexpected
+results.
+
+If you have any feedback on the results page, please open an
+issue in the [website repository][scanner-issues]. If it is related
+about the results themselves, then in the [sonar repository][sonar-issues].
```
alrra: Maybe use: 

> If the issue is related to the results themselves, then open an issue in the sonar repository.

docs/about/FAQ.md
```diff
@@ -46,8 +55,17 @@ list.
...
 
-Currently there isn't an online service, but having one is part of our
-priorities for the v1.
+Yes! It's in preview so there might be some intermitent issues, but
+you can scan an online website in [here][scanner].
+
+We are working on tweaking the list of rules we run and its
+configuration, but if it is a recommended one it will probably be run.
+Also, it uses the `jsdom connector` so there might be unexpected
+results.
```
alrra: Maybe use:

> We are working on tweaking the list of rules it runs and its
configuration, but if a rule is a recommended one it will probably 
be run. 
> 
> Also, note that the scanner uses the `jsdom connector` so there 
might  be unexpected results.

PR #633, opened 2017-11-03
Fix: Improve third party service integration

molant:
Fix an issue where we could get an exception before reaching the
`try/catch` for the `Promise` that communicates with the 3rd party
service.

## Pull request checklist

Make sure you:

- [X] Signed the [Contributor License Agreement](https://cla.js.foundation/sonarwhal/sonar)

### DIFF
--- a/src/lib/rules/html-checker/html-checker.ts
+++ b/src/lib/rules/html-checker/html-checker.ts
@@ -36,8 +36,10 @@ const rule: IRuleBuilder = {
             format: 'json',
             validator: ''
         };
-        /** If the result messages should be grouped */
+        /** If the result messages should be grouped. */
         let groupMessage: boolean;
+        /** Error processing the request if any. */
+        let failed: boolean = false;
 
         type HtmlError = {
             extract: string; // code snippet
@@ -89,6 +91,11 @@ const rule: IRuleBuilder = {
             };
         };
 
+        const notifyError = async (resource: string, error: any) => {
+            debug(`Error getting HTML checker result for ${resource}.`, error);
+            await context.report(resource, null, `Couldn't get results from HTML checker for ${resource}. Error: ${error}`);
+        };
+
         const start = (data: ITargetFetchEnd) => {
             const { response } = data;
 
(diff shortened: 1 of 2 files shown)


src/lib/rules/html-checker/html-checker.ts
```diff
@@ -100,7 +102,15 @@ const rule: IRuleBuilder = {
             const htmlChecker = require('html-validator');
 
             scanOptions.data = response.body.content;
-            htmlCheckerPromise = scanOptions.data ? htmlChecker(scanOptions) : Promise.resolve({messages: []});
+            htmlCheckerPromise = scanOptions.data ? htmlChecker(scanOptions) : Promise.resolve({ messages: [] });
+            htmlCheckerPromise.catch((error) => {
+                failError = error;
```
sarvaje: why don't you notify the error here, and then just `return` in the `end` method?

src/lib/rules/ssllabs/ssllabs.ts
```diff
@@ -115,7 +117,15 @@ const rule: IRuleBuilder = {
             debug(`Starting SSL Labs scan for ${resource}`);
             scanOptions.host = resource;
 
-            promise = ssllabs(scanOptions);
+            promise = ssllabs(scanOptions)
+                .catch((error) => {
+                    failError = error;
```
sarvaje: the same as before

molant:
This should be ready for review!

molant:
@alrra, @sarvaje can I have some 👀 in this PR?

src/lib/rules/html-checker/html-checker.ts
```diff
@@ -89,6 +91,11 @@ const rule: IRuleBuilder = {
             };
         };
 
+        const notifyError = async (resource: string, error: any) => {
```
alrra: Maybe make this function more generic and move it into utils?

molant: I think it will complicate things because we will have to pass a lot of variables to have the right context. This is how it could look like:

```ts
await notifyError(debug, context, resource, error, `Couldn't get results for RULENAME for ${resource}`);
```

If we start having to do more things when notifying we could look again into that.

alrra: ok.

molant:
I've squashed and rebased. Will merge once all the CI tests pass.

PR #685, opened 2017-12-01
New: Support multiple rules in a package

sarvaje:
Chore: Move `template-rule` repository to sonarwhal
New: Support multiple rules in a package for external rules
generator

## Pull request checklist

Make sure you:

- [x] Signed the [Contributor License Agreement](https://cla.js.foundation/sonarwhal/sonarwhal)
- [x] Followed the [commit message guidelines](https://sonarwhal.com/docs/contributor-guide/contributing/pull-requests/#commitmessages)

For non-trivial changes, please make sure you also:

- [ ] Added/Updated related documentation.
- [x] Added/Updated related tests.

PR #690, opened 2017-12-04
New: Add JSLL script check

qzhou1607:
New Rule to check if JSLL script is in the page head

## Pull request checklist

Make sure you:

* Validate that JSLL script is in `<head>` tag.
* Validate that the JSLL script is placed prior to all other scripts.
* Verify version number format in the file names.

### DIFF
--- a/src/lib/rules/jsll-script-included/jsll-script-included.ts
+++ b/src/lib/rules/jsll-script-included/jsll-script-included.ts
@@ -1,0 +1,112 @@
+/**
+ * @fileoverview This rule confirms that JSLL script is included in the page
+ */
+import { Category } from '../../enums/category';
+import { RuleContext } from '../../rule-context';
+import { IAsyncHTMLDocument, IAsyncHTMLElement, IRule, IRuleBuilder, IElementFound, Severity } from '../../types';
+import { isHTMLDocument, normalizeString } from '../../utils/misc';
+
+/*
+ * ------------------------------------------------------------------------------
+ * Public
+ * ------------------------------------------------------------------------------
+ */
+
+const rule: IRuleBuilder = {
+    create(context: RuleContext): IRule {
+        // Messages.
+        const noScriptInHeadMsg = `No JSLL script was included in the <head> tag.`;
+        const redundantScriptInHeadMsg = `More than one JSLL scripts were included in the <head> tag.`;
+        const warningScriptVersionMsg = `Use the latest release of JSLL with 'jsll-4.js'. It is not recommended to specify the version number unless you wish to lock to a specific release.`;
+        const invalidScriptVersionMsg = `The jsll script versioning is not valid.`;
+        const wrongScriptOrderMsg = `The JSLL script isn't placed prior to other scripts.`;
+
+        const jsllDir = `https://az725175.vo.msecnd.net/scripts/jsll-`;
+
+        const isJsllScript = (element: IAsyncHTMLElement) => {
+            const link = normalizeString(element.getAttribute('src'));
+
+            return link.startsWith(jsllDir);
+        };
+
+        const getJsllScripts = (elements: Array<IAsyncHTMLElement>) => {
+            return elements.filter((element) => {
+                return isJsllScript(element);
+            });
+        };
+
+        const validateHead = async (data: IElementFound) => {
+            const { element, resource }: { element: IAsyncHTMLElement, resource: string } = data;
+
+            if (!isHTMLDocument(resource, context.pageHeaders)) {
+                return;
+            }
+
+            const pageDOM: IAsyncHTMLDocument = context.pageDOM as IAsyncHTMLDocument;
+            const scriptsInHead: Array<IAsyncHTMLElement> = await pageDOM.querySelectorAll('head script');
+            const jsllScriptsInHead: Array<IAsyncHTMLElement> = getJsllScripts(scriptsInHead);
+
+            if (!jsllScriptsInHead.length) {
+                await context.report(resource, element, noScriptInHeadMsg);
+
+                return;
+            }
+
+            if (jsllScriptsInHead.length > 1) {
+                await context.report(resource, element, redundantScriptInHeadMsg);
+
+                return;
+            }
+
+            const jsllScript: IAsyncHTMLElement = jsllScriptsInHead.pop();
+
+            if (scriptsInHead.indexOf(jsllScript) !== 0) {
+                await context.report(resource, element, wrongScriptOrderMsg);
+
+                return;
+            }
+        };
+
+        const validateLink = async (data: IElementFound) => {
+            const { element, resource }: { element: IAsyncHTMLElement, resource: string } = data;
+            const passRegex = new RegExp(`^(\\d+\\.)js`); // 4.js
+            const warningRegex = new RegExp(`^(\\d+\\.){2,}js`); // 4.2.1.js
+
+            if (!isHTMLDocument(resource, context.pageHeaders) || !isJsllScript(element)) {
+                return;
+            }
+
+            const fileName: string = normalizeString(element.getAttribute('src').replace(jsllDir, ''));
+
+            if (passRegex.test(fileName)) {
+                return;
+            }
+
+            if (warningRegex.test(fileName)) {
+                await context.report(resource, element, warningScriptVersionMsg, null, null, Severity.warning);
+
+                return;
+            }
+
+            await context.report(resource, element, invalidScriptVersionMsg);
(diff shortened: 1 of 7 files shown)


molant:
Adding the `blocked` label because this should go into another package as soon as #685 is finished and merged

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+                    return element.getAttribute('src').startsWith(jsllDir);
+                });
+            };
+
+            if (!isHTMLDocument(resource, context.pageHeaders)) {
+                return;
+            }
+
+            const pageDOM: IAsyncHTMLDocument = context.pageDOM as IAsyncHTMLDocument;
+            const scriptsInHead: Array<IAsyncHTMLElement> = await pageDOM.querySelectorAll('head script');
```
molant: We should be using the `element::script` event instead of `traverse::end`.
This should make things slightly faster in some connectors because you will not have to query the DOM to get the elements and you make a better of use of our API.

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+        };
+
+        return { 'traverse::end': validate };
+    },
+    meta: {
+        docs: {
+            category: Category.performance,
+            description: `This rule confirms that JSLL script is included in the head of the page`
+        },
+        recommended: true,
```
molant: should be false

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+            await context.report(resource, null, invalidScriptVersionMsg);
+
+            return;
+        };
+
+        return { 'traverse::end': validate };
+    },
+    meta: {
+        docs: {
+            category: Category.performance,
```
molant: We probably need to create another category. @sarvaje, @alrra, do we have another category for custom rules? Should we create one? Or should the category be empty for custom rules?

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Public
+ * ------------------------------------------------------------------------------
+ */
+
+const rule: IRuleBuilder = {
+    create(context: RuleContext): IRule {
+        const validate = async (data: ITraverseEnd) => {
+            const { resource }: { resource: string } = data;
+            const jsllDir = `https://az725175.vo.msecnd.net/scripts/jsll-`;
+            const passRegex = new RegExp(`^(\\d+\\.)js`);
```
molant: Add an example of a version that passes

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+            const warningRegex = new RegExp(`^(\\d+\\.){2,}js`);
```
molant: Add an example of version that fails

src/lib/rules/jsll-script-included/jsll-script-included.ts
qzhou1607: @molant Hmm... How do you verify that the script is in `head` then?

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+            const jsllScriptsInHead: Array<IAsyncHTMLElement> = getJsllScripts(scriptsInHead);
+
+            if (!jsllScriptsInHead.length) {
+                await context.report(resource, null, noScriptInHeadMsg);
+
+                return;
+            }
+
+            if (jsllScriptsInHead.length > 1) {
+                await context.report(resource, null, redundantScriptInHeadMsg);
```
alrra: Maybe flag each script that shouldn't be included? That way it's easier to know exactly which scripts should be removed.

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+            const warningRegex = new RegExp(`^(\\d+\\.){2,}js`);
+
+            // Messages.
+            const noScriptInHeadMsg = `No JSLL script was included in the <head> tag.`;
+            const redundantScriptInHeadMsg = `More than one JSLL scripts were included in the <head> tag.`;
+            const warningScriptVersionMsg = `Use the latest release of JSLL with 'jsll-4.js'. It is not recommended to specify the version number unless you wish to lock to a specific release.`;
+            const invalidScriptVersionMsg = `The jsll script versioning is not valid.`;
+            const wrongScriptOrderMsg = `The JSLL script isn't placed prior to other scripts.`;
+
+            const getJsllScripts= (elements: Array<IAsyncHTMLElement>) => {
```
alrra: Maybe this should be outside of `validate`?

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+            if (jsllScriptsInHead.length > 1) {
+                await context.report(resource, null, redundantScriptInHeadMsg);
+
+                return;
+            }
+
+            const jsllScript: IAsyncHTMLElement = jsllScriptsInHead.pop();
+
+            if (scriptsInHead.indexOf(jsllScript) !== 0) {
+                await context.report(resource, null, wrongScriptOrderMsg);
```
alrra: Replace `null` with the actual script element.

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+
+                return;
+            }
+
+            const fileName: string = jsllScript.getAttribute('src').replace(jsllDir, '');
```
alrra: `normalizeString(jsllScript.getAttribute('src'))`

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+
+            if (passRegex.test(fileName)) {
+                return;
+            }
+
+            if (warningRegex.test(fileName)) {
+                await context.report(resource, null, warningScriptVersionMsg, null, null, Severity.warning);
```
alrra: Replace `null` with the actual script element.

src/lib/rules/jsll-script-included/jsll-script-included.ts
```diff
@@ -0,0 +1,94 @@
...
+
+                return;
+            }
+
+            await context.report(resource, null, invalidScriptVersionMsg);
```
alrra: Replace `null` with the actual script element.

src/lib/rules/jsll-script-included/jsll-script-included.ts
molant: Vía the traverse::up/down events. If you need more information in those events to make it easier let me know.

qzhou1607: @molant I split the `validate` function into two validate functions, and attach them to `element::head` and `element::script` separately. By listening to the `element::head` event,  it checks the number/location of the jsll scripts without throwing duplicate messages. Does this work?

molant: You are still doing `querySelectorAll` at the page level. If you prefer to continue down this path then do it from the element (do we support `querySelectorAll` at the element level?). If not it is just the same thing as before but being executed in another event.

What I'd do is something like:

1. subscribe to head event or `traverse::down` to the heard
1. set a property `inHead = true`
1. in the `element::script` event, check if you are in the head via the `inHead` property. If not, quit
1. do all the validations you need to do there
1. when you do the `traverse::up`, set `inHead` to `false`

something like:

```js
let inHead = false;

const traverse = (data) =>{
  if( traverseDown of head) {
    inHead = true;
    return;
  }
 
  if (traverseUp of head) {
    inHead = false;
    return;
  }
};

const validateScript = (data) => {
  // validations here taking into account the value of inHead
};
```

Does it make sense?

qzhou1607:
@molant  I added additional `element` information in the `traverse` events and changed the code accordingly. Can you review the PR again? I will squash the commits once it's approved.

molant:
This PR is being handled in the repo for Microsoft rules so closing.

PR #720, opened 2017-12-15
NEW: Add support for parsers and the first parser.

sarvaje:
## Pull request checklist

Make sure you:


src/lib/parsers/javascript/javascript.ts
```diff
@@ -0,0 +1,89 @@
...
+        const scriptData: IScriptParse = {
+            resource,
+            sourceCode: new eslint.SourceCode(code, ast)
+        };
+
+        await this.sonarwhal.emitAsync('parse::javascript', scriptData);
+    }
+
+    private parseJavascript(fetchEnd: IFetchEnd) {
+        if (fetchEnd.response.mediaType !== 'text/javascript') {
```
qzhou1607: What about `application/javascript`? I know a few issues were opened in sonarwhal discussing which media type is right, but here maybe both should be considered?

sarvaje: The `response.mediaType` is calculated in sonar, and in case of a javascript, it will be always `text/javascript`

docs/contributor-guide/parsers/index.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+The way `parser`s receive and share information is via events. To access
+a resource the parser needs to subscribe in the `constructor` to one of
+the `*fetch::end` events, most likely the `fetch::end`. Because that
+event is emitted for all downloaded resources, you need to check first
+if you can understand it. A good approach would be:
+
+1. check first the `mediaType` of the response (`fetchEndEvent#response#mediaType#`)
+2. use a schema to validate if `json` or `xml` or something similar
```
alrra: > `json`, `xml`, or something similar

docs/contributor-guide/parsers/index.md
```diff
@@ -0,0 +1,54 @@
...
+}
+```
+
+The way `parser`s receive and share information is via events. To access
+a resource the parser needs to subscribe in the `constructor` to one of
+the `*fetch::end` events, most likely the `fetch::end`. Because that
+event is emitted for all downloaded resources, you need to check first
+if you can understand it. A good approach would be:
+
+1. check first the `mediaType` of the response (`fetchEndEvent#response#mediaType#`)
```
alrra: Should this be: `fetchEndEvent.response.mediaType`?

sarvaje:
@alrra done!

docs/user-guide/concepts/parsers.md
```diff
@@ -0,0 +1,22 @@
...
+{
+    "parsers": ["parser1", "parser2"]
+}
+```
+
+## List of official `parser`s
+
+The built-in `parser`s are:
+
+* `javascript`: A `JavaScript` parser built on top of `ESLint` so rules for
```
molant: We need to add more information on how a rule can subscribe to this events and create rules based on it.

src/lib/parsers/javascript/javascript.ts
```diff
@@ -0,0 +1,89 @@
+import * as eslint from 'eslint';
+import * as espree from 'espree';
+
+import { determineMediaTypeForScript } from '../../utils/content-type';
+import { IAsyncHTMLElement, IElementFound, IFetchEnd, IScriptParse, Parser } from '../../types';
+import { Sonarwhal } from '../../sonarwhal';
+
+const scriptContentRegex: RegExp = /^<script[^>]*>([\s\S]*)<\/script>$/;
+const defaultParserOptions = {
+    comment: true,
```
alrra: We should add a comment why we are using these options by default, and maybe a link?

docs/user-guide/concepts/parsers.md
sarvaje: Do you have some rule in mind? Can it be addressed in another PR?

src/lib/parsers/javascript/javascript.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private isSrcPresent(element: IAsyncHTMLElement) {
+        const src = element.getAttribute('src');
+
+        return !!src;
+    }
+
+
+    private isJavascript(element: IAsyncHTMLElement) {
```
alrra: Can you rename this to something such as: `isJavaScriptType`.

src/lib/parsers/javascript/javascript.ts
```diff
@@ -0,0 +1,89 @@
...
+            return;
+        }
+
+        const code = fetchEnd.response.body.content;
+        const resource = fetchEnd.resource;
+
+        this.emitScript(code, resource);
+    }
+
+    private isSrcPresent(element: IAsyncHTMLElement) {
```
alrra: can you rename this to: `hasSrcAttribute`?

docs/user-guide/concepts/parsers.md
molant: Documentation on what's exposed via the parser and how to use it should be in this PR IMO.

sarvaje: @molant I have add something else to the documentation.

sarvaje:
@alrra done!

molant:
@sarvaje there are conflicts here. Can you fix them? Also there seems to be errors with AppVeyor?

sarvaje:
@molant I think the error with appVeyor is not related with this PR :s

sarvaje:
@molant conflics solved

molant:
🎉

PR #731, opened 2018-01-03
 New: Add rule to check if resources are served compressed 

### LINKED ISSUE #1: Find a way to (efficiently) detect if a resource was compressed using Zopfli
Since the [`Zopfli`](https://github.com/google/zopfli#readme) output (for the `gzip` option) is valid `gzip` content, there doesn't seem to be a straightforward and foolproof way to identify files compressed with `Zopfli`.

From an email discussion with @lvandeve:

> There is no way to tell for sure. Adding information to the output to indicate zopfli, would actually add bits to the output so such thing is not done :) Any compressor can set the FLG, MTIME, and so on to anything it wants, and users of zopfli can also change the MTIME bytes that zopfli had output to an actual time.
> 
> One heuristic to tell that it was compressed with zopfli or another dense deflate compressor is to compress it with regular gzip -9 (which is fast), and compare that the size of the file to test is for example more than 3% smaller.

---

### `gzip`

A [`gzip` member header](https://tools.ietf.org/html/rfc1952#page-5) has the following structure

```text
  +---+---+---+---+---+---+---+---+---+---+
  |ID1|ID2|CM |FLG|     MTIME     |XFL|OS | (more-->)
  +---+---+---+---+---+---+---+---+---+---+
```

where:

* `ID1` = `1f`  and  `ID2` = `8b` - these are the magic numbers that uniquely identify the content as being `gzip`.
* `CM` = `8`  - this is a value customarily used by `gzip` 
* `FLG` and `MTIME` are usually non-zero values.
* ` XFL` will be either `0`, `2`, or `4`:

   * `0` - default, compressor used intermediate levels of compression (when any of the `-2` ... `-8` options are used).     
   * `2` - the compressor used maximum compression, slowest algorithm (when the `-9` or `--best` option is used).
   * `4` - the compressor used fastest algorithm (when the `-1` or `--fast` option is used).

### `Zopfli`

On thing that `Zopfli` does is that [it sets `FLG` and `MTIME` to zero, `XFL` to `2`,  and `OS` to `3`](https://github.com/google/zopfli/blob/6818a0859063b946094fb6f94732836404a0d89a/src/zopfli/gzip_container.c#L90-L101), so basically files compressed with `Zopfli` will most likely start with `1f8b 0800 0000 0000 0203`, unless things are changed by the user (which in general doesn't seem very likely to happen).

Now, regular `gzip` output might also start with that, even thought the chance of doing so is smaller:

   * Most web servers (e.g.: [Apache](http://httpd.apache.org/docs/current/mod/mod_deflate.html#DeflateCompressionLevel), [NGINX](http://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_comp_level)), by default, will not opt users into the best compression level, therefore, the output shouldn't have `XFL` set to `2`.
   * Most utilities that output regular `gzip` will have non-zero values for `MTIME` and `FLG`.

So, if a file doesn't start with `1f8b 0800 0000 0000 0203`, it's a good (not perfect) indication that `Zopfli` wasn't used, but it's a fast check compared to compressing files and comparing file sizes. However, if a file does start with that, it can be either `Zopfli` or `gzip`, and we cannot really make assumptions here.

### LINKED ISSUE #12: Add rule(s) for HTTP compression
## gzip

1. Check is something is served compressed using `gzip`:

    Make a request  with the `Accept-Encoding: "gzip"` header, then verify if the response is served with the `Content-Encoding: "gzip"` header and the body of the response starts with [`1f 8b`](https://tools.ietf.org/html/rfc1952#page-5).

2. What should be compressed with `gzip`:

    | File type | Commonly used<br>file extension(s) |Commonly used<br>media types(s) |
    |:---|:---|:---|
    | Atom | `.atom` | `application/atom+xml` | 
    | App Cache Manifest | `.appcache` | `text/cache-manifest` |
    | BMP | `.bmp` | `image/bmp` |
    | CSS | `.css` | `text/css` | 
    |  [Cursors Images](https://en.wikipedia.org/wiki/ICO_(file_format)) | `.cur` | `image/x-icon`<br>`image/vnd.microsoft.icon`|
    | Embedded OpenType font | `.eot` |  `application/vnd.ms-fontobject` |
    | [Favicon](https://en.wikipedia.org/wiki/Favicon) | `.ico` | `image/x-icon`<br>`image/vnd.microsoft.icon`|
    | HTML | `.html`<br>`.htm`<br>... | `text/html`<br>`application/xhtml+xml` |
    | [HTML Components](https://en.wikipedia.org/wiki/HTML_Components) | `.htc` | `text/x-component` |
    | JavaScript | `.js` | `application/javascript`<br>`text/javascript` | 
    | JSON | `.json`<br>... | `application/json`<br>`application/<something>+json`|
    | OpenType  font |  `.otf` | `font/opentype` |
    | RDF | `.rdf` | `application/rdf+xml` |
    | RSS | `.rss` | `application/rss+xml` |
    | Source Maps | `.map` | `application/json` |
    | SVG | `.svg` | `image/svg+xml` |
    | TrueType font | `.ttc`<br>`.ttf` | `application/x-font-ttf` |
    | TXT | `.txt` | `text/plain` |
    | vCard | `.vcard`<br>`vcf` | `text/vcard` |
    | VTT | `.vtt` | `text/vtt` |
    | XML | `.xml`<br>... | `application/xml`<br>`text/xml`<br>`application/<something>+xml` |
    | Web App Manifest | `.webmanifest`<br>`.json` | `application/manifest+json` |

    Notes:
     
      * Under 1K things should not necessarily be compressed.
      * `WOFF` fonts should not be compressed (see: https://github.com/h5bp/server-configs-apache/issues/42).
      * `SVGZ` should be served with the `Accept-Encoding: "gzip"` header as they are compressed by default.
       

## Zopfli

1. Check is something is served compressed using `Zopfli`:

    Same as with `gzip`, just that we need to [detect `Zopfli`](https://github.com/MicrosoftEdge/Sonar/issues/1).

2. What should be compressed with `Zopfli`:
   
    Same as with `gzip`.

    Note(s):
   
      * We will need to a check if `WOFF` fonts use `Zopfli` compression internally.

## Brotli

1. Check is something is served compressed using `Brotli`.

    Make a request  with the `Accept-Encoding: "br"` header, then verify if the response is served with the `Content-Encoding: "br"` header ([no magic numbers?](https://github.com/google/brotli/issues/298#issuecomment-172549140)).

   Note: `Brotli` compressed responses should be [served only over HTTPS](https://medium.com/@yoavweiss/well-the-technical-reason-for-brotli-being-https-only-is-that-otherwise-there-s-a-very-high-508f15f0ad95).

2. What should be compressed with `Brotli`:
  
    Same as with `gzip`.

--- 

Note: This is just a starting point, we will probably split this into more specific issues / rules.

alrra:
## Pull request checklist

Make sure you:

Fix #1
Fix #12


src/lib/rules/http-compression/http-compression.ts
```diff
@@ -645,6 +645,7 @@ const rule: IRuleBuilder = {
             category: Category.performance,
             description: 'Require resources to be served compressed'
         },
+        ignoredConnectors: ['chrome'], // Headless Chrome on Travis CI fails miserably. :(
```
alrra: Hopefully this is temporary, all the test pass even for the chrome connector on my machine. :(

alrra:
FYI: I've merged some of the fixes from this PR into `master`.

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  a bigger file size.
+
+  The same goes for resources that are very small because of the
+  overhead of compression file formats.
+
+* Use most efficien compression method.
+
+  `gzip` is the most used encoding nowadays as it strikes a good
+  balance between compression ratio (as high as 70-90% especially
+  for larger files) and encoding time, and is supported pretty much
```
molant: Do you have a link for the 70-90% claim?

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  eveywere.
```
molant: > everywhere

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+
+  Better savings can be achieved using [`Zopfli`][zopfli] which
+  can reduce the size on average with [3–8% more then `gzip`][[zopfli
```
molant: > on average 3-8% more than `gzip`

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  blog post]. Since `Zopfli` output (for the `gzip` option) is valid
+  `gzip` content, `Zopfli` works eveywere `gzip` works. The only
+  downsize is that encoding takes way more time than with `gzip`,
```
molant: > encoding takes ~~way~~ more time

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  thus, making `Zopfli` more suitable for static content (i.e. encoding
+  resources with is as part of build script, not on the fly).
```
molant: > ~~with is~~ as part of a build script

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+
+  But, things can be improved even futher using [Brotli][brotli]. This
+  new encoding allows to get [20–26% higher compression ratios][brotli
```
molant: Maybe remove `new`?  Don't know when we will review this docs and it's been out for some time already.

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  blog post] even over `Zopfli`. However, being a new encoding and
```
molant: In the same spirit as the previous comment:

> However, this encoding is not compatible with `gzip`, limiting the support to modern browsers...

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  that is not compatible with `gzip`, limits its support to modern
+  browsers, and its usage to [only over HTTPS (as proxies misinterpreting
+  unknown encodings)][brotli over https].
+
+  So, in general, for best performance and interoperability resources
+  should be served compress with `Zopfli` over HTTP and `Brotli` over
+  HTTPS.
```
molant: Maybe clarify that `Brotli` shouldn't be the only encoding over HTTPS?

> and `Brotli` over HTTPS with a fallback to `Zopfli` if not supported

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  `compress`, `bzip2`, [`SDCH`][unship sdch], etc.), and/or may not
+  be as efficient, or can create problems (e.g.: [`deflate`][deflate
+  issues]). In general these should be avoided, and one should just
+  stick to the encoding specified in the previous point.
+
+* Avoid potential caching related issues.
+
+  When resources are served compressed, they should be served with
+  the `Vary` header containing the `Accept-Encoding` value (or with
+  something such as `Cache-Control: private` that prevents caching
```
molant: > that prevents caching in proxy caches and such altogether
 
The final user can cache the response so we are not preventing caching altogether:

> Indicates that all or part of the response message is intended for a single user and MUST NOT be cached by a shared cache, such as a proxy server.

[source](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html)

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  This needs to be done in order to avoid problems such as an
+  intermediate proxy caching the compress version of the resource and
+  then sending it to all user agents regardless if they support that
+  particular encoding or not, or if they even want the compressed
+  version or not.
+
+* Resources should be served compressed only when requested as such,
+  appropriately encoded, and without relying on user agent sniffing.
+
+  What the `Accept-Encoding` request header specified should be
```
molant: > ~~What~~ The `Accept-Encoding`

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  respected. Sending a the content encoded with a different encoding
```
molant: > Sending a ~~the~~ content

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+  then one of the ones accepted can lead to problems.
```
molant: > than one

docs/user-guide/rules/http-compression.md
```diff
@@ -0,0 +1,567 @@
...
+...
+Content-Encoding: gzip
+Content-Type: image/svg+xml
+
+<SVGZ content>
+```
+
+## Can the rule be configured?
+
+You can overwrite the defaults by specifying what type of compression
```
molant: > override

src/lib/rules/http-compression/http-compression.ts
```diff
@@ -0,0 +1,672 @@
+/**
+ * @fileoverview Check is resources are served compressed.
```
molant: > Checks if resources

src/lib/rules/http-compression/http-compression.ts
```diff
@@ -0,0 +1,672 @@
...
+             *       2 - the compressor used maximum compression,
+             *           slowest algorithm (when the -9 or --best
+             *           option is used).
+             *
+             *       4 - the compressor used fastest algorithm (when
+             *           the -1 or --fast option is used).
+             *
+             * 2. Zopfli
+             *
+             *    On thing that Zopfli does is that it sets FLG and
```
molant: > One thing

src/lib/rules/http-compression/http-compression.ts
```diff
@@ -0,0 +1,672 @@
...
+                return;
+            }
+
+            // - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+            // Check compressed vs. uncompressed sizes.
+
+            /*
+             * TODO: Remove the following once connectors
+             *       support Brotli compression.
```
molant: Can you open an issue for this?
Basically what we need is `requester` to auto decompress `Brotli`.

src/lib/rules/http-compression/http-compression.ts
```diff
@@ -0,0 +1,672 @@
...
+        const generateDisallowedCompressionMessage = (encoding: string) => {
+            return `Disallowed compression method: '${encoding}'.`;
+        };
+
+        const generateContentEncodingMessage = (encoding: string, notRequired?: boolean, suffix?: string) => {
+            return `Should${notRequired ? ' not' : ''} be served with the 'content-encoding${encoding ? `: ${encoding}` : ''}' header${suffix ? ` ${suffix}` : ''}.`;
+        };
+
+        const generateCompressionMessage = (encoding: string, notRequired?: boolean, suffix?: string) => {
+            return `Should${notRequired ? ' not' : ''} be served compressed${encoding ? ` with ${encoding}` : ''}${suffix ? ` ${suffix}` : ''}.`;
```
molant: We should make the message clearer.
Right now if a resources is not compressed the user is going to get 2 messages:

> Should be served compressed with Zopfli
> Should be served compressed with Brotli

which can be confusing. A way could be to say the protocol as you proposed the other day:

> Should be served compressed with Zopfli for HTTP
> Should be served compressed with Brotli for HTTPS

docs/user-guide/rules/http-compression.md
alrra: Fixed.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

alrra: Also, there is another issue here: double `[[`.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

docs/user-guide/rules/http-compression.md
alrra: Done.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

docs/user-guide/rules/http-compression.md
alrra: Fixed.

src/lib/rules/http-compression/http-compression.ts
alrra: Fixed.

src/lib/rules/http-compression/http-compression.ts
alrra: > Can you open an issue for this?

[Done](https://github.com/sonarwhal/sonarwhal/issues/735).

docs/user-guide/rules/http-compression.md
alrra: Fixed.

src/lib/rules/http-compression/http-compression.ts
alrra: Fixed.

(Note: Only changed the message for `Brotli`, since `Zopfli` applies to both)

alrra:
@molant I addressed all your feedback!

src/lib/rules/http-compression/http-compression.ts
```diff
@@ -0,0 +1,673 @@
...
+             */
+
+            /*
+             * Check if the media type is one of the common
+             * ones for which it is known the response should
+             * be compressed.
+             */
+
+            if (isTextMediaType(mediaType) ||
+                ['image/x-icon',
```
sarvaje: I think you should have these kind of arrays of the method as a constant.

alrra: Fixed.

molant:
🎉

PR #736, opened 2018-01-05
New: Upgrade `axe-core` to `v2.6`

alrra:


molant:
Why didn't we get any notification from greenkeeper?

alrra:
> Why didn't we get any notification from greenkeeper?

Because it's not a major version.

PR #763, opened 2018-01-17
New: Add interactive formatter

### LINKED ISSUE #761: Add interactive formatter
Things that need to be done here:
* Group results by category/domain.
* The user can select to expand/collapse some category/domain results.
-----
I looked into the interactive formatter this morning and couldn't find a package that allows up to expand/collapse the scan results in place. I wonder if the alternative solution below works:
1.	After a scan, the results are grouped by category/domain (depending on users’ setting) and presented in the format of `summary`.
2.	A question is prompted at the end of the summary list, asking user to select the category/domain results they want to expand. (Maybe the summary list and the inquirer `choices` can be combined, since they are presenting the same content).
3.	The selected category/domain results are expanded. At the end of the expanded result, a question is prompted asking if the user wants to go back to the `main menu` (the scan result summary), if yes, it goes back to 2, if no, exit.

qzhou1607:
## Pull request checklist

Make sure you:

1. Add the interactive formatter
2. Refactor the shared code into a separate file

Fix #761


qzhou1607:
Preview:

interactive formatter_2

src/lib/formatters/interactive/interactive.ts
```diff
@@ -0,0 +1,124 @@
...
+                    errors += stats.totalErrors;
+                });
+
+
+                reportSummary(errors, warnings);
+            });
+        };
+
+        const askAndDisplay = async (categories: _.Dictionary<Array<IProblem>>, formattedRows: Array<string>, categoryIds: Array<string>, checkedCategories?: inquirer.Answers) => {
+            const choices: Array<inquirer.ChoiceType> = _.map(formattedRows, (row: string) => {
```
sarvaje: `_.map` give you the index in a second parameter.
`_.map(formattedRows, (row: string, index: number) => {`

qzhou1607:
Both the `group by category` and `group by domain` options are done. Users can select which to use by setting formmater as `"formatters": "interactive: category"` in `.sonarwhalrc`.

molant:
@qzhou1607 can you update the documentation? We should have a `gif`  with the final version in the list of formatters. Also you need to document how to configure it.

What happens if I put several formatters in the configuration?

alrra:
> Will it be possible to simplify a bit the flow too?
>
> Ideally we should have the list of categories/domains and as soon as we select one the list gets expanded.

I agree, we should have something as the typical collapsible list.


Note: In order to avoid having something like  `**► b**`, I highlight the current selected item with an uppercase equivalent letter, but in the code we can highlight it making the text bold or another color.

---

Let's say we have

```
► A
► b
► c 
```

If the user navigates to `b`

```
► a
► B
► c 
```
and presses enter/space, `b` extends

```
► a
▼ B
     b1
     b2 
     ...
► c 
```

If the user navigates back to `a`

```
► A
▼ b
     b1
     b2 
     ...
► c 
```

and presses enter/space, `a` extends

```
► A
    a1
    a2 
     ...
▼ b
     b1
     b2 
     ...
► c 
```

If the user goes back to `b` and presses enter/space,  `b` collapses

```
▼ a
    a1
    a2 
     ...
► B
► c 
```

qzhou1607:
@molant I wish I could implement the flow like @alrra listed, but I couldn't really find an existing package to do so 😟 . So I went on with `inquirer` instead, which unfortunately waits to display the results after the user finishes the selection. I will do more research about this, please do let me know if you have more information.
> ideally we should have the list of categories/domains and as soon as we select one the list gets expanded.

alrra:
@qzhou1607, @molant If there isn't a package to help abstract some of the things, I'm sure this will take us some time to build. 

Maybe it would be better to just have a HTML formatter? That way we can, for example, inject an object with the result in a template page, and have have that page allow users to sort things based on different categories, not just one.

alrra:
> and have have that page users to sort things based on different categories, not just one.

Also that code could be shared with the online scanner.

qzhou1607:
This is what it looks like after the pushed changes. Will finalize the documentation if the current flow works with everyone.

interactive formatter_domain

molant:
We are going to prioritize the HTML formatter that should be more helpful so closing this PR.

Issue #748, opened 2018-01-09
Publish rules as standalone npm packages


alrra:
- [x] Refactor code to make this possible.
- [x] Publish rules as standalone `npm` packages.
- [x] Update infrastructure and release script.
- [x] Update the wizard logic that displays the rules that can be installed. I'm thinking about doing an npm search and all rules that are under `@sonarwhal` should be in there
- [x] Delete the logic for core/no-core rules
- [x] Update the skeleton of the rules to include `nyc`

molant:
For context, some of the reasons why we want to do this:

* shorter and more reliable builds
* update rules without having to update sonarwhal
* be prepared to install rules on demand on first run

molant:
Some other things that we need to do in here:

* [ ] Update the wizard logic that displays the rules that can be installed. I'm thinking about doing an npm search and all rules that are under `@sonarwhal` should be in there
* [ ] Delete the logic for core/no-core rules
* [ ] Probably something else?

alrra:
Note: We are working on this in the [`fix-748` branch](https://github.com/sonarwhal/sonarwhal/tree/fix-748).

molant:
Making a list to know what rules are passing or not with the current project. The ones that aren't checked and don't have a comment I haven't tested them yet.

* [ ] amp-validator: has an issue installing on windows in the post-install script
* [X] apple-touch-icons
* [X] axe
* [X] content-type
* [X] disown-opener
* [X] highest-available-document-mode
* [ ] html-checker: 2 failing tests with chrome
* [X] http-cache
* [X] http-compression: 63 skipped tests (chrome is ignored)
* [X] image-optimization-cloudinary: 6 skipped tests (guess chrome is ignored)
* [X] manifest-app-name
* [X] manifest-exists
* [X] manifest-file-extension
* [ ] manifest-is-valid
* [ ] meta-charset-utf-8
* [ ] meta-viewport
* [ ] no-disallowed-headers
* [ ] no-friendly-error-pages
* [ ] no-html-only-headers
* [ ] no-http-redirects
* [ ] no-protocol-relative-urls
* [ ] no-vulnerable-javascript-libraries
* [ ] ssllabs
* [ ] strict-transport-security
* [ ] validate-set-cookie-header
* [ ] x-content-type-options

molant:
Forgot to mention the results above are with my current work. I had to fix something on jsdom when accessing the attributes so you might not get the same until my work is merged.

[...]

Issue #718, opened 2017-12-15
Allow CLI to be run without .sonarwhalrc in current directory


poshaughnessy:
Proposal: Given that the recommended use of the CLI is to point to a URL (rather than a local project), I would suggest we do not need to pollute the current directory with a `.sonarwhalrc` file.

How about...

1. If no `.sonarwhalrc` found in the current directory, look for `.sonarwhalrc` in user's home directory.
2. If no `.sonarwhalrc` found in user's home directory either, proceed with default configuration? (With an appropriate warning / note?). This would have the added benefit of reducing a necessary step (`sonarwhal init`) for first time users to try out the CLI.

(This may be related to #713?)

alrra:
> I would suggest we do not need to pollute the current directory with a .sonarwhalrc file.
>
> If no .sonarwhalrc found in the current directory, look for .sonarwhalrc in user's home directory.

I personally agree that we should do that.

(I don't remember why we didn't do it that way in the first place 🤔)

> If no .sonarwhalrc found in user's home directory either, proceed with default configuration? (With an appropriate warning / note?).

As long as we make it clear (with an appropriate warning/note as you specified), I'm personally ok with it.

---

@sonarwhal/core Thoughts?

alrra:
@poshaughnessy, @diekus, @torgo 

I was discussing this issue with @molant, and we think this can be a good first issue to solve that will also allow you to better understand how `sonarwhal` works.

Do any of you want to take this issue? We can help you along the way!

poshaughnessy:
@alrra Sorry for the delay. If it's not too late, yes, let's give this a try! I'll see how far I get and hopefully can put in an initial PR for review and feedback :)

alrra:
> If it's not too late, yes, let's give this a try!

Go for it, and don't hesitate to let us know if we can help!

molant:
@poshaughnessy we've separated everything into packages now so if you want to give this a try let us know. Otherwise we will probably take care of this in the next few days.
There have been some changes in the projects structure so make sure to check the contributor guide (PR for those changes is https://github.com/sonarwhal/sonarwhal/pull/827 in case @alrra doesn't get to merge it before you read this).

PR #785, opened 2018-01-27
New: Ability to search npm packages

sarvaje:
## Pull request checklist

Make sure you:

This PR allow us to get the npm packages with rules/parsers/etc. directly from npm, and it will be useful for futures updates in sonarwhal.


src/lib/utils/resource-loader.ts
```diff
@@ -275,3 +279,57 @@ export const loadFormatter = (formatterId: string): IFormatter => {
...
+export const getExternalRulesFromNpm = async () => {
+    const rules = await getNpmPackages('sonarwhal-rule');
+
+    return filterPackages(rules, 'sonarwhal-rule');
+};
+
+export const getCoreRulesFromNpm = async () => {
+    const rules = await getNpmPackages('@sonarwhal/rule');
+
+    return filterPackages(rules, '@sonarwhal/rule');
```
molant: Can you put a comment explaining why we are filtering even though you are searching for `@sonarwhal/rule`?

Maybe we could also do a search just by `@sonarwhal`  and then do the grouping in memory by rule/parser/connector/etc.?

src/lib/types.ts
```diff
@@ -74,3 +74,17 @@ export interface IORA {
     fail(): void;
     text: string;
 }
+
+export type NpmMaintainer = {
```
molant: Can you add some descriptions?

molant:
This is in preparation for #748. We will merge this PR once the other one is merged.

The idea is to replace the results of getting the installed core rules with the npm search. The reason is that there will not be any rule installed by default.

Some things to take into account while this PR continues:

* We will probably have issues when installing `sonarwhal` globally because we might need `sudo` powers to install the rules. I think that ideally we should have a `post-install` task so the user can at least install a few ones during the initial process.
* If the user runs `--init` when installed globally (after install) then we can tell them to run `sudo npm rulesNames` if the process fails
* This will probably affect the work of #718 so @poshaughnessy you might want to hold on that until we finish this refactor (or we can take care of migrating if it lands before the rest)

@sarvaje, any other thing that we discussed that I forgot?

poshaughnessy:
@molant re. #718, no worries - sorry, I got delayed by other work again anyway! So holding off until this refactor is finished sounds good to me. I can start investigating / preparing in the meantime.

src/lib/utils/resource-loader.ts
sarvaje: I don't know if it is because I'm doing something weird or not, but the search doesn't work like you expect to work. With the packages there is right now, search for `@sonarwhal` or `sonarwhal` return the same elements (and not, not all the packages start with @sonarwhal`.

About the filter, I will write some explanation, but look for `@sonarwhal/rule` doesn't have to return only packages that start with `@sonarwhal/rule` :(

src/lib/types.ts
sarvaje: ok

PR #798, opened 2018-02-06
New: Rule `performance-budget`

### LINKED ISSUE #790: New rule: Web performance budget
# Description

> A web performance budget is a group of limits to certain values that affect site performance that should not be exceeded in the design and development of any web project. This could be the total size of a page, size of images you are uploading, or even the number of HTTP requests that your webpage generates.

https://www.keycdn.com/blog/web-performance-budget/

# Implementation

Ideally we should throttle to get real data. Doing so can be really complicated, especially cross-connector and cross-platform. Also, the results can change a lot from one run to another and thus confusing the user.
 
Instead, we can take a more __naive__ approach and do the math ourselves using established metrics and preconfigured speeds. 
I like Alex Russell's approach to calculate the size that can be delivered over the wire in [Can You Afford It?: Real-world Web Performance Budgets](https://infrequently.org/2017/10/can-you-afford-it-real-world-web-performance-budgets/)

> We can start with our first-load budget of 5 seconds and begin to calculate how much transfer we can afford.
> First we subtract 1.6 seconds from our budgets for DNS lookup and TLS handshaking, leaving us 3.4s to work with.
> Then, we calculate how much data we can send over this link in 3.4 seconds: 400 Kbps = 50KB/s. 50KB/s * 3.4 = 170KB.
>
> RTT: round-trip-time
> TTI: time-to-interactive
> 
>| Speed | RTT | TTI | Resources size |
>|--------|------|------|------------------|
>| 400Kbps | 400ms | 5s | 130-170 KB  | 

We should use [WebPageTest network configurations](https://github.com/WPO-Foundation/webpagetest/blob/master/www/settings/connectivity.ini.sample) so there's a unified message across similar tools.

We should be using a mix of `rawResponse`, to get the size sent, and `content` length.  There could be some differences until we implement #735 (brotli on in `requester`) but it should give us a good enough value to go from there.

Default speed should be 3G to [align with Lighthouse](https://github.com/GoogleChrome/lighthouse/blob/bedb9a153eef670d1fc6c22aaac5ef0b243a1607/docs/throttling.md#the-3g-network-throttling-preset) and load time 5 secs to align with Alex's guidance. 

The biggest concern with this approach is calculating how much time we should allocate to DNS lookup and TLS handshaking.

Additionally, we should probably take into account the size of the uncompressed JavaScript to add it to the total time spent. If we decide to do this, we need to look into averages of parsing and executing JS.

# User configuration

User should be capable to change:
* the connection speed from the list of values proposed by WebPageTest
* the load time for the site

# Further reading

* https://medium.com/dev-channel/the-cost-of-javascript-84009f51e99e
* https://infrequently.org/2017/10/can-you-afford-it-real-world-web-performance-budgets/
* https://www.keycdn.com/blog/web-performance-budget/

molant:
## Pull request checklist

Make sure you:

Fix #790


.sonarwhalrc
```diff
@@ -2,37 +2,38 @@
     "connector": {
         "name": "chrome",
         "options": {
-            "waitFor": 1000
+            "waitFor": 5000
```
alrra: This was committed by mistake?

docs/user-guide/rules/performance-budget.md
```diff
@@ -0,0 +1,147 @@
...
+> are uploading, or even the number of HTTP requests that your webpage
+> generates.
+
+https://www.keycdn.com/blog/web-performance-budget/
+
+## Why is this important?
+
+As of January 2018, the average size of a website is 3,545kB:
+
+![image](https://chart.googleapis.com/chart?chs=400x225&cht=p&chco=007099&chd=t:1818,70,98,504,120,851,27&chds=0,1818&chdlp=b&chdl=total%203545%20kB&chl=Images+-+1818+kB%7CHTML+-+70+kB%7CStylesheets+-+98+kB%7CScripts+-+504+kB%7CFonts+-+120+kB%7CVideo+-+851+kB%7COther+-+27+kB&chma=|5&chtt=Average+Bytes+per+Page+by+Content+Type)
```
alrra: Move the URL at the bottom.

docs/user-guide/rules/performance-budget.md
```diff
@@ -0,0 +1,147 @@
...
+
+Although the global average connection is 7.2Mb/s (check [_Akamai's state of
+the Internet 2017_][state of the internet]), "no bit is faster than one that is
```
alrra:
```text
_"no bit is faster than one that is not sent"_
```

docs/user-guide/rules/performance-budget.md
```diff
@@ -0,0 +1,147 @@
...
+       (total number of requests * TCP slow-start phase) +
+       (total size of resources / bandwidth)
+```
+
+This is the list of things taken into account:
+
+* Everything is a first load, no values are cached, and no connections are
+  opened.
+* [`RTT` (Round-Trip Time)][rtt] is fixed and changes depending on the
+  configured network (more about this in #). It assumes all servers respond
```
alrra: in #?

docs/user-guide/rules/performance-budget.md
```diff
@@ -0,0 +1,147 @@
...
+the Internet 2017_][state of the internet]), "no bit is faster than one that is
+not sent" (quote by [Ilya Grigorik][faster bit]). Web developers need to be
+mindful not only about the size of their sites, but also the number of
+requests, different domains, third party scripts, etc. The time required by a
+browser to download a 200kB file is not the same than 20 files of 10kB.
+
+## What does the rule check?
+
+This rule calculates how long it will take to download all the resources loaded
+initially by the website under a `3G Fast` network. If the load time is
```
alrra: Maybe add something in the lines of: `3G Fast` network (but that can be changed, see "Can the rule be configured?" section.

docs/user-guide/rules/performance-budget.md
```diff
@@ -0,0 +1,147 @@
...
+
+`loadTime` has to be a number greater than `1` and indicates the time in
+seconds to load all the resources.
+
+### Examples that **trigger** the rule
+
+Assuming a `3G Fast` connection, the following will fail:
+
+* A site with 2 redirects and 2 images over 362KB each
+* A site with just HTML, loaded over https, with a size over 1MB
```
alrra: HTTPS

docs/user-guide/rules/performance-budget.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Examples that **pass** the rule
+
+* A site with 2 images smaller than 350KB
+* A site with just HTML, loaded over https, with a size less than 900KB
```
alrra: HTTPS

src/lib/rules/performance-budget/performance-budget.ts
```diff
@@ -0,0 +1,284 @@
+/**
+ * @fileoverview Performance budget checks if your site will load fast enough based on the size of your resources and a given connection speed
+ */
+
+import * as url from 'url';
+
+import { Category } from '../../enums/category';
+import { RuleContext } from '../../rule-context';
+// The list of types depends on the events you want to capture.
```
alrra: Remove this line.

src/lib/rules/performance-budget/performance-budget.ts
```diff
@@ -0,0 +1,284 @@
...
+ */
+
+const rule: IRuleBuilder = {
+    create(context: RuleContext): IRule {
+
+        /** An array containing all the responses. */
+        const responses: Array<ResourceResponse> = [];
+        /** Set with all the different domains loaded by the site. */
+        const uniqueDomains: Set<string> = new Set();
+        /** Set with all the different https domains loaded by the site. */
```
alrra: HTTPS

docs/user-guide/rules/performance-budget.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Can the rule be configured?
+
+You can change the type of connection and/or the target load time as follows:
+
+```json
+{
+    "performance-budget": ["error", {
+        "connectionType": "Dial",
+        "loadTime": 10
```
alrra: Maybe document what happens if `loadTime` is not specified.

docs/user-guide/rules/performance-budget.md
```diff
@@ -0,0 +1,147 @@
...
+| Dial   |  49 Kbps |  30 Kbps | 120ms |
+
+`loadTime` has to be a number greater than `1` and indicates the time in
+seconds to load all the resources.
+
+### Examples that **trigger** the rule
+
+Assuming a `3G Fast` connection, the following will fail:
+
+* A site with 2 redirects and 2 images over 362KB each
```
alrra: I think these examples are a bit confusing. Maybe also specify the HTML and it's size?

docs/user-guide/rules/performance-budget.md
alrra: So, restricting the the today size will be part of another rule?

molant: > restricting the the total size will be part of another rule?

I'm not sure about that. All things are connected (size, load time, network type, number of requests, etc.). I'd rather make this rule more intelligent (e.g.: take into account JS parse time) than creating a new one that just checks the size without any other variable (like connection speed).

src/lib/rules/performance-budget/connections.ini
```diff
@@ -0,0 +1,91 @@
+# Original file from WebPageTest (https://github.com/WPO-Foundation/webpagetest/blob/master/www/settings/connectivity.ini.sample)
```
qzhou1607: This might be a stupid question: Why not just use `json` file instead of a text file, then we don't need the parser?

src/lib/rules/performance-budget/connections.ts
```diff
@@ -0,0 +1,80 @@
...
+    lines.forEach((line) => {
+        const [key, value] = line.trim().split('=');
+
+        if (key.startsWith('[')) {
+            config.id = key.replace(/[[\]]/g, '');
+
+            return;
+        }
+
+        config[key] = value;
```
qzhou1607: The type of the `bwIn`, `bwOut` and `latency` values are `string` instead of `number`. Maybe `parseInt` should be used to wrap around the `value` for those keys?

src/lib/rules/performance-budget/connections.ini
molant: I thought about it, but because this file comes from WebPageTest I'd like to avoid having to reformat it when there's an update. This way we can potentially update it dynamically the same way we do with the no vulnerable libraries rule.

src/lib/rules/performance-budget/connections.ts
molant: Good catch!

molant:
This should be ready for another review

PR #818, opened 2018-02-16
New: Add `connector-edge`

molant:
This add the code from https://github.com/sonarwhal/connector-edge into the main repo.

To test this you need to manually do a `npm install`, otherwise `yarn` will complain on Windows (and this connector only runs on Windows).

### DIFF
--- a/packages/connector-edge/appveyor.yml
+++ b/packages/connector-edge/appveyor.yml
@@ -1,0 +1,19 @@
+environment:
+  matrix:
+   - nodejs_version: "LTS"
+   - nodejs_version: 6
+   # will add 9/"Stable" once edge-diagnostics-adapter has a build for it
+
+# Install scripts (runs after repo cloning).
+install:
+  - ps: Install-Product node $env:nodejs_version x64
+  - node --version
+  - npm --version
+  - npm install
+
+# Post-install test scripts.
+test_script:
+  - npm run build
+
+# Don't actually build.
+build: off
--- a/packages/connector-edge/package.json
+++ b/packages/connector-edge/package.json
@@ -1,0 +1,57 @@
+{
+  "dependencies": {
+    "edge-diagnostics-adapter": "^0.6.0",
+    "node-windows": "^0.1.14"
+  },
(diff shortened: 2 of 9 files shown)


packages/connector-edge/appveyor.yml
```diff
@@ -0,0 +1,19 @@
+environment:
```
alrra: Does this file have any effect?

alrra:
This fixes: https://github.com/sonarwhal/connector-edge/issues/22.

packages/connector-edge/appveyor.yml
molant: I was using `appveyor` to verify it at least compiles. The problem we are going to have is that we have a dependency on Windows to download the `edge-diagnostics-adapter` so I'm not sure how that's going to play out with travis 😞

alrra: I will remove this file for the time being.

packages/connector-edge/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "clean": "rimraf dist",
+    "lint": "npm-run-all lint:*",
+    "lint:md": "markdownlint README.md",
+    "lint:js": "eslint . --cache --ext js --ext md --ext ts --ignore-path ../../.eslintignore --report-unused-disable-directives",
+    "preinstall": "node scripts/preinstall.js",
+    "test": "echo",
+    "watch": "npm run build && npm-run-all --parallel -c watch:*",
+    "watch:ts": "npm run build:ts -- --watch"
+  },
+  "version": "1.0.0"
```
alrra: This should be  [`0.1.5`](https://github.com/sonarwhal/connector-edge/blob/2b4732637cda4d2b834de84ab197f1c0f0c82db2/package.json#L54) for now.

And will add `"private": true`.

alrra:
Merged with some modifications in: https://github.com/sonarwhal/sonarwhal/commit/e57c089d348496cd15df375f709479e6e51f3701.

Also, for the time being I also made yarn ignore this workspace.

PR #819, opened 2018-02-16
Fix 748

molant:
Creating the PR just to track things we have to do before merging that aren't elsewhere:

* [x] add `pluralize` dependency to all packages that use it (basically rules, formatters, etc.)
* [x] find a solution to `amphtml-validator` on windows with yarn #820
* [ ] find a solution to issues with `connector-edge` on *nix
* [ ] Group all `manifest` rules under one single package? 
* [ ] Update `readme.md` and contributor guides with the new architecture and yarn (and issues with `amphtml-validator` and how to by-pass them)
* [ ] Declare events as types and not interfaces?


molant:
* [ ] Group all `manifest` rules under one single package?

molant:
* [ ] Update `readme.md` and contributor guides with the new architecture and yarn

PR #820, opened 2018-02-16
Fix: Issue with `amphtml-validator` and Windows

molant:
This partially fixes the problem with `amphtml-validator` on windows machines.
If you have issues installing you need to run `scripts\set-NPM-CONFIG_PREFIX.bat`.
In appveyor we are running that script before doing anything so it should be fine.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,7 @@ environment:
 
 # Install scripts (runs after repo cloning).
 install:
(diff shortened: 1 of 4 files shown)


sarvaje:
you should enable the `amp` rule in the package.json, isn't it?

molant:
> you should enable the amp rule in the package.json, isn't it?

Done

molant:
BTW, I'm waiting to see if appveyor does what it is suppose to do.

PR #855, opened 2018-03-06
Fix: Install rules before validate

qzhou1607:
## Pull request checklist

Make sure you:

In the current code, if a rule has not been installed, when user scans a site, an error is thrown and the user will be asked to re-generate a valid `.sonarwhalrc` even though there already is one. The correct behavior should be prompt to ask the user if they want to install the missing rules first. So this PR moves the validation of rule configurations after the rules have been installed.

### DIFF
--- a/packages/sonarwhal/src/lib/config.ts
+++ b/packages/sonarwhal/src/lib/config.ts
@@ -190,16 +190,18 @@ const loadIgnoredUrls = (userConfig: UserConfig): Map<string, RegExp[]> => {
 /** Validates that the given configuration for a rule is valid */
 const validateRules = (rulesConfig: RulesConfigObject, userConfig: UserConfig) => {
     const rules = Object.keys(rulesConfig);
-
-    rules.forEach((rule) => {
+    const invalidRuleConfigs = rules.reduce((invalidRules, rule) => {
         const Rule = resourceLoader.loadRule(rule, userConfig.extends);
-
         const valid: boolean = validateRule(Rule.meta, userConfig.rules[rule], rule);
 
         if (!valid) {
-            throw new Error(`Rule ${rule} has an invalid configuration`);
+            invalidRules.push(rule);
         }
-    });
+
+        return invalidRules;
+    }, []);
+
+    return invalidRuleConfigs;
 };
 
 export class SonarwhalConfig {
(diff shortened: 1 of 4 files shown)


packages/sonarwhal/src/lib/config.ts
```diff
@@ -190,16 +190,19 @@ const loadIgnoredUrls = (userConfig: UserConfig): Map<string, RegExp[]> => {
 /** Validates that the given configuration for a rule is valid */
 const validateRules = (rulesConfig: RulesConfigObject, userConfig: UserConfig) => {
     const rules = Object.keys(rulesConfig);
+    const invalidRuleConfigs = [];
```
molant: Can we have a `reduce` here?

qzhou1607: @molant Done.

packages/sonarwhal/src/lib/config.ts
```diff
@@ -190,16 +190,18 @@ const loadIgnoredUrls = (userConfig: UserConfig): Map<string, RegExp[]> => {
...
         if (!valid) {
-            throw new Error(`Rule ${rule} has an invalid configuration`);
+            invalidRules.push(rule);
         }
-    });
+
+        return invalidRules;
+    }, []);
+
+    return invalidRuleConfigs;
```
molant: I think I would change the return of this to be more aligned with what we do in other places and also so the name of the function is still valid. Maybe return something like:
```js
{
  valid: [],
  invalid: []
}
```

packages/sonarwhal/src/lib/config.ts
```diff
@@ -265,11 +267,19 @@ export class SonarwhalConfig {
...
-        validateRules(rules, userConfig);
-
         return new SonarwhalConfig(userConfig, browsers, ignoredUrls, normalizeRules(rules));
     }
 
+    /**
+     * Return a list of rules that have invalid configurations.
+     * @param config
+     */
+    public static validateRuleConfig(config: UserConfig): Array<string> {
```
molant: Shouldn't this be `validateRulesConfig`?

molant: Also, do we need to normalize here? I thought this was done earlier in the process.

PR #856, opened 2018-03-07
Fix: Validate resource name on load

qzhou1607:
## Pull request checklist

Make sure you:

* Change the path to load external rules back to `process.cwd()`.
* Add additional check to verify the name of the resource loaded.

### DIFF
--- a/packages/sonarwhal/src/lib/utils/misc.ts
+++ b/packages/sonarwhal/src/lib/utils/misc.ts
@@ -145,6 +145,11 @@ const normalizeStringByDelimiter = (value: string, delimiter: string) => {
     return normalizeString(value).replace(/[^a-z0-9]/gi, delimiter);
 };
 
+/** Return if normalized `source` string includes normalized `included` string. */
+const isNormalizedIncluded = (source: string, included: string) => {
+    return normalizeString(source).includes(normalizeString(included));
+};
+
 /** Convert '-' delimitered string to camel case name. */
 const toCamelCase = (value: string) => {
     return value.split('-').reduce((accu: string, w: string) => {
@@ -321,6 +326,7 @@ const getSonarwhalPackage = () => {
 export {
     cutString,
     delay,
+    isNormalizedIncluded,
     findNodeModulesRoot,
     findPackageRoot,
     getFileExtension,
--- a/packages/sonarwhal/src/lib/utils/resource-loader.ts
+++ b/packages/sonarwhal/src/lib/utils/resource-loader.ts
@@ -16,7 +16,7 @@ import * as path from 'path';
 import * as globby from 'globby';
 import * as semver from 'semver';
 
-import { getPackage, getSonarwhalPackage, findNodeModulesRoot, findPackageRoot, readFile } from '../utils/misc';
+import { getPackage, getSonarwhalPackage, findNodeModulesRoot, findPackageRoot, isNormalizedIncluded, readFile } from '../utils/misc';
 import { debug as d } from '../utils/debug';
 import { Resource, IRuleConstructor, SonarwhalResources } from '../types';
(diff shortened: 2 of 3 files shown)


packages/sonarwhal/src/lib/utils/resource-loader.ts
```diff
@@ -232,7 +232,7 @@ export const loadResource = (name: string, type: ResourceType, configurations: A
         throw new Error(`Resource ${name} isn't compatible with current sonarwhal version`);
     }
 
-    if (!resource) {
+    if (!resource || !resource.name || !equalNormalizedString(resource.name, `${name}${type}`)) {
```
sarvaje: I think `resource.name` is not going to exists. To read the name you need to read the `package.json` of the package (you can use something like `require(path.join(process.cwd(), 'package.json'))`

qzhou1607: @sarvaje Hmm...I don't see `name` in package.json either, should that parameter be added to package.json first? Also I tried loading connectors/rules/parsers/external rules and `resource.name` is always the same as the exported class name in each package.

sarvaje: take a look to any package in the folder `packages` and take a look to their `package.json` you will see the `name` property.

molant:
@qzhou1607 can you fix the conflicts please?

packages/sonarwhal/src/lib/utils/resource-loader.ts
molant: > you can use something like `require(path.join(process.cwd(), 'package.json'))`

You should be using [`getPackage()` from `misc`](https://github.com/sonarwhal/sonarwhal/blob/cec7d5bed384cddae5bf23a2d54005b00562c2b7/packages/sonarwhal/src/lib/utils/misc.ts#L312), that way it will be easier to mock if needed.

qzhou1607: @sarvaje For the package that contains multiple rules (such as `jsll`), the `name` property in package.json is `@sonarwhal/packagename` (@sonarwhal/jsll), it doesn't contain the name of the rule to be loaded. I can verify the package name, but to do that it will require that we always use `packageName/rule-ruleName` in `.sonarwhal` when there are multiple rules in a package. Will this be okay?

molant: > but to do that it will require that we always use `packageName/rule-ruleName` in `.sonarwhal` when there are multiple rules in a package

This is what we should be doing already

qzhou1607:
@molant Done.

packages/sonarwhal/src/lib/utils/misc.ts
```diff
@@ -145,6 +145,10 @@ const normalizeStringByDelimiter = (value: string, delimiter: string) => {
     return normalizeString(value).replace(/[^a-z0-9]/gi, delimiter);
 };
 
+const isParentString = (str1: string, str2: string) => {
```
molant: Maybe something like `isNormalizedIncluded` ? Also I'd change `str1` for `source`, and `str2` for `included` (or maybe something different?).

And then add a description like:

> Checks if `source` contains `included` after normalizing both

packages/sonarwhal/src/lib/utils/resource-loader.ts
```diff
@@ -261,6 +262,18 @@ export const loadResource = (name: string, type: ResourceType, configurations: A
             resource = res;
         }
 
+        if (resource && (source === currentProcessDir)) {
```
sarvaje: we are already checking if resource exists or not.

Move this inside the if above.

PR #862, opened 2018-03-08
Fix: Handle when `parsers` are not defined

qzhou1607:
## Pull request checklist

Make sure you:

It doesn't seem to be right that  the scanner fails if `Parsers` are not defined in `.sonarwhalrc`, the `Parsers` are empty most of the time so it's probably not ideal to force the user to put `"Parsers":[]` there. Or if we want to reinforce that, we need to make the error message more friendly.


packages/sonarwhal/src/lib/utils/resource-loader.ts
```diff
@@ -295,11 +295,13 @@ export const loadResource = (name: string, type: ResourceType, configurations: A
 const loadListOfResources = (list: Array<string> | Object = [], type: ResourceType, configurations: Array<string> = []): { incompatible: Array<string>, missing: Array<string>, resources: Array<any> } => {
     const missing: Array<string> = [];
     const incompatible: Array<string> = [];
+    // In case `Parsers` are not defined by the user.
+    const safeList: Array<string> | Object = list || [];
```
sarvaje: Is this necessary? list has defined a default value of `[]`. (line #295)

qzhou1607:
Closing as it was already fixed.

PR #880, opened 2018-03-13
Fix: Call `format`s asynchronously

molant:
## Pull request checklist

Make sure you:

There were some reports that Excel formatter was generating just an empty file. It turns out it's the first `formatter` that actually requires to be called asynchronously.

This PR makes the call to `formatter.format` using `await`.

### DIFF
--- a/packages/sonarwhal/src/lib/cli/analyze.ts
+++ b/packages/sonarwhal/src/lib/cli/analyze.ts
@@ -1,3 +1,6 @@
+import { promisify } from 'util';
+
+import * as async from 'async';
 import * as inquirer from 'inquirer';
 import * as ora from 'ora';
 import * as pluralize from 'pluralize';
@@ -13,6 +16,7 @@ import * as resourceLoader from '../utils/resource-loader';
 import { installPackages } from '../utils/npm';
 import { initSonarwhalrc } from './init';
 
+const each = promisify(async.each);
 const debug: debug.IDebugger = d(__filename);
 
 /*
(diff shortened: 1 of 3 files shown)


packages/sonarwhal/src/lib/cli/analyze.ts
```diff
@@ -204,15 +208,15 @@ export const analyze = async (actions: CLIOptions): Promise<boolean> => {
...
+    const print = async (reports: Array<Problem>, target: string): Promise<void> => {
         if (hasError(reports)) {
             endSpinner('fail');
         } else {
             endSpinner('succeed');
         }
 
-        sonarwhal.formatters.forEach((formatter) => {
-            formatter.format(reports, target);
+        await each(sonarwhal.formatters, async (formatter) => {
```
sarvaje: why do you use the package `async` instead of use a `for … of..`?

molant: We should start using `async` in all places instead of `for … of` for legibility purposes and this was a good moment to start.

sarvaje: `for … of` is legible :P

molant: Let's replace all the `map`, `reduce`, `forEach`, etc. for regular `for` loops then 😛

sarvaje: It is not the same example, you don't need to add any extra package to do that.

molant: We are using `lodash` in parts of the code 🙄

PR #881, opened 2018-03-13
New: Define media type for config files

qzhou1607:
## Pull request checklist

Make sure you:

Change per https://github.com/sonarwhal/sonarwhal/pull/808#issuecomment-372785511

### DIFF
--- a/packages/sonarwhal/src/lib/utils/content-type.ts
+++ b/packages/sonarwhal/src/lib/utils/content-type.ts
@@ -250,7 +250,13 @@ const determineMediaTypeBasedOnFileExtension = (resource: string): string => {
     return getMediaTypeBasedOnFileExtension(fileExtension);
 };
 
-/** Set file type as `text` if the file name matches the pattern of `.somethingrc` */
+/**
+ * Determine the media type based on the file name, extension and content.
+ * This is only for edge cases. So far it detects:
+ *
+ * * `.configrc` files: If the content is a valid `json`, it will return `text/json`, `text/plain` otherwise
+ *
+ */
 const determineMediaTypeBasedOnFileName = (resource: string, rawContent: Buffer): string => {
     const fileName = getFileName(resource);
 
--- a/packages/sonarwhal/src/lib/utils/content-type.ts
+++ b/packages/sonarwhal/src/lib/utils/content-type.ts
@@ -5,7 +5,7 @@ import { parse, MediaType } from 'content-type';
 
 import { debug as d } from './debug';
 import { IAsyncHTMLElement } from '../types';
-import { getFileExtension, normalizeString } from './misc';
+import { getFileExtension, getFileName, normalizeString } from './misc';
 
 const debug = d(__filename);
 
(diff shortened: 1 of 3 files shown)


packages/sonarwhal/src/lib/utils/content-type.ts
```diff
@@ -250,6 +250,18 @@ const determineMediaTypeBasedOnFileExtension = (resource: string): string => {
...
+const determineMediaTypeBasedOnFileName = (resource: string) => {
+    const fileName = getFileName(resource);
+
+    if (!fileName) {
+        return null;
+    }
+
+    const configFileNameRegex = /^\.[a-z]+rc$/i;
+
+    return configFileNameRegex.test(fileName) ? 'text/json' : null;
```
alrra: IMHO, most .<something>rc file are not JSON. We should do a `JSON.parse(content)` and try catch, and if it the content can be parsed as JSON, then it's JSON.

packages/sonarwhal/src/lib/utils/content-type.ts
```diff
@@ -250,6 +250,18 @@ const determineMediaTypeBasedOnFileExtension = (resource: string): string => {
...
 };
 
+const determineMediaTypeBasedOnFileName = (resource: string) => {
+    const fileName = getFileName(resource);
+
+    if (!fileName) {
+        return null;
+    }
+
+    const configFileNameRegex = /^\.[a-z]+rc$/i;
```
sarvaje: I think we should allow numbers in the name.

packages/sonarwhal/src/lib/utils/content-type.ts
molant: And otherwise end text/plain or something like that

packages/sonarwhal/src/lib/utils/misc.ts
```diff
@@ -40,11 +40,16 @@ const delay = (millisecs: number): Promise<object> => {
     });
 };
 
+/** Try to determin the resource's file name. */
```
molant: > determine

packages/sonarwhal/src/lib/utils/misc.ts
```diff
@@ -40,11 +40,16 @@ const delay = (millisecs: number): Promise<object> => {
...
+const getFileName = (resource: string) => {
+    return path.basename(url.parse(resource).pathname);
+};
+
 /*
  * Try to determine the resource's file extension.
  */
 const getFileExtension = (resource: string): string => {
-    return path.extname(url.parse(resource).pathname).split('.')
+    return getFileName(resource).split('.')
```
molant: Why not use [`path.extname`](https://nodejs.org/dist/latest-v8.x/docs/api/path.html#path_path_extname_path)

packages/sonarwhal/src/lib/utils/misc.ts
```diff
@@ -40,11 +40,16 @@ const delay = (millisecs: number): Promise<object> => {
     });
 };
 
+/** Try to determin the resource's file name. */
+const getFileName = (resource: string) => {
+    return path.basename(url.parse(resource).pathname);
```
molant: Why not use [`path.basename`](https://nodejs.org/dist/latest-v8.x/docs/api/path.html#path_path_basename_path_ext)

packages/sonarwhal/src/lib/utils/content-type.ts
```diff
@@ -250,6 +250,29 @@ const determineMediaTypeBasedOnFileExtension = (resource: string): string => {
     return getMediaTypeBasedOnFileExtension(fileExtension);
 };
 
+const determineMediaTypeBasedOnFileName = (resource: string, rawContent: Buffer) => {
```
molant: Need a description for intellisense here.

molant: Also the return value is missing

qzhou1607:
@molant Done.

qzhou1607:
The `html checker` rule tests are failing in this PR and it could be reproduced in local. It doesn't look like it's caused by the changes in this PR. I will take a look into it if nobody else is already doing it.

molant:
@qzhou1607 is the rule failing with the code in `master`? If so, I'll merge this and in the meantime you can do a separate PR to fix that.

Thanks!

PR #893, opened 2018-03-15
Fix: Install packages as devDependencies

### LINKED ISSUE #868: Make `sonarwhal` install other packages the same way it was installed
If `sonarwhal` is installed as a `devDependency` the packages it installs are not installed as `devDependencies`, but rather as dependencies.

---

Steps to reproduce:

* `npm i --save-dev sonarwhal@next`
*  In the `package.json` have:
   ```json
   ...
   "scripts": {
     "sonarwhal": "sonarwhal"
   },     
   ...
   ```
* `npm run sonarwhal -- --init`
* Choose:
   * Do you want to use a predefined configuration or create your own based on your installed packages? => `predefined`
   * Choose the configuration you want to extend from => `web-recommended`

* `npm run sonarwhal -- example.com` + say `Yes` to install question
* I get:

  ```json
   ...
  "devDependencies": {
    "sonarwhal": ...
  },
  "dependencies": {
    "@sonarwhal/configuration-web-recommended": ...
    "@sonarwhal/formatter-summary": ...
    "@sonarwhal/rule-apple-touch-icons": ...
    "@sonarwhal/rule-axe": ...
    "@sonarwhal/rule-content-type": ...
    "@sonarwhal/rule-disown-opener": ...
    "@sonarwhal/rule-highest-available-document-mode": ...
    "@sonarwhal/rule-html-checker": ...
    "@sonarwhal/rule-http-cache": ...
    "@sonarwhal/rule-http-compression": ...
    "@sonarwhal/rule-manifest-app-name": ...
    "@sonarwhal/rule-manifest-exists": ...
    "@sonarwhal/rule-manifest-file-extension": ...
    "@sonarwhal/rule-manifest-is-valid": ...
    "@sonarwhal/rule-meta-charset-utf-8": ...
    "@sonarwhal/rule-meta-viewport": ...
    "@sonarwhal/rule-no-disallowed-headers": ...
    "@sonarwhal/rule-no-friendly-error-pages": ...
    "@sonarwhal/rule-no-html-only-headers": ...
    "@sonarwhal/rule-no-http-redirects": ...
    "@sonarwhal/rule-no-protocol-relative-urls": ...
    "@sonarwhal/rule-no-vulnerable-javascript-libraries": ...
    "@sonarwhal/rule-ssllabs": ...
    "@sonarwhal/rule-strict-transport-security": ...
    "@sonarwhal/rule-validate-set-cookie-header": ...
    "@sonarwhal/rule-x-content-type-options": ...
  }                                           
  ...
  ```

qzhou1607:
## Pull request checklist

Make sure you:

Fix #868

### DIFF
--- a/packages/sonarwhal/src/lib/utils/npm.ts
+++ b/packages/sonarwhal/src/lib/utils/npm.ts
@@ -1,5 +1,3 @@
-import * as path from 'path';
-import * as fs from 'fs';
 import * as esearch from 'npm/lib/search/esearch';
 import * as npm from 'npm';
 import { promisify } from 'util';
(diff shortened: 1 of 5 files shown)


packages/sonarwhal/src/lib/utils/npm.ts
```diff
@@ -26,7 +26,7 @@ export const installPackages = (packages: Array<string>): boolean => {
         return true;
     }
 
-    const command: string = `npm install ${packages.join(' ')}${global ? ' -g' : ''}`;
+    const command: string = `npm install ${packages.join(' ')}${global ? ' -g' : ' --save-dev'}`;
```
molant: Things are a bit more complicated. Maybe someone installed `sonarwhal` as a regular dependency and thus the package should be installed as a regular dependency. You should check the `package.json` and find where sonarwhal has been installed (`dependencies` or `devDependencies`).

qzhou1607: @molant Done.

packages/sonarwhal/src/lib/utils/npm.ts
```diff
@@ -12,21 +12,43 @@ import * as logger from './logging';
 const debug: debug.IDebugger = d(__filename);
 const npmLoadAsync = promisify(npm.load);
 
-
-const packageExists = () => {
+const getPackageContent = (): string => {
```
molant: Shouldn't this go into `misc` instead of `npm`? We have a few methods about `package.json` in there already (for finding, etc.). Maybe you can use `misc.findPackageRoot()` and check if the returned path is the same as `process.cwd()`?

packages/sonarwhal/src/lib/utils/npm.ts
```diff
@@ -12,21 +12,43 @@ import * as logger from './logging';
...
+
+    try {
+        packageContent = fs.readFileSync(packagePath, 'utf8'); // eslint-disable-line no-sync
+    } catch (err) {
+        // package.json` doesn't exist.
+        return null;
+    }
 
-    return fs.existsSync(packagePath); // eslint-disable-line no-sync
+    return packageContent;
```
molant: Maybe you can do a parse here already. Maybe use `misc.loadJSONFile`, or directly `request('package.json')`

packages/sonarwhal/src/lib/utils/npm.ts
```diff
@@ -12,21 +12,43 @@ import * as logger from './logging';
...
 };
 
 export const installPackages = (packages: Array<string>): boolean => {
-    const global: boolean = !packageExists();
+    /** Content of package.json in the current directory */
+    const packageContent: string = getPackageContent();
```
molant: Based on the feedback above I would do something like:

```ts

const packagePath = misc.findPackageRoot(process.cwd());

if(!packagePath || packagePath !== process.cwd()) {
  // Installation is global
}

const content = misc.loadJSONFile(packagePath); // might need to add `package.json`
...
```

qzhou1607: @molant One question: What if the `packagePath` exists but the parsing of JSON fails? Should we install the package globally or locally?

molant: I think that at that moment we should stop the execution. Something weird is going on and we can't recover from it. Maybe print a message saying that and exit?

packages/sonarwhal/src/lib/utils/npm.ts
```diff
@@ -8,27 +6,48 @@ import { spawnSync, SpawnSyncReturns } from 'child_process';
...
+        // `package.json` is not found.
+        global = true;
+    }
 
     try {
+        if (!global) {
+            const jsonContent = loadJSONFile(packagePath);
+
+            // If `sonarwhal` is a devDependency, then set all packages as devDependencies.
+            isDev = jsonContent && jsonContent.devDependencies.hasOwnProperty('sonarwhal');
```
sarvaje: I think `devDependencies` is not mandatory in the file `package.json`. If so, you need to check if `devDependencies` exists before call to `hasOwnProperty`.

PR #900, opened 2018-03-19
Fix: Better detection for charset metatag position

### LINKED ISSUE #789: False positive meta-charset-utf-8 error
We are seeing a `meta-charset-utf-8` error on our site. It _is_ the first thing in the `<head>`.

Report:
https://sonarwhal.com/scanner/20a37912-3eed-46de-a03a-1bf005c9d2f6

Error:
```
Charset meta tag should be the first thing in '<head>'
https://theundefeated.com/:2:5
<meta charset="UTF-8">
```

Page source:
```
<!DOCTYPE html>
<html lang="en" class="no-js">
<head>
	<meta charset="UTF-8">
	<meta name="viewport" content="width=device-width, initial-scale=1">
	<link rel="profile" href="http://gmpg.org/xfn/11">
	<link rel="pingback" href="https://theundefeated.com/xmlrpc.php">
```

molant:
## Pull request checklist

Make sure you:

This fixes #789 by using the original HTML to know the exact position of the `meta` tag.

### DIFF
--- a/packages/rule-meta-charset-utf-8/src/rule.ts
+++ b/packages/rule-meta-charset-utf-8/src/rule.ts
@@ -9,8 +9,10 @@
  * ------------------------------------------------------------------------------
  */
 
+import * as cheerio from 'cheerio';
+
 import { Category } from 'sonarwhal/dist/src/lib/enums/category';
-import { IAsyncHTMLDocument, IAsyncHTMLElement, IRule, TraverseEnd, RuleMetadata } from 'sonarwhal/dist/src/lib/types';
+import { IAsyncHTMLDocument, IAsyncHTMLElement, IRule, FetchEnd, RuleMetadata, TraverseEnd } from 'sonarwhal/dist/src/lib/types';
 import { isHTMLDocument, normalizeString } from 'sonarwhal/dist/src/lib/utils/misc';
 import { RuleContext } from 'sonarwhal/dist/src/lib/rule-context';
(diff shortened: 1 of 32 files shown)


packages/rule-meta-charset-utf-8/src/rule.ts
```diff
@@ -143,6 +164,7 @@ export default class MetaCharsetUTF8Rule implements IRule {
              */
         };
 
+        context.on('fetch::end::html', setReceivedDom);
```
sarvaje: Why two events? if you already have the html in `fetch::end::html`, why don't validate there?

alrra: @sarvaje See https://github.com/sonarwhal/sonarwhal/pull/900/files#diff-427f2e06ab7b36990c5b6ddb1f0b1f85R56.

molant: I also tried that at the beginning but because the elements are different I'd have to change quite a few things from `context.repo` and I wasn't sure it was worth it.

I could look into doing everything for each `fetch::end::html` (would that make sense? most of the responses are garbage from ads, trackers, etc.) but that will require more time.

sarvaje: I'm not talking about doing it for every `html` event, what I mean is why do this in two events, instead of doing the validation directly in the event `fetch::end::html`.

PR #916, opened 2018-03-22
Chore: Remove `same-origin` dependency.

### LINKED ISSUE #879: Remove package `same-origin` with new method in `misc`
We are using the package `same-origin` somewhere in the code. It's a bit outdated and the new WHATWG URL parser available in node 8 makes things a lot easier to verify if two urls have the same origin.

We need to:

1. Create a new method under `utils/misc`: `sameOrigin` with something similar to the following:

```typescript
export const sameOrigin(url1: string, url2: string): boolean {
  return (new URL(url1)).origin === (new URL(url2)).origin;
}
```

2. Remove `same-origin` from the dependencies and use this method.

[Original conversation](https://github.com/sonarwhal/sonarwhal/pull/875#discussion_r173954925)

sarvaje:
Fix #879

## Pull request checklist

Make sure you:

- [ ] Signed the [Contributor License Agreement](https://cla.js.foundation/sonarwhal/sonarwhal)
- [ ] Followed the [commit message guidelines](https://sonarwhal.com/docs/contributor-guide/getting-started/pull-requests/#commit-messagess)

### DIFF
--- a/packages/sonarwhal/src/lib/utils/misc.ts
+++ b/packages/sonarwhal/src/lib/utils/misc.ts
@@ -344,6 +344,10 @@ const getSonarwhalPackage = () => {
     return require(path.join(__dirname, '../../../../package.json'));
 };
 
+const sameOrigin = (url1: string, url2: string): boolean => {
+    return new URL(url1).origin === new URL(url2).origin;
+};
+
 export {
     cutString,
     delay,
@@ -375,6 +379,7 @@ export {
     readFileAsync,
     requestJSONAsync,
     requestAsync,
+    sameOrigin,
     toCamelCase,
     writeFileAsync
 };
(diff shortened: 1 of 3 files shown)


alrra:
Will update `yarn.lock` before merging.

packages/sonarwhal/src/lib/utils/misc.ts
```diff
@@ -344,6 +344,10 @@ const getSonarwhalPackage = () => {
     return require(path.join(__dirname, '../../../../package.json'));
 };
 
+const sameOrigin = (url1: string, url2: string): boolean => {
```
molant: I actually thought we were using it in more places. Maybe it wasn't worth it to add a new method to `misc`?

alrra: I can modify things.

alrra: https://github.com/sonarwhal/sonarwhal/commit/f8845722ea6b2e06578a6606590d927859abf6a9

PR #917, opened 2018-03-22
TypeScript rules

molant:
## Pull request checklist

Make sure you:

This PR adds a couple new rules for TypeScript. Couple things that I'd like to do first this PR is merged:

* Find a better name for the `target` rule
* Complete the `target` support matrix, `browserslist` returns a lot of browsers in some of the queries and this should be more complete
* Maybe add what browser(s) are affecting the `target`? 

In any case, this is ready to start looking in case there's any major change required.

### DIFF
--- a/packages/rule-typescript-config/README.md
+++ b/packages/rule-typescript-config/README.md
@@ -23,7 +23,9 @@ configuration file:
     "formatters": [...],
     "parsers": [...],
     "rules": {
-        "typescript-config/is-valid": "error"
+        "typescript-config/is-valid": "error",
+        "typescript-config/no-comment": "error",
+        "typescript-config/target": "error"
     },
     ...
 }
@@ -37,10 +39,14 @@ need to be sure that your configuration is the best for your needs.
 ## Rules
 
 * [typescript-config/is-valid][is-valid]
+* [typescript-config/no-comment][no-comment]
+* [typescript-config/target][terget]
 
 ## Further Reading
 
 * [TypeScript Documentation][typescript docs]
 
 [is-valid]: ./docs/is-valid.md
+[no-comment]: ./docs/no-comment.md
+[target]: ./docs/target.md
 [typescript docs]: https://www.typescriptlang.org/docs/home.html
--- a/packages/rule-typescript-config/docs/target.md
+++ b/packages/rule-typescript-config/docs/target.md
@@ -1,0 +1,120 @@
+# `target`
+
+`typescript-config/target` takes into account your `sonarwhal`'s `browserslist`
+configuration and warns you if your `target` (`es3`, `es2015`, etc.) is
+appropriate.
+
+### Why is this important?
+
+Not having the right `ES` target can increase the size of your bundle. At the same
+time, having a `ES` target that not all the browsers you care support can create
+issues with your users.
+
+### What does the rule check?
+
+This rule checks the `compilerOptions` property `target` is the appropriate based
+on your `browserslist` configuration.
+
+The matrix is as follows:
+
+| Target | Chrome | Edge | Firefox | IE | Safari |
+| ------ | ------ | ---- | ------- | -- | ------ |
+|  ES3   |   ✔    |  ✔  |    ✔    | ✔ |    ✔   |
+|  ES5   |   5    |  ✔  |   4     | 9 |    5    |
+| ES2015 |   49   |  13  |   37    | ❌ |   10   |
+| ES2016 |   57   |  14  |   52    | ❌ |  10.1  |
+| ES2017 |   58   |  16  |   53    | ❌ |  10.1  |
+| ESNext |        |      |         |   |        |
+
+Browsers not in this list will be ignored if they are in your browserslist.
+
+Keep in mind that this is an approximative list. I.E.: There are still features
+of `ES2015` that haven't been implemented in all browsers, but the numbers
+above should give you good support for the most common ones.
+
+The following sites were used to create this table:
+
+* http://kangax.github.io/compat-table/es5/
+* http://kangax.github.io/compat-table/es6/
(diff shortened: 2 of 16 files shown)


molant:
This PR also fixes some stuff that could be merged before (first 3 commits)

alrra:
> This PR also fixes some stuff that could be merged before (first 3 commits)

Done.

packages/rule-typescript-config/src/target.ts
```diff
@@ -0,0 +1,235 @@
...
+         * account ranges and minor versions. E.g.: `4.4.3-4.4.4` will use `4.4.3` to compare
+         */
+        const isOlder = (version1: number | string, version2: number | string): boolean => {
+
+            if (typeof version1 === 'number' && typeof version2 === 'number') {
+                return version1 < version2;
+            }
+
+            // Can't store `typeof` or otherwise TypeScript complains about `includes` not being on `number`
+            if (typeof version1 === 'string' && typeof version2 === 'string') {
```
sarvaje: IMO, this is to complex (or I'm missing something). Why don't you just parse the versions?
```
if(typeof v1 === 'string') {
v1 = getMajor(v1)
}
if(typeof v2 === 'string') {
v2 = getMajor(v2)
}
return isOlder(v1, v2)
```

packages/rule-typescript-config/src/target.ts
```diff
@@ -0,0 +1,235 @@
...
+            const versions = Object.keys(compatMatrix);
+            let maxVersion = 'es3';
+
+            /**
+             * This will check all the ES versions and compare all the
+             * browsers from the compat matrix agains the given ones.
+             * If there's one in the compat matrix newer than the ones
+             * provided, it will fail.
+             */
+            versions.forEach((version) => {
```
sarvaje: shouldn't  this be a `reduce`?

packages/rule-typescript-config/src/target.ts
```diff
@@ -0,0 +1,235 @@
...
+        };
+
+        /**
+         * Returns the configured normalized target:
+         *
+         * * The `es` part will be upper cased (e.g.: ES2015, ESNext)
+         * * ES6 --> ES2015
+         */
+        const getConfiguredTarget = (config: TypeScriptConfig): string => {
+            const target = config && config.compilerOptions && config.compilerOptions.target || 'es3';
```
sarvaje: I think you don't need to check if `config` exists, if you receive an `parse::typescript-config` then the config has to exist.

packages/rule-typescript-config/src/target.ts
```diff
@@ -0,0 +1,235 @@
...
+
+            return Targets.get(target.toLowerCase());
```
sarvaje: `Target` is not necessary, you can return `target.toLowerCase()` directly

packages/rule-typescript-config/README.md
```diff
@@ -50,9 +62,171 @@ The configuration is valid:
...
+issues with your users.
+
+### What does the rule check?
+
+This rule checks the `compilerOptions` property `target` is the appropriate based
+on your `browserslist` configuration.
+
+The matrix is as follows:
+
+| Target | Chrome | Edge | Firefox | IE | Safari |
```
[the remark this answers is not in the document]
molant: Nope, see the comment in `Short description of the change(s)`

packages/rule-typescript-config/src/target.ts
molant: target could be `es6` and I need to transform it to `es2015` (thus my issue about normalizing things)

packages/rule-typescript-config/README.md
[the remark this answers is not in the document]
sarvaje: ups, I missed that

packages/rule-typescript-config/src/target.ts
sarvaje: You are right, I missed that value.

packages/rule-typescript-config/src/target.ts
molant: Browser versions are complicated...
We could be comparing:
* `safari 10.1` and `safari 10.0`
* `android 3.0.1` and `android 4.4.3`
* In some cases we get a range as a version: `android 4.4.3-4.4.4`.

I should put some comments or try to refactor the code to make it clearer.

packages/rule-typescript-config/src/target.ts
molant: Yes, I can probably change it to `reduce`

packages/rule-typescript-config/src/target.ts
sarvaje: but at the end, you just compare the major version, isn't it?

molant: I should compare the mayor only if the other tests have failed.

sarvaje: ah, ok

packages/rule-typescript-config/docs/target.md
```diff
@@ -0,0 +1,113 @@
...
+
+This rule checks the `compilerOptions` property `target` is the appropriate based
+on your `browserslist` configuration.
+
+The matrix is as follows:
+
+| Target | Chrome | Edge | Firefox | IE | Safari |
+| ------ | ------ | ---- | ------- | -- | ------ |
+|  ES3   |   ✔    |  ✔  |    ✔    | ✔ |    ✔   |
+|  ES5   |   57   |  12  |   45    | 9 |    8    |
```
molant: I'm sure the values for chrome are not the right ones. Need to recheck the kangax tables and update this accordingly.

molant:
@sarvaje @alrra, I've updated the browser support matrix and added a few more links. Let me know if you have any other feedback or otherwise I'll merge it as soon as I can.

PR #924, opened 2018-03-23
New: Add brotli to `requester`

### LINKED ISSUE #735: Make connectors / `Requester` support Brotli compression
Ref: https://github.com/sonarwhal/sonarwhal/pull/731/files#r159750416

molant:
## Pull request checklist

Make sure you:

Close #735

### DIFF
--- a/packages/sonarwhal/package.json
+++ b/packages/sonarwhal/package.json
@@ -24,6 +24,7 @@
     "globby": "^8.0.1",
     "handlebars": "^4.0.11",
     "iconv-lite": "^0.4.19",
+    "iltorb": "^2.0.6",
     "inquirer": "^5.1.0",
     "is-ci": "^1.1.0",
     "is-my-json-valid": "^2.17.1",
--- a/packages/sonarwhal/src/lib/connectors/utils/requester.ts
+++ b/packages/sonarwhal/src/lib/connectors/utils/requester.ts
@@ -8,22 +8,50 @@
  */
 
 import * as url from 'url';
+import { promisify } from 'util';
+import * as zlib from 'zlib';
 
+import * as async from 'async';
+import * as brotli from 'iltorb';
 import * as request from 'request';
 import * as iconv from 'iconv-lite';
 
 import { debug as d } from '../../utils/debug';
+import { toLowerCaseKeys } from '../../utils/misc';
 import { getContentTypeData } from '../../utils/content-type';
 import { NetworkData } from '../../types'; //eslint-disable-line
 import { RedirectManager } from './redirects';
 
 const debug = d(__filename);
+const decompressBrotli = promisify(brotli.decompress);
+const decompressGzip = promisify(zlib.gunzip);
+const inflateAsync = promisify(zlib.inflate);
+const inflateRawAsync = promisify(zlib.inflateRaw);
+const inflate = (buff: Buffer): Promise<Buffer> => {
+    /*
+     * We detect if the data conforms to RFC 1950 Section 2.2:
+     * * CM (Compression Method, bits 0-3) field should be 8
+     * * FCHECK (bits 0-4) should be a multiple of 31
+     *
+     * http://www.ietf.org/rfc/rfc1950.txt
+     */
+    if ((buff[0] & 0x0f) === 8 && (buff.readUInt16BE(0) % 31 === 0)) {
+        return inflateAsync(buff) as any;
+    }
+
+    return inflateRawAsync(buff) as any;
+};
+
+const identity = (buff: Buffer): Promise<Buffer> => {
+    return Promise.resolve(Buffer.from(buff));
+};
+const asyncSome = promisify(async.someSeries);
 
 const defaults = {
     encoding: null,
     followRedirect: false,
-    gzip: true,
     headers: {
+        'Accept-Encoding': 'gzip, deflate, br',
         'Accept-Language': 'en-US,en;q=0.8,es;q=0.6,fr;q=0.4',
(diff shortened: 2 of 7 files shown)


alrra:
@molant You may have broke something in the `http-compression` rule.

molant:
Yes, I'm testing the fix at the moment.

molant:
OK, this should be ready for review. `requester` manually decompresses now the response and supports `brotli`, `gzip` and `deflate`. Adding a new compression method should be trivial now.

packages/sonarwhal/package.json
```diff
@@ -24,6 +24,7 @@
     "globby": "^8.0.1",
     "handlebars": "^4.0.11",
     "iconv-lite": "^0.4.19",
+    "iltorb": "^2.0.6",
```
alrra: Is this faster then https://www.npmjs.com/package/brotli? Or what is the reasoning behind choosing it?

packages/sonarwhal/src/lib/connectors/utils/requester.ts
```diff
@@ -8,22 +8,50 @@
...
+const decompressGzip = promisify(zlib.gunzip);
+const inflateAsync = promisify(zlib.inflate);
+const inflateRawAsync = promisify(zlib.inflateRaw);
+const inflate = (buff: Buffer): Promise<Buffer> => {
+    /*
+     * We detect if the data conforms to RFC 1950 Section 2.2:
+     * * CM (Compression Method, bits 0-3) field should be 8
+     * * FCHECK (bits 0-4) should be a multiple of 31
+     *
+     * http://www.ietf.org/rfc/rfc1950.txt
```
alrra: httpS

packages/sonarwhal/src/lib/connectors/utils/requester.ts
```diff
@@ -45,14 +73,75 @@ export class Requester {
...
+        };
+
+        const functions = [
+            decompressBrotli,
+            decompressGzip,
+            inflate,
+            identity
+        ];
+
+        // In case we an algorithm we don't understand
```
alrra: // In case of an algorithm not defined by us.

packages/sonarwhal/src/lib/connectors/utils/requester.ts
```diff
@@ -45,14 +73,75 @@ export class Requester {
...
+            priorities[algorithm];
+
+        return functions.slice(priority);
+    }
+
+    /**
+     * Tries to uncompresses a buffer with fallbacks in case `content-encoding`
+     * is not accurate. E.g.:
+     * `Content-Encoding` is `br` but content is actually `gzip`. It will try
+     * first with brotli, then gzip, then return a copy of the original Buffer
```
alrra: > Brotli

packages/sonarwhal/src/lib/connectors/utils/requester.ts
```diff
@@ -111,7 +200,8 @@ export class Requester {
                     }
                 }
 
-                const { charset, mediaType } = getContentTypeData(null, uri, response.headers, response.body.rawContent);
+                const rawBody: Buffer = await this.decompressResponse(response.headers['content-encoding'], rawBodyResponse);
```
alrra: `response.headers['content-encoding']` =>  `getHeaderValueNormalized(response.headers, 'content-encoding')`

packages/sonarwhal/src/lib/connectors/utils/requester.ts
```diff
@@ -45,14 +73,75 @@ export class Requester {
...
+
+        const functions = [
+            decompressBrotli,
+            decompressGzip,
+            inflate,
+            identity
+        ];
+
+        // In case we an algorithm we don't understand
+        const priority = typeof priorities[algorithm] === 'undefined' ?
```
alrra: From https://tools.ietf.org/html/rfc7231#section-3.1.2.2

> The "Content-Encoding" header field indicates what content codings
   have been applied to the representation, beyond those inherent in the
   media type, and thus what decoding mechanisms have to be applied in
   order to obtain data in the media type referenced by the Content-Type
   header field.  Content-Encoding is primarily used to allow a
   representation.

So, the `Content-Encoding` header can specify multiple encodings. 

E.g.: If for some reason the resource was compressed with gzip, and then, the result of that was again compressed with Brotli, you will have `Content-Encoding: gzip, br`.

packages/sonarwhal/package.json
molant: IIRC this is the same one we are using in the compression rule and it's a native implementation.

alrra: > IIRC this is the same one we are using in the compression rule and it's a native implementation.

See: https://github.com/sonarwhal/sonarwhal/blob/master/packages/rule-http-compression/package.json#L10

molant: I've done a bit of research and I think we should switch the compression rule to `iltorb` even though it's a binary one. The reason is that [`brotli` hasn't been updated in about a year](https://github.com/devongovett/brotli.js/commits/master) and new versions of the algorithm have been released. The maintainers of `iltorb` seem to be doing a [good job at having everything up to date](https://github.com/MayhemYDG/iltorb/blob/master/CHANGELOG.md).

alrra: > I've done a bit of research and I think we should switch the compression rule to iltorb even though it's a binary one. The reason is that brotli hasn't been updated in about a year and new versions of the algorithm have been released. The maintainers of iltorb seem to be doing a good job at having everything up to date.

SGTM!

PR #935, opened 2018-03-26
New: Refactor ajv into a new util in sonarwhal

### LINKED ISSUE #929: Refactor ajv into a new util in sonarwhal
The idea is have a new util that given a schema and a json, returns a object with the data parsed and the errors prettified.

Something more or less like this:
```js
const validate = (schema, json) => {
    var x = ajv.validate(schame, json);

    x.errors = pretify(x.errors);

    return x;
}
```

sarvaje:
Fix #929

## Pull request checklist

Make sure you:


sarvaje:
I still need to add tests.

packages/parser-babel-config/package.json
```diff
@@ -6,9 +6,7 @@
     ],
     "timeout": "1m"
   },
-  "dependencies": {
-    "ajv": "^6.2.1"
-  },
+  "dependencies": {},
```
molant: I think this can be completely removed

packages/parser-typescript-config/package.json
```diff
@@ -6,9 +6,7 @@
     ],
     "timeout": "1m"
   },
-  "dependencies": {
-    "ajv": "^6.2.1"
-  },
+  "dependencies": {},
```
molant: I think this can be removed

packages/sonarwhal/src/lib/utils/schema-validator.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    return results.join(' or ');
+};
+
+const errorGenerators: Array<((error: ajv.ErrorObject, errors?: Array<ajv.ErrorObject>) => string)> = [generateAdditionalPropertiesError, generateEnumError, generatePatternError, generateTypeError, generateAnyOfError];
+
+const prettify = (errors: Array<ajv.ErrorObject>) => {
+    const grouped: _.Dictionary<Array<ajv.ErrorObject>> = _.groupBy(errors, 'dataPath');
+
+    const result = _.reduce(grouped, (allMessages, groupErrors: Array<ajv.ErrorObject>) => {
```
molant: Here you are using `_.reduce` and a bit above directly `reduce`. Let's be consistent in this file with that.

packages/sonarwhal/src/lib/utils/schema-validator.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const property = error.dataPath.substr(1);
+
+    return `'${property}' ${error.message.replace(/"/g, '\'')}. Value found '${error.data}'`;
+};
+
+/**
+ * Returns a readable error for 'type' errors.
+ */
+const generateTypeError = (error: ajv.ErrorObject) => {
```
molant: There's a lot of repeated code. Maybe something like the following?

```ts
const generateError = (type, action) => {
  return (error) => {
    if(error.keyword !== type) {
      return null;
    }

    const property = error.dataPath.substr(1);

   return action(error, property);
  };
}

const generatePatternError = generateError(ErrorKeyword.enum, (error, property) => {
  return `'${property}' ${error.message.replace(/"/g, '\'')}. Value found '${error.data}'`; 
});

// ...

const generators = [
  generatePatternError,
// ...
  generateEnumError
];


```

packages/sonarwhal/src/lib/utils/schema-validator.ts
```diff
@@ -0,0 +1,145 @@
...
+        });
+
+        return allMessages;
+    }, []);
+
+    return result;
+};
+
+export const validate = (schema, json): SchemaValidationResult => {
+    // We are clone the incoming data because the validator can modify it.
```
molant: > We clone the incoming data

packages/sonarwhal/src/lib/utils/schema-validator.ts
```diff
@@ -0,0 +1,145 @@
...
+    const property = error.dataPath.substr(1);
+
+    return `'${property}' ${error.message.replace(/"/g, '\'')}.`;
+};
+
+/**
+ * Returns a readable error message.
+ */
+const generate = (error: ajv.ErrorObject, errors?: Array<ajv.ErrorObject>): string => {
+    // eslint-disable-next-line typescript/no-use-before-define, no-use-before-define
```
molant: Why don't you move `errorGenerators` above here instead of disabling `ESLint`?

sarvaje: I can do it, but then I need to disable `Eslint` in another place.

`errorGenerators` use `generateAnyOfError` and this use `generate`, and generate use `errorGenerators`

molant: Then I think `generate` should be bellow `errorGenerators`. Let's group all the `generators` together.

sarvaje:
@molant @alrra Tests added and Feedback addressed.

packages/sonarwhal/src/lib/types/schemavalidationresult.ts
```diff
@@ -0,0 +1,8 @@
+import * as ajv from 'ajv';
```
alrra: For consistency with everything else, can we rename this file to `schema-validation-result.ts`?

packages/rule-babel-config/package.json
```diff
@@ -6,14 +6,10 @@
     ],
     "timeout": "1m"
   },
-  "dependencies": {
-    "lodash.groupby": "^4.6.0",
-    "lodash.map": "^4.6.0"
-  },
+  "dependencies": {},
```
alrra: Remove this.

packages/rule-typescript-config/package.json
```diff
@@ -6,15 +6,10 @@
     ],
     "timeout": "1m"
   },
-  "dependencies": {
-    "lodash.groupby": "^4.6.0",
-    "lodash.map": "^4.6.0",
-    "lodash.without": "^4.4.0"
-  },
+  "dependencies": {},
```
alrra: Remove this.

packages/sonarwhal/src/lib/types/schemavalidationresult.ts
sarvaje: sure thing!

packages/sonarwhal/src/lib/utils/schema-validator.ts
```diff
@@ -0,0 +1,132 @@
...
+    };
+};
+
+/**
+ * Returns a readable error for 'additionalProperty' errors.
+ */
+const generateAdditionalPropertiesError = generateError(ErrorKeyword.additionalProperties, (error: ajv.ErrorObject, property: string): string => {
+    const additionalProperty = (error.params as ajv.AdditionalPropertiesParams).additionalProperty;
+
+    return `'${property}' ${error.message}. Additional property found '${additionalProperty}'.`;
```
alrra: Maybe check if `property` is empty string, and only print the surrounding `'` if it's not?

(Same goes for the others)

sarvaje: I don't know if I get what do you want, right now the result is:
if property has value: `'propertyName' message`
if property is empty: `'' message`.

How do you thing the output should be?

alrra: Yes, right not for an empty property you get something such as: `'' should NOT have additional properties. Additional property found 'z'`. That looks kinda odd, and I was thinking we should drop the `''` part so that the message would be: `'Should NOT have additional properties. Additional property found 'z'`.

sarvaje: Ahh, ok!

sarvaje: I think this is the only case where property can be an empty string.

alrra: OK. We will updated things if needed as we use this more. Thanks!

alrra:
@sarvaje I'll fix the [outdated test](https://travis-ci.org/sonarwhal/sonarwhal/jobs/359110067#L2424) before merging.

sarvaje:
>@sarvaje I'll fix the outdated test before merging.

Thanks!!

Issue #960, opened 2018-04-02
Use `fetch` events in parsers for events that are not related to parsing and are related to getting the content of the file


alrra:
Parsers currently have custom `parse` events for some actions that can be covered by `fetch` events.

For example instead of events like `parse::<something>::error::not-found` or `parse::<something>::error::not-specified`, we can just have  `fetch::missing::<something>`.

In general I think the following events should cover most cases.

* `fetch::end::<resource-type>`
* `fetch::error::<resource-type>`
* `fetch::missing::<resource-type>`
* `fetch::start::<resource-type>`

I think having this (or adding these events in parsers) will not only reduce the need to introduce new custom events, and thus, new types, but also make more rules possible and easier to implement.

Also, in some cases as with the web app manifest, it will make things backward compatible with what [we have right now](https://sonarwhal.com/docs/contributor-guide/getting-started/events/)

---

See also: #930

alrra:
This is now fixed.

PR #961, opened 2018-04-02
New: Add `fetch::missing::<resource-type>` event

alrra:
Ref: #960

### DIFF
--- a/packages/sonarwhal/docs/contributor-guide/getting-started/events.md
+++ b/packages/sonarwhal/docs/contributor-guide/getting-started/events.md
@@ -7,8 +7,9 @@ their signature, and the `interface` they implement.
 
 * [`element::<element-type>`](#elementelement-type)
 * [`fetch::end::<resource-type>`](#fetchendresource-type)
-* [`fetch::error`](#fetcherror)
-* [`fetch::start`](#fetchstart)
+* [`fetch::error`](#fetcherrorresource-type)
+* [`fetch::missing`](#fetchmissingresource-type)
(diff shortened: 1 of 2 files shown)


packages/sonarwhal/docs/contributor-guide/getting-started/events.md
```diff
@@ -77,10 +78,23 @@ type FetchError {
...
+## `fetch::missing::<resource-type>`
+
+Event is emitted **when** a resource is not found.
+
+**Format:**
+
+```ts
+type FetchMissing {
+    /** The URL to download */
+    resource: string;
```
sarvaje: When the resource is missing, I thing that most of the time we don't now the url/path

alrra: I forgot to change this to `/** The URL of the target. */`. 

Also:
  * `resource` this should be optional (for cases that do not have it)? 
  * Maybe this should also include an `element` also optional for cases such as `<link rel="manifest" href="">`? Or should that be another case?

sarvaje: > `resource` this should be optional (for cases that do not have it)?

I think so.

alrra:
Talking a bit with @molant and @sarvaje, and we decided to see in the future how to better address this if we event have to (most of the time there will probably be just one rule that will use this event).

---

Note: I merged the improvements in https://github.com/sonarwhal/sonarwhal/commit/c62ad58e247af019ec1d6ea330b2317e963bdb08.

PR #985, opened 2018-04-06
New: Add rule to check the usage of `theme-color` meta tag

alrra:
Ref: https://html.spec.whatwg.org/multipage/semantics.html#meta-theme-color

### DIFF
--- a/packages/rule-meta-theme-color/README.md
+++ b/packages/rule-meta-theme-color/README.md
@@ -1,0 +1,189 @@
+# Require a 'theme-color' meta tag with a valid value (`@sonarwhal/rule-meta-theme-color`)
+
+`rule-meta-theme-color` rule checks if a single `theme-color` meta tag
+is specified in the `<head>` with a valid value.
+
+## Why is this important?
+
+The [`theme-color` meta tag][spec] provides a way to suggest a color
+that browsers should use to customize the display of the page or of
+the surrounding user interface. For example, browsers might use the
+color for the page's title bar, or use it as a color highlight in a
+tab bar or task switcher.
+
+So, especially in the context of progressive web apps, for a more
+app-like feel, providing a theme color is essential.
+
+Note: Always specify the theme color using the meta tag. Even though
+it can also be declared in the [web app manifest file][manifest],
+browsers only acknowledge it from there once the user has added the
+page to their homescreen.
+
+## How to use this rule?
+
+To use it you will have to install it via `npm`:
+
+```bash
+npm install @sonarwhal/rule-meta-theme-color
+```
+
+Note: You can make `npm` install it as a `devDependency` using the
+`--save-dev` parameter, or to install it globally, you can use the
+`-g` parameter. For other options see [`npm`'s
+documentation](https://docs.npmjs.com/cli/install).
+
+And then activate it via the [`.sonarwhalrc`][sonarwhalrc]
+configuration file:
+
+```json
+{
+    "connector": {...},
+    "formatters": [...],
+    "parsers": [...],
+    "rules": {
+        "meta-theme-color": "error"
+    },
+    ...
+}
+```
+
+## What does the rule check?
+
+The rule checks if a single `theme-color` meta tag is specified in
+the `<head>` and the value of its `content` attribute is a [valid CSS
+color][color] supported by the [targeted browsers](browserslist).
+
+### Examples that **trigger** the rule
+
+The `theme-color` meta tag is not specified:
+
+```html
+<!doctype html>
+<html lang="en">
+    <head>
+        <title>example</title>
+        ...
+    </head>
+    <body>...</body>
+</html>
+```
+
+The `theme-color` meta tag is wrongly specified as `<space>theme-color`:
+
+```html
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="utf8">
+        <meta name=" theme-color" content="#f00">
+        ...
+    </head>
+    <body>...</body>
+</html>
+```
+
+The `theme-color` meta tag is specified with an invalid value:
+
+```html
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="utf8">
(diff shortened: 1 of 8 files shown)


packages/rule-meta-theme-color/README.md
```diff
@@ -0,0 +1,193 @@
...
+        <title>example</title>
+        ...
+    </head>
+    <body>...</body>
+</html>
+```
+
+<!-- markdownlint-disable MD038 -->
+
+The `theme-color` meta tag is wrongly specified as ` theme-color`:
```
molant: Maybe add a note that there's an space at the beginning to make it clear?

packages/rule-meta-theme-color/src/rule.ts
```diff
@@ -0,0 +1,184 @@
...
+        },
+        id: 'meta-theme-color',
+        schema: [],
+        scope: RuleScope.any
+    }
+
+    public constructor(context: RuleContext) {
+
+        /*
+         * This function exists because not all connector (e.g.: jsdom)
```
molant: > connectors

packages/rule-meta-theme-color/src/rule.ts
```diff
@@ -0,0 +1,184 @@
...
+                    !rgbaIsSupported)
+            ) {
+                await context.report(resource, element, `'content' attribute value ('${contentValue}') is not unsupported`);
+            }
+        };
+
+        const validate = async (event: TraverseEnd) => {
+            const { resource }: { resource: string } = event;
+
+            // The following check don't make sense for non-HTML documents.
```
molant: > doesn't

packages/rule-meta-theme-color/src/rule.ts
```diff
@@ -0,0 +1,184 @@
...
+
+            if (!isHTMLDocument(resource, context.pageHeaders)) {
```
molant: Isn't `TraverseEnd` send only for HTML documents? If so this check is not needed.

packages/rule-meta-theme-color/src/rule.ts
```diff
@@ -0,0 +1,184 @@
...
+            // Check if meta tag is specified.
+
+            if (themeColorMetaTags.length === 0) {
+                await context.report(resource, null, `No 'theme-color' meta tag was specified`);
+
+                return;
+            }
+
+            /*
+             * Treat the first charset meta tag as the one
```
molant: Copy paste from `charset meta tag`?

packages/rule-meta-theme-color/src/rule.ts
molant: Any reason why you are not subscribed to `traverse::element::meta`? This way you are sure to only receive the element for HTML, you can remove parts of the code and you only need to keep track if a previous `theme-color` has been emitted already.

You only need to subscribe to `traverse::end` and if none found previously report it.

packages/rule-meta-theme-color/src/rule.ts
```diff
@@ -0,0 +1,184 @@
...
+
+            const bodyMetaTags: Array<IAsyncHTMLElement> = getThemeColorMetaTags(await pageDOM.querySelectorAll('body meta'));
+
+            if ((bodyMetaTags.length > 0) && bodyMetaTags[0].isSame(themeColorMetaTag)) {
+                await context.report(resource, themeColorMetaTag, `Meta tag should not be specified in the '<body>'`);
+
+                return;
+            }
+
+            // All other charset meta tags should not be included.
```
sarvaje: Review this comment

packages/rule-meta-theme-color/README.md
sarvaje: Yes, please, the fist time I though it didn't support the short version of the color hehehe.

packages/rule-meta-theme-color/src/rule.ts
alrra: Yeap.

packages/rule-meta-theme-color/src/rule.ts
alrra: Done.

alrra:
@molant @sarvaje Should be ready for another review.

packages/rule-meta-theme-color/README.md
```diff
@@ -0,0 +1,189 @@
...
+    ...
+}
+```
+
+## What does the rule check?
+
+The rule checks if a single `theme-color` meta tag is specified in the
+`<head>` and the value of its `content` attribute is a [valid CSS
+color][color] supported by the [targeted
+browsers](https://sonarwhal.com/docs/user-guide/further-configuration/browser-context/).
```
molant: Shouldn't this be a foot link?

packages/rule-meta-theme-color/src/rule.ts
```diff
@@ -0,0 +1,188 @@
...
+    public constructor(context: RuleContext) {
+
+        let bodyElementWasReached: boolean = false;
+        let firstThemeColorMetaTag: IAsyncHTMLElement;
+
+        const checkIfThemeColorMetaTagWasSpecified = async (event: TraverseEnd) => {
+            const { resource } = event;
+
+            /*
+             * TODO: Remove this when `traverse::end` is
```
molant: Did you open an issue about it? We should link to it.

alrra: @molant It's basically the same issue as https://github.com/sonarwhal/sonarwhal/issues/982.

packages/rule-meta-theme-color/README.md
alrra: Fixed.

packages/rule-meta-theme-color/src/rule.ts
molant: Can you link to it in the code?

alrra: Done.

PR #986, opened 2018-04-06
Make minor consistency changes

alrra:
* Remove extra unnecessary call from `rule` packages.
* Use more consistent naming convention for files.


molant:
@sarvaje are those changes going to affect the generation of the index page in https://github.com/sonarwhal/sonarwhal.com/pull/406 ?

sarvaje:
> @sarvaje are those changes going to affect the generation of the index page in sonarwhal/sonarwhal.com#406 ?

At least for rules it shouldn't be a problem. We are using just the documentation files.

PR #1010, opened 2018-04-19
New: Rule `https-only`

sarvaje:
## Pull request checklist

Make sure you:


packages/rule-https-only/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "failFast": false,
+    "files": [
+      "dist/tests/**/*.js"
+    ],
+    "timeout": "1m"
+  },
+  "description": "sonarwhal rule that checks if your site is using https and if it has Mixed Content.",
+  "devDependencies": {
+    "@types/debug": "0.0.30",
+    "@types/node": "8.0.14",
```
alrra: Do you need this? We removed it from everywhere else as it creates a lot of problems.

packages/rule-https-only/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "lint:md": "markdownlint --ignore CHANGELOG.md *.md",
+    "sonarwhal": "node node_modules/sonarwhal/dist/src/bin/sonarwhal.js",
+    "test": "npm run lint && npm run build && nyc ava",
+    "init": "npm install && npm run build",
+    "watch": "npm run build && npm-run-all --parallel -c watch:*",
+    "watch:assets": "npm run build:assets -- -w --no-initial",
+    "watch:test": "ava --watch",
+    "watch:ts": "npm run build:ts -- --watch"
+  },
+  "peerDependencies": {
```
alrra: Can you sort things alphabetically? Thanks!

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,89 @@
+/**
+ * @fileoverview sonarwhal rule that checks if your site is using https and if it has Mixed Content.
```
alrra: > HTTPS

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,89 @@
...
+    private targetResource: string = '';
+
+    public static readonly meta: RuleMetadata = {
+        docs: {
+            category: Category.security,
+            description: `sonarwhal rule that checks if your site is using https and if it has Mixed Content.`
+        },
+        id: 'https-only',
+        schema: [
+            /*
```
alrra: Remove this.

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+export default class HttpsOnlyRule implements IRule {
+    private targetHTTPs: boolean = false;
+    private targetResource: string = '';
+
+    public static readonly meta: RuleMetadata = {
+        docs: {
+            category: Category.security,
+            description: `sonarwhal rule that checks if your site is using https and if it has Mixed Content.`
```
alrra: > HTTPS

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        const validateTarget = async (event: Event): Promise<boolean> => {
+            const { resource } = event;
+
+            this.targetResource = resource;
+
+            if (!isHTTPS(resource)) {
+                debug('HTTPS no detected');
+
+                await context.report(resource, null, 'The site should be https');
```
alrra: > Should be served over HTTPS.

Can you use `HTTPS` everywhere else? Thanks!

packages/rule-https-only/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "ava": {
+    "failFast": false,
+    "files": [
+      "dist/tests/**/*.js"
+    ],
+    "timeout": "1m"
+  },
+  "description": "sonarwhal rule that checks if your site is using https and if it has Mixed Content.",
```
alrra: > HTTPS 

> mixed content

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,89 @@
...
+
+                return false;
+            }
+
+            debug('HTTPS detected');
+
+            this.targetHTTPs = true;
```
alrra: Maybe use something more descriptive, like: `targetIsServedOverHTTPS` ?

packages/rule-https-only/tests/tests.ts
```diff
@@ -0,0 +1,55 @@
+import { readFileSync } from 'fs';
+
+import { generateHTMLPage } from 'sonarwhal/dist/tests/helpers/misc';
+import { RuleTest } from 'sonarwhal/dist/tests/helpers/rule-test-type';
+import * as ruleRunner from 'sonarwhal/dist/tests/helpers/rule-runner';
+
+const ruleName = 'https-only';
```
alrra: Use `getRuleName`, don't hardcode the value.

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,89 @@
...
+                return;
+            }
+
+            if (!this.targetHTTPs) {
+                return;
+            }
+
+            const { resource } = event;
+
+            if (resource === this.targetResource) {
```
alrra: Why is this check necessary?

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,89 @@
...
+                return;
+            }
+
+            if (!isHTTPS(resource)) {
+                await context.report(resource, null, 'Requested insecure content');
```
alrra: Why not just used `Should be served over HTTPS`?

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,32 @@
...
+It's good to put some edge cases in here.
+
+## Can the rule be configured?
+
+If this rule allows some configuration, please put the format and
+options for the user.
+
+## Further Reading
+
+What can the user read to know more about this subject?
```
alrra: When you write the docs, add this https://doesmysiteneedhttps.com/ (and also feel free to get some inspiration from it).

packages/rule-https-only/package.json
sarvaje: We have it in the root one, maybe it is not necessary here.

packages/rule-https-only/src/rule.ts
sarvaje: yep, maybe it is not necessary

packages/rule-https-only/src/rule.ts
sarvaje: I like it (I'm not very good for messages heheh)

packages/rule-https-only/README.md
sarvaje: ok!

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,85 @@
+# https-only (`https-only`)
+
+`https-only` check if your site is using HTTPS and warns against having
```
molant: > checks

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,85 @@
...
+mixed content.
+
+## Why is this important?
+
+HTTPS is important to guarantees content integrity. Even when your site
```
molant: > guarantee

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,85 @@
...
+doesn't have sensitive information, an attacker can change the content
+or inject scripts.
```
molant: > an attacker could change the content or inject malicious scripts (like a [crypto miner](https://scotthelme.co.uk/protect-site-from-cryptojacking-csp-sri/) to use your users CPU power).

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,85 @@
...
+    },
+    ...
+}
+```
+
+## What does the rule check?
+
+This rule checks two things:
+
+* The url is served using HTTPS
```
molant: > The main target is served using HTTPS
> If the main target is an HTML file, all its resources should be on HTTPS too

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,85 @@
...
+* If so, all the content served is HTTPS too.
+
+### Examples that **trigger** the rule
+
+If your site is not served as HTTPS.
```
molant: > using HTTPS

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```
+sonarwhal http://example.com
+```
+
+If your site is served as HTTPS, but your assets are HTTP.
```
molant: > using HTTPS, but one or more resources use HTTP.

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,85 @@
...
+```html
+<body>
+    <img src="http://assets.com/image.png" />
+    <script src="http://assets.com/script.js"></script>
+</body>
+```
+
+### Examples that **pass** the rule
+
+Your site is served as HTTPS and your assets too.
```
molant: > using HTTPS and its resources too.

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,85 @@
...
+```html
+<body>
+    <img src="https://assets.com/image.png" />
+    <script src="https://assets.com/script.js"></script>
+</body>
+```
+
+## Can the rule be configured?
+
+This rule cannot be configured.
```
molant: I'd remove this section. Don't think we have it for rules that can't be configured.

packages/rule-https-only/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "ava": {
+    "failFast": false,
+    "files": [
+      "dist/tests/**/*.js"
+    ],
+    "timeout": "1m"
+  },
+  "description": "sonarwhal rule that checks if your site is using HTTPS and if it has mixed content.",
+  "devDependencies": {
+    "@types/debug": "0.0.30",
```
molant: Do we need the `debug` types? Aren't they pulled from the main?

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            const { resource } = event;
+
+            if (!isHTTPS(resource)) {
+                await context.report(resource, null, 'Should be served over HTTPS');
+            }
+        };
+
+        context.on('fetch::start::target', validateTarget);
```
molant: I think we should subscribe to the `fetch::end::target` because there could be a redirect from HTTP to HTTPS and I don't think we should penalize that. If there are redirects for resources we should fail those.

molant: Also I would subscribe to all `fetch::end`. There could be a weird redirect case like HTTPS -->HTTP that is not detected here so we should check that all the `hops` are on HTTPS.

packages/rule-https-only/tests/tests-https.ts
```diff
@@ -0,0 +1,48 @@
...
+const generateResponse = (content: Buffer, type: string): Object => {
+    return {
+        content,
+        headers: { 'Content-Type': type }
+    };
+};
+
+const testsHTTPS: Array<RuleTest> = [
+    {
+        name: 'This test should pass',
```
molant: > HTML page with no resources over HTTPS passes

packages/sonarwhal/src/lib/connectors/debugging-protocol-common/debugging-protocol-connector.ts
```diff
@@ -673,9 +673,11 @@ export class Connector implements IConnector {
 
         try {
             debug(`resource ${href} to be fetched`);
-            await this._server.emitAsync('fetch::start', { resource: href });
+            const fullFaviconUrl = this._finalHref + href.substr(1);
```
molant: The favicon part doesn't seem to be part of this PR?

sarvaje: It is part, this PR has some changes in sonarwhal too (in a different commit)

packages/rule-https-only/src/rule.ts
sarvaje: There is no `fetch::end::target` event, but probably it would be nice to have it.

> Also I would subscribe to all fetch::end. There could be a weird redirect case like HTTPS -->HTTP that is not detected here so we should check that all the hops are on HTTPS.

I think that starting a fetch to `http` can be also a problem, I think that an http petition is vulnerable even if it redirect to a `https`. But I'm not sure...

packages/sonarwhal/src/lib/connectors/debugging-protocol-common/debugging-protocol-connector.ts
molant: OK!

packages/rule-https-only/package.json
sarvaje: I will take a look

packages/rule-https-only/src/rule.ts
molant: Yes, but we already have some rules that touch on this like [STS](https://sonarwhal.com/docs/user-guide/rules/rule-strict-transport-security/) so I think we should be ok with that. Otherwise you will need to know reuse a lot of the code in there, verify if the site is in the list, etc.

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,69 @@
...
+        context.on('fetch::start', validate);
```
alrra: Shouldn't the manifest be included?

molant: Also don't we need to subscribe using fetch::start::* ?

packages/rule-https-only/src/rule.ts
sarvaje: Then, I need to add the event `fetch::end::target` and change the events to `fetch::end::*`, right?

packages/rule-https-only/src/rule.ts
sarvaje: we only have `fetch::start`, `fetch::start::target` and `fetch::start::manifest`, `fetch::start::*` doesn't work for `fetch::start`

molant: ok, in any case we should be changing to the `fetch::end` so does this matter?

packages/rule-https-only/src/rule.ts
molant: Yes, that's what I think. And check the `hops` property of `response`

sarvaje: ok

packages/rule-https-only/package.json
sarvaje: The `package.json` in main doesn't include `@types/debug`

packages/rule-https-only/src/rule.ts
molant: Only the target

sarvaje:
@molant @alrra this should be ready to review

packages/rule-https-only/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+`https-only` checks if your site is using HTTPS and warns against having
+mixed content.
+
+## Why is this important?
+
+HTTPS is important to guarantee content integrity. Even when your site
+doesn't have sensitive information, an attacker can change the content
+or inject malicious scripts (like a [crypto miner][crypto miner] to
+use your users CPU power).
```
alrra: We should also add a note that some browser features are not restricted to HTTPS only.

See also: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts/features_restricted_to_secure_contexts

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,92 @@
...
+
+                return;
+            }
+
+            if (anyHopHTTP(response)) {
+                await context.report(resource, null, `Shouldn't be redirected from HTTP`);
+            }
+        };
+
+        context.on('fetch::end::*', validate);
```
alrra: I don't think we are fetching `<audio>`, `<video>`, `<object>`..., so we are not validating a lot of things.

See: https://w3c.github.io/webappsec-mixed-content/

molant: I guess it will depend on the connector and if the case of media if it has `autoplay` enabled?

What I'm thinking on doing:
* Have a list with all the resources we've analyzed
* Subscribe to those elements events and check it there
* If we receive them via a `fetch::end`, use that information instead

molant:
@alrra, @sarvaje I've updated the code and docs with the feedback. This should be ready for review. I've added checks for a bunch of stuff (`srcset`, `source`, `poster`, `data`). Hopefully I haven't missed anything (but adding them should be trivial).

@kshyju if you want to participate of code reviews feel free (and let me know so I can ping you).

Thanks!

packages/rule-https-only/src/rule.ts
```diff
@@ -0,0 +1,179 @@
...
+        /** Returns an array with all the URLs in the given `srcset` attribute or an empty string if none. */
+        const parseSrcSet = (srcset: string): Array<string> => {
+            if (!srcset) {
+                return [];
+            }
+
+            const urls = srcset
+                .split(',')
+                .map((entry) => {
+                    return entry.split(' ')[0].trim();
```
sarvaje: you should `trim` entry, before the `split(' ')` to be sure you don't have any whitespace at the beginning of the string.

sarvaje:
@molant LGTM too, but I can't approve because it is my PR hehehehe

molant:
@alrra @kshyju want to take a look at this PR?

molant:
@alrray I've updated the PR with your feedback. I'll merge this tomorrow unless there are any objections.

PR #1029, opened 2018-04-25
New: Rules via cli

### LINKED ISSUE #1020: Support define rules via CLI (would ignore .sonarwhalrc rules)
Allow

```
sonarwhal https://example.com --rules http-cache manifest-is-valid ssllabs
```

that would skip .sonarwhalrc rules configuration and apply only rules **http-cache**, **manifest-is-valid** and **ssllabs** with its default configuration (`rule: "error"`).

The benefit: running all defined rules in .sonarwhalrc may be very slow and, while working on website fixes for specific role, it would be easier, quicker and "directly to target" to revalidate by single/few rule/s defined in CLI.
Since it might be complicated to defined custom config per rule in CLI (maybe like `sonarwhal https://example.com --rules ssllabs:["error", {"grade": "A+"}]`; also applying .sonarwhalrc rule config from  would be wrong here since actually wanted to ignore it here), I suggest currently apply the rules default configuration.

kshyju:
This change will allow users to explicitly pass the rules via cmd line.

Example : `sonarwhal https://example.com --rules http-cache,html-checker`

Fix #1020

## Pull request checklist

Make sure you:

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ To use a different rule than the one specified in your `.sonarwhalrc` file:
 ```bash
 sonarwhal https://example.com --rules html-checker
 ```
+
 For more in depth information on how to get started, configurations,
 and more, see the online [user guide](https://sonarwhal.com/docs/user-guide/),
 or the [local version](./packages/sonarwhal/docs/user-guide/index.md)
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ To use a different rule than the one specified in your `.sonarwhalrc` file:
 sonarwhal https://example.com --rules html-checker
 ```
 
+Multiple rules can be specified as a comma sepreated string:
+
+```bash
+sonarwhal https://example.com --rules axe,html-checker
+```
+
 For more in depth information on how to get started, configurations,
 and more, see the online [user guide](https://sonarwhal.com/docs/user-guide/),
 or the [local version](./packages/sonarwhal/docs/user-guide/index.md)
--- a/README.md
+++ b/README.md
@@ -55,6 +55,11 @@ To use a different formatter than the one specified in your `.sonarwhalrc` file
 sonarwhal https://example.com --formatters excel
 ```
 
+To use a different rule than the one specified in your `.sonarwhalrc` file:
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -55,6 +55,12 @@ To use a different formatter than the one specified in your `.sonarwhalrc` file
 sonarwhal https://example.com --formatters excel
 ```
 
+To use a different rule than the one specified in your `.sonarwhalrc` file:
+
+```bash
+sonarwhal https://example.com --rules html-checker
+```
+
```
alrra: Maybe we should also add an example for using multiple rules?

packages/sonarwhal/src/lib/config.ts
```diff
@@ -129,6 +142,14 @@ const updateConfigWithCommandLineValues = (config: UserConfig, actions: CLIOptio
         config.formatters = actions.formatters.split(',');
         debug(`Using formatters option provided from command line: ${actions.formatters}`);
     }
+
+    // If rules are provided, use them
+    if (actions && actions.rules) {
+        const rulesarray = actions.rules.split(',');
+
+        config.rules = buildRulesConfigFromRuleNames(rulesarray, 'warning');
```
alrra: Shouldn't we use `error`?

Personally find useful to have the right status code, even for quick things.

README.md
kshyju: Updated the readme with a multi rule example. Perhaps we should keep the read me with simple ones and create a wiki page with complete set of examples/options and use cases

packages/sonarwhal/src/lib/config.ts
kshyju: Most of the time, I use `warning` for rules,  as i did not want the program to exit. If you have a strong opinion it should be `error`, i will change it.

kshyju: Switched to `error`.

README.md
alrra: >  Perhaps we should keep the read me with simple ones 

We could consider moving the information into its own dedicated page.

Issue #576, opened 2017-10-12
Enable strict mode of TypeScript


molant:
Should we enable [`"strict": true`](http://www.typescriptlang.org/docs/handbook/compiler-options.html) in `tsconfig.json`?

I've tried running it and there are quite a few things to fix, but might be better in the long-term for the project.

@alrra, @sarvaje @qzhou1607 ?

alrra:
👍

sarvaje:
👍

molant:
Looks like there's agreement so removing the `question` tag

sarvaje:
@molant @alrra do you want me to start doing this?

molant:
Making a list of all the things that need someone to work on or where there are currently some issues

| package | owner | comments |
|-----------|-------|--------------|
| create-hint | @molant |  Issue with appinsights `Namespace '"http"' has no exported member 'ServerRequest'.` |
| create-hintrc | @molant |  Issue with appinsights `Namespace '"http"' has no exported member 'ServerRequest'.` |
| create-parser | @molant |  Issue with appinsights `Namespace '"http"' has no exported member 'ServerRequest'.` Also I think there's something broken in the logic of the package |
| connector-chrome | | |
| connector-jsdom | | |
| parser-babel-config | | |
| parser-javascript | | |
| parser-manifest | | |
| parser-typescript-config | | |
| parser-webpack-config | | |
| utils-connector-tools | | |
| utils-debugging-protocol-common | | |

@antross, @sarvaje if you can update this table when you grab one it would be awesome.

Thanks!

molant:
All the packages that are not listed above are already done and merged or waiting to be merged.

sarvaje:
>Issue with appinsights Namespace '"http"' has no exported member 'ServerRequest'.

@molant that happens because of the version of `@types/node`. I had the same problem in the online scanner and the solution was fix the version of that package.

molant:
> I've updated to appinsights 1.0.5 and that seems to solve the problem.

I had to pin to @types/node to 10.9.1 to make it work 😔

molant:
Looks like the remaining thing is the code under `scripts` 🎉

PR #1031, opened 2018-04-26
Chore: Make ESLint disallow explicit type declarations

alrra:
Ref: https://github.com/nzakas/eslint-plugin-typescript/blob/d38862eb13eddc964d7263475f1b0be89ca9e960/docs/rules/no-inferrable-types.md

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -260,6 +260,10 @@
         "typescript/member-naming": "error",
         "typescript/no-angle-bracket-type-assertion": "error",
         "typescript/no-array-constructor": "error",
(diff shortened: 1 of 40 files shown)


molant:
If we want to switch to `strict` mode we shouldn't be doing this (#576)

I think the more explicit we are the better TBH, I've had a few problems in this project because I didn't declare the type of the variable even though I was assigning it directly there (and I was receiving a type different than the one I was expecting).

alrra:
> I think the more explicit we are the better TBH

alrra:
How it feels declaring everything all the time:



---

But yes, closing this one in favor of #576.

PR #1055, opened 2018-05-11
Update tests and update event for target url to `fetch::end::{type}`

sarvaje:
## Pull request checklist

Make sure you:

### DIFF
--- a/packages/parser-babel-config/tests/tests.ts
+++ b/packages/parser-babel-config/tests/tests.ts
@@ -19,7 +19,7 @@ test.beforeEach((t) => {
 });
 
 test(`If the resource doesn't match the target file names, nothing should happen`, async (t) => {
-    const sandbox = sinon.sandbox.create();
+    const sandbox = sinon.createSandbox();
 
     new BabelConfigParser(t.context.sonarwhal); // eslint-disable-line no-new
     sandbox.spy(t.context.sonarwhal, 'emitAsync');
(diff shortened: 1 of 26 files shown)


molant:
image

🤣

molant:
The test suite passes with this change but we should verify a few more things to make sure everything works as expected. I've talked with @sarvaje and we should at least do the following:

* Remove the `isHTMLDocument` checks we have in some of the rules (like `meta-charset-utf-8`). If we are doing that check in another event other than `fetch::end::html` then we need to know why. For example, for `meta-charset-utf-8` we subscribe to `traverse::end` which should only be sent after traversing a document (what happens if we pass a text file as target? maybe we need to rely on the content type)
* We should add tests to all the rules that subscribe to `traverse::end::html` and send something as the target that's not an HTML document. E.g.: `sslllabs`, `meta-charset-utf-8`, etc. We shouldn't receive any errors but we need to verify this.
* something else?

@sarvaje, @alrra any other things we should check?

sarvaje:
hahahaha, I don't know why the show the commit message like that hahahahaha

sarvaje:
>We should add tests to all the rules that subscribe to traverse::end::html and send something as the target that's not an HTML document. E.g.: sslllabs, meta-charset-utf-8, etc. We shouldn't receive any errors but we need to verify this.

`fetch::end::html`

sarvaje:
@molant this is ready to review again

alrra:
Closing as this PR is not merged in the [`hint`](https://github.com/sonarwhal/sonarwhal/commits/hint) branch.

PR #1075, opened 2018-05-16
Abort sonarwhal execution if a resource dependency is missing

### LINKED ISSUE #1014: Trying to load a wrong connector throws an error
Currently, specifying the wrong formatter prompts ask the user to install/update a package: 

Config 

image

Result
image

With a connector we get also an exception:

Config

image

Result

image

IMO `connector`s should do the same as `formatter`s and in any case the behavior should be normalized across resources.

sarvaje:
Fix #1014

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/sonarwhal/src/lib/types/resourceerror.ts
+++ b/packages/sonarwhal/src/lib/types/resourceerror.ts
@@ -1,0 +1,13 @@
+import { ResourceErrorStatus } from '../enums/errorstatus';
+
+export class ResourceError extends Error {
+    public status: ResourceErrorStatus;
+
+    public constructor(error: Error, status: ResourceErrorStatus, message?: string) {
+        super(error.message);
+        this.message = message || error.message;
+        this.name = 'ResourceError';
+        this.stack = error.stack;
+        this.status = status;
+    }
+}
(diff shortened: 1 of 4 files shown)


molant:
@sarvaje there are some errors running the tests with `resource-loader`. Can you please take a look?

sarvaje:
sure thing!

packages/sonarwhal/src/lib/types/resourceerror.ts
```diff
@@ -0,0 +1,13 @@
+import { ResourceErrorStatus } from '../enums/errorstatus';
+
+export class ResourceError extends Error {
+    public status: ResourceErrorStatus;
+
+    public constructor(error: Error, status: ResourceErrorStatus, message?: string) {
+        super(error.message);
```
molant: This signature looks a bit weird, especially when you do later a:

```js
throw new ResourceError(new Error(`Resource ${name} isn't compatible with current sonarwhal version`), ResourceErrorStatus.NotCompatible);
```

We shouldn't have to create a new error outside the constructor. It looks like there are a few things going on with TS and inheriting from built-ins in TS:

* https://www.metachris.com/2017/01/custom-errors-in-typescript-2.1/
* https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#support-for-newtarget


But those seem to be from TS 2.1 and 2.2, not sure a bout the current status for us in node. In any case, I think we should have something like:

```ts
public constructor(error: Error | string, status: ResourceErrorStatus) { }
```

or even 

```ts
public constructor(error: string, status: ResourceErrorStatus) { }
```

to simplify things.

sarvaje:
@molant done!

PR #1082, opened 2018-05-18
Revving regex

### LINKED ISSUE #1080: http-cache doesn't allow multiple dots
A js file named `/jquery.plugin.1234.js` will give an error `'No configured patterns for cache busting match'` because the regex doesn't allow dots in the name of the file.

```js
/\/(\w|-|_)+(\.|-|_)\w+\.\w+$/i
```

should be changed to 

```js
/\/.+(\.|-|_)\w+\.\w+$/i
```

UziTech:
## Pull request checklist

Make sure you:

Updated the [file revving regex](https://sonarwhal.com/docs/user-guide/rules/rule-http-cache/#what-does-the-rule-check) from:

```js
/\/(\w|-|_)+(\.|-|_)\w+\.\w+$/i
```

to: 

```js
/\/.+[._-]\w+\.\w+$/i
```

This will allow file names like `jquery.plugin.123.js` to pass the http-cache filename revving check

Fixes #1080

### DIFF
--- a/packages/rule-http-cache/README.md
+++ b/packages/rule-http-cache/README.md
@@ -34,20 +34,36 @@ validate that the page and resources have a good caching strategy:
     one based on query string parameters (see: [problems with
     proxys][revving files])
 
-The built-in regular expression for file revving is:
+The built-in regular expressions for file revving are:
 
 ```regexp
-/\/(\w|-|_)+(\.|-|_)\w+\.\w+$/i
+/\/[^/]+[._-]v?\d+(\.\d+(\.\d+)?)?[^/]*\.\w+$/i
+/\/v?\d+\.\d+\.\d+.*?\//i
+/\/\d+\//
+/\/v\d.*?\//i
+/\/([^/]+[._-])?([0-9a-f]{5,})([._-].*?)?\.\w+$/i
 ```
 
-This will match urls like the following:
+This will match URLs like the following:
 
 ```text
-https://example.com/assets/script.13b55av.js
-https://example.com/assets/script-2.13b55av.js
-https://example.com/assets/script_2-13b55av.js
-https://example.com/assets/sw_script_13b553442.js
+https://example.com/assets/jquery-2.1.1.js
+https://example.com/assets/jquery-2.1.1.min.js
+https://example.com/assets/jquery-3.0.0-beta.js
+https://example.com/assets/favicon.123.ico
+https://example.com/wp-content/uploads/fvm/out/header-cb050ccd-1524626949.min.js
+https://cdn.example.com/jquery.lazy/1.6.5/jquery.lazy.min.js
+https://s3.amazonaws.com/example/wp-content/uploads/2017/07/15084048/jquery.mediaBoxes.dropdown.js
+https://example.com/site/javascript/v5/jquery.cookie.js
+https://static.xx.fbcdn.net/rsrc.php/v3iJhv4/yG/l/en_US/sqNNamBywvN.js
+https://example.com/assets/unicorn-d41d8cd98f.css
+https://example.com/assets/app.e1c7a.bundle.js
+https://example.com/assets/9f61f58dd1cc3bb82182.bundle.js
+https://example.com/assets/9f61f.js
+https://example.com/assets/9f61f.min.js
 ```
+
+[Test your URLs](https://regex101.com/r/6VduJO/36/)
 
 ### Examples that **trigger** the rule
 
--- a/packages/rule-http-cache/src/rule.ts
+++ b/packages/rule-http-cache/src/rule.ts
@@ -90,15 +90,44 @@ export default class HttpCacheRule implements IRule {
         /** The predefined patterns for file revving.*/
         const predefinedRevvingPatterns: Array<RegExp> = [
             /*
-             * E.g.:
-             * - https://example.com/assets/script.12345.js
-             * - https://example.com/assets/script-2.12345.js
-             * - https://example.com/assets/script_2-12345.js
-             * - https://example.com/assets/icon_meca_1473335663.svg
+             * E.g.: version/timestamp/hash
              *
-             * Live example: https://regex101.com/r/6VduJO/2/
+             * Live example: https://regex101.com/r/6VduJO/36/
              */
-            /\/(\w|-|_)+(\.|-|_)\w+\.\w+$/i
+
+            /*
+             * - https://example.com/assets/jquery-2.1.1.js
+             * - https://example.com/assets/jquery-2.1.1.min.js
+             * - https://example.com/assets/jquery-3.0.0-beta.js
+             * - https://example.com/assets/favicon.123.ico
+             * - https://example.com/wp-content/uploads/fvm/out/header-cb050ccd-1524626949.min.js
+             */
+            /\/[^/]+[._-]v?\d+(\.\d+(\.\d+)?)?[^/]*\.\w+$/i,
+
+            /*
+             * - https://cdn.example.com/jquery.lazy/1.6.5/jquery.lazy.min.js
+             */
+            /\/v?\d+\.\d+\.\d+.*?\//i,
+
+            /*
+             * - https://s3.amazonaws.com/example/wp-content/uploads/2017/07/15084048/jquery.mediaBoxes.dropdown.js
+             */
+            /\/\d+\//,
+
+            /*
+             * - https://example.com/site/javascript/v5/jquery.cookie.js
+             * - https://static.xx.fbcdn.net/rsrc.php/v3iJhv4/yG/l/en_US/sqNNamBywvN.js
+             */
+            /\/v\d.*?\//i,
+
+            /*
+             * - https://example.com/assets/unicorn-d41d8cd98f.css
+             * - https://example.com/assets/app.e1c7a.bundle.js
+             * - https://example.com/assets/9f61f58dd1cc3bb82182.bundle.js
+             * - https://example.com/assets/9f61f.js
+             * - https://example.com/assets/9f61f.min.js
+             */
+            /\/([^/]+[._-])?([0-9a-f]{5,})([._-].*?)?\.\w+$/i
         ];
 
         /** The cache revving patterns to use for matching.*/
--- a/packages/rule-http-cache/tests/tests.ts
+++ b/packages/rule-http-cache/tests/tests.ts
@@ -254,6 +254,90 @@ const defaultTests: Array<RuleTest> = [
         }
     },
     {
+        name: 'JS with long max-age, immutable and file revving with multiple dots passes',
+        serverConfig: {
+            '/': generateHTMLPage('<link rel="icon" href="/favicon.123.ico"><script src="/js.script.123.js"></script>'),
+            '/favicon.123.ico': {
+                content: '',
+                headers: { 'Cache-Control': 'max-age=31536000, immutable' }
+            },
+            '/js.script.123.js': {
+                content: 'var a = 10;',
+                headers: { 'Cache-Control': 'max-age=31536000, immutable' }
+            }
+        }
+    },
+    {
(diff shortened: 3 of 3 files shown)


packages/rule-http-cache/src/rule.ts
```diff
@@ -98,7 +98,7 @@ export default class HttpCacheRule implements IRule {
              *
              * Live example: https://regex101.com/r/6VduJO/2/
```
sarvaje: the `Live example` contains the old regex, can you update it too?

UziTech: nice catch

packages/rule-http-cache/README.md
```diff
@@ -65,7 +65,7 @@ validate that the page and resources have a good caching strategy:
 The built-in regular expression for file revving is:
 
 ```regexp
-/\/(\w|-|_)+(\.|-|_)\w+\.\w+$/i
+/\/.+[._-]\w+\.\w+$/i
```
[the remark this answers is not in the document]
UziTech: I would just check for a number after a`.` or `_` or `-`

```js
/\/.+[._-].*\d.*\.\w+/
```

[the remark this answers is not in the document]
molant: That regex seems to work fine. I've added a few more examples in https://regex101.com/r/6VduJO/5/

@UziTech can you please update your PR with the new link and regex?

Thanks!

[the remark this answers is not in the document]
UziTech: is this one supposed to match?

```
https://static.xx.fbcdn.net/rsrc.php/v3iJhv4/yG/l/en_US/sqNNamBywvN.js
```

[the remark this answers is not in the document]
molant: Facebook uses another approach to cache busting which is by path instead of file IIRC so that one is supposed to match.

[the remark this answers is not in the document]
UziTech: in that case just checking that a number exists would be enough

```js

/\d/
```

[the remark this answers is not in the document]
UziTech: a couple of regex that check for what?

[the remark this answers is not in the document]
UziTech: This is going to be a tough problem to get everyone to agree on since there are so many ways someone could version a file name.

Off the top of my head:

 - filename.v1.2.3.js (semver)
 - filename-f1leha5h.js (file hash)
 - filename_1234567890.js (last modified time)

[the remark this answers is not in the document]
UziTech: - /page.php/vp47hv3rs10n/filename.js (path version)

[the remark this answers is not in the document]
sarvaje: >a couple of regex that check for what?

For different patterns.

>This is going to be a tough problem to get everyone to agree on since there are so many ways someone could version a file name.

IIRC the rule can be configured with your own regex, so we don't need to support all of them, just the more commons, and if someone needs an specific regex, they can configure it.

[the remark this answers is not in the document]
UziTech: what are the most common ones?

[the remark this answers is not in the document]
sarvaje: I don't know hahaha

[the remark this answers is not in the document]
molant: A couple thoughts on how to do this:

* Look into the most common tools for cache busting (`gulp-rev`,  webpack's cache busting with `[hash]` and `[chunkhash]`, etc.)
* Get a list of all the resources from Httparchive that have a significant `max-age` value and then figure out the patterns.

I think that to make it easier to maintain we should have regex per categories, kind of what @UziTech is doing in the comments (semver, file hash, modified time, etc.).

What do you think?

UziTech:
I split the revving rules up.

https://regex101.com/r/6VduJO/13/

There are some webpack config articles that suggest just using a short hash as the file name (i.e. `cr6dw.js`) those of course would be indistinguishable from `jquery.js` so I think we can disregard those by default. I, somewhat arbitrarily, used 10 as the minimum hash length.

UziTech:
I know I will have to fix the tests and the docs but I wanted to make sure we had the right idea before continuing

molant:
@UziTech this looks great. One improvement would be to limit to hexa some of the regular expressions. E.g.: `gulp-rev` and webpack's `hash` use that format as an output.

packages/rule-http-cache/src/rule.ts
```diff
@@ -89,16 +89,48 @@ export default class HttpCacheRule implements IRule {
...
+             * - https://example.com/assets/unicorn-d41d8cd98f.css
+             * - https://example.com/assets/app.9f61f58dd1cc3bb82182.bundle.js
+             *
+             * Live example: https://regex101.com/r/6VduJO/21/
+             *
+             * - https://example.com/assets/9f61f58dd1cc3bb82182.bundle.js
+             *
+             * Live example: https://regex101.com/r/6VduJO/22/
+             */
+            /\/[^/]+[._-]([0-9a-f]{10,})(\.bundle)?\.\w+$/i,
```
alrra: Why are we using the limit of 10?

There are a lot of websites that don't use a hash that long.

molant: I'm trying to find stats for that number but can't seem to find anything conclusive. webpack's default is 20 IIRC, gulp-rev is 10. Don't know about other popular tools.

molant:
I've played a bit with Httparchive and I've come out with [this Query](https://bigquery.cloud.google.com/savedquery/339878413135:904b7bf5977f4e58b86ab37cd593c3dd).

The query:
* Looks for assets with a long cache (1 year to be exact)
* Removes URLs with query parameters
* Removes urls from google syndication, doubleclick, etc.

In total it finds 332911 results. Some thoughts about it:

1. People are caching a lot of things without cache busting 😔:    
   * http://runtu.org/media/system/js/mootools.js
1. It looks like versioning is very popular, which makes sense because people use frameworks and libraries, but the strategies vary a bit. Some examples:
   * https://static.ilaoniu.cn/www/js/modernizr-2.6.2.min.js, this should be cover
   * https://unpkg.com/workbox-sw@2.0.1/build/importScripts/workbox-sw.prod.v2.0.1.js, this should be cover
   * http://cdn.jsdelivr.net/jquery.lazy/1.6.5/jquery.lazy.min.js, this CDN is used a lot so we should cover it
   * http://www.meteo-bruxelles.be/site/javascript/v5/jquery.cookie.js, this is similar to the facebook one but doesn't match
1. Looks like wordpress uses a variant of the date version and random things (depending on the version?). There are a bit more than 1200 so I'm not super worried about covering this one:
   * https://s3.amazonaws.com/remixcontestcymatics/wp-content/uploads/2017/07/15084048/jquery.mediaBoxes.dropdown.js
   * https://www.remediocaseronatural.com/wp-content/uploads/fvm/out/header-cb050ccd-1524626949.min.js
1. For the hash, there seems to be variants with lower numbers (5, 7, 8). Don't have a strong opinion about changing the default of 10:
   * https://www.tv.nu/dist/common.e2fd9.js
   * https://az813057.vo.msecnd.net/styles/home.a6114ef.css
   * http://www.arazdownload.com/cache/widgetkit/wk-scripts-22c44c61.js
1. Mix of multiple things?
   * https://www.helppost.gr/wp-content/uploads/fvm/out/footer-b36032d3-1521967386.min.js

Based on the previous, I think that covering 2 is the most important and maybe we can change the hash to a minimum of 5 (although I haven't seen many with a hash that small) 

What do you think?

UziTech:
I added new regexes to cover all of those cases.

- I covered #2 which actually let me remove the last modified time regexes since a timestamp is just a very high version number
- I changed the minimum hash length to 5. Since we are limiting it to hexadecimal I don't see any problems with small hashes (until someone creates a popular library called `deadbeef.js` 😄)

molant:
LGTM, @alrra @sarvaje do you want to take a look too?

@UziTech can you rebase from master? I think there are some commits that shouldn't be in here.

Thanks a lot!

packages/rule-http-cache/README.md
```diff
@@ -34,21 +34,37 @@ validate that the page and resources have a good caching strategy:
...
 ```regexp
-/\/(\w|-|_)+(\.|-|_)\w+\.\w+$/i
+/\/[^/]+[._-]v?\d+(\.\d+(\.\d+)?)?[^/]*\.\w+$/i
+/\/v?\d+\.\d+\.\d+.*?\//i
+/\/\d+\//
+/\/v\d.*?\//i
+/\/([^/]+[._-])?([0-9a-f]{5,})([._-].*?)?\.\w+$/i
 ```
 
 This will match urls like the following:
```
alrra: > URLs

packages/rule-http-cache/src/rule.ts
```diff
@@ -90,15 +90,44 @@ export default class HttpCacheRule implements IRule {
...
+
+            /*
+             * - https://cdn.example.com/jquery.lazy/1.6.5/jquery.lazy.min.js
+             */
+            /\/v?\d+\.\d+\.\d+.*?\//i,
+
+            /*
+             * - https://s3.amazonaws.com/example/wp-content/uploads/2017/07/15084048/jquery.mediaBoxes.dropdown.js
+             */
+            /\/\d+\//,
```
alrra: I think this is a bit to general and it will wrongly match a lot of online sites where you have URLs such as `.../id/human-redable-name-of-product`.

alrra:
@UziTech Thank you for all the great work! 💜

packages/rule-http-cache/src/rule.ts
UziTech: Which urls will it match that it shouldn't?

UziTech: I think it would be fine to assume that a script is versioned any time there is a folder of just numbers.

alrra: > any time there is a folder of just numbers.

You can have something like a blog, and then: `2018/<some_title>/<some_script_that_is_related_to_the_blog_post>.js`

UziTech: how are we supposed to distinguish between `/2017/07/<some_revved_wordpress_script>.js` and `/2017/07/<some_non_revved_blog_script>.js` ?

UziTech: How common is it to include non-revved scripts/css in a blog directory?

UziTech: should there be a length limit for the numbers?

```js
/\/\d{5,}\//
```

alrra: > how are we supposed to distinguish between 

We are not, and even if we try that won't be possible without over complicating things, which we shouldn't. The rule allows configurations exactly for cases like this one that are not common conventions or easily distinguishable.

> How common is it to include non-revved scripts/css in a blog directory?

That was just an example, but having `/<digits>/` is pretty common on the web without actually meaning versioning. 

---

As I said before, matching just `/digits/` is to general, so I'm going to merge this PR without this regex.

alrra:
@UziTech Your changes are now available in [`v3.1.0`](https://www.npmjs.com/package/@sonarwhal/rule-http-cache). 🎉

UziTech:
@alrra looks like the live example url still has the `/\d+/` listed

alrra:
> looks like the live example url still has the /\d+/ listed

@UziTech Sorry, I don't follow, where does the `/\d+/` still appear?

UziTech:
> Live example: https://regex101.com/r/6VduJO/36/

and

> [Test your URLs] (https://regex101.com/r/6VduJO/36/)

image

alrra:
@UziTech Good catch! Should be fixed by 99c01370c40f3e7adb0372cbd4826c4a0ffda6fa. Thanks!

PR #1101, opened 2018-05-25
Fix: Lazy load modules depending on the task

molant:
## Pull request checklist

Make sure you:

- [ ] ~~Added/Updated related documentation.~~

Use dynamic imports to load the task the user has decided to execute so
there are less modules loaded directly and the memory footprint is
smaller.

Ref: https://github.com/sonarwhal/sonarwhal/issues/896#issuecomment-392114447

### DIFF
--- a/packages/sonarwhal/src/lib/cli/actions.ts
+++ b/packages/sonarwhal/src/lib/cli/actions.ts
@@ -5,12 +5,11 @@ import { CLIOptions } from '../types';
 
 /** Wrapper to dynamically load the different CLI tasks depending on a condition */
 const action = (pkg: string, condition?: string): (actions: CLIOptions) => Promise<boolean> => {
-    return async (actions: CLIOptions): Promise<boolean> => {
-        if (!condition || actions[condition]) {
+    return async (options: CLIOptions): Promise<boolean> => {
+        if (!condition || options[condition]) {
             const { default: task } = await import(pkg);
 
-            // TODO: we shouldn't have to pass actions
-            return await task(actions);
+            return await task(options);
         }
 
         return false;
--- a/packages/sonarwhal/src/lib/cli/actions.ts
+++ b/packages/sonarwhal/src/lib/cli/actions.ts
@@ -1,13 +1,29 @@
 /**
  * @fileoverview Exports all the actions the CLI is capable of doing.
  */
-import { CLIOptions } from '../types'; // eslint-disable-line no-unused-vars
-import { newRule } from './wizards/new-rule';
-import { initSonarwhalrc } from './wizards/init';
-import { printHelp } from './help';
-import { printVersion } from './version';
-import { analyze } from './analyze';
-import { newParser } from './wizards/new-parser';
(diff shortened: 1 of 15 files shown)


packages/sonarwhal/src/lib/cli/analyze.ts
```diff
@@ -16,7 +16,7 @@ import * as logger from '../utils/logging';
 import { cutString } from '../utils/misc';
 import * as resourceLoader from '../utils/resource-loader';
 import { installPackages } from '../utils/npm';
-import { initSonarwhalrc } from './wizards/init';
+import initSonarwhalrc from './wizards/init';
```
sarvaje: this should be load only if we need it.

packages/sonarwhal/src/lib/cli/wizards/init.ts
```diff
@@ -163,8 +163,8 @@ const customConfig = async (): Promise<InitUserConfig> => {
  * * an existing published configuration package
  * * the installed resources
  */
-export const initSonarwhalrc = async (options: CLIOptions): Promise<boolean> => {
-    if (!options.init) {
+export default async (actions: CLIOptions): Promise<boolean> => {
+    if (!actions.init) {
```
sarvaje: This check shouldn't be necessary. Neither in the other actions

packages/sonarwhal/src/lib/cli/wizards/new-parser.ts
```diff
@@ -265,7 +265,7 @@ const generateFiles = async (data: NewParser) => {
  * ------------------------------------------------------------------------------
  */
 /** Generate a new parser. */
-export const newParser = async (actions: CLIOptions): Promise<boolean> => {
+export default async (actions: CLIOptions): Promise<boolean> => {
     if (!actions.newParser) {
```
sarvaje: This check shouldn't be necessary

packages/sonarwhal/src/lib/cli/actions.ts
```diff
@@ -1,13 +1,29 @@
...
-import { newParser } from './wizards/new-parser';
+import { CLIOptions } from '../types';
+
+/** Wrapper to dynamically load the different CLI tasks depending on a condition */
+const action = (pkg: string, condition?: string): (actions: CLIOptions) => Promise<boolean> => {
+    return async (actions: CLIOptions): Promise<boolean> => {
+        if (!condition || actions[condition]) {
+            const { default: task } = await import(pkg);
+
+            // TODO: we shouldn't have to pass actions
```
sarvaje: I think this depends on the action. `analyze` needs it. The problem I think that is calling the parameter `actions`, it should be called `options`.

alrra:
@molant It seems some [test hangs](https://travis-ci.org/sonarwhal/sonarwhal/jobs/383902391#L2706).

molant:
Yes, I need to check that but it won't be until next week when I'm back from 🇪🇸
Feel free if you want to fix them (or list which ones break) 😁

PR #1103, opened 2018-05-26
New: Add rule to check stylesheet limits

### LINKED ISSUE #570: Add rule to check for stylesheet limits
Ref: https://blogs.msdn.microsoft.com/ieinternals/2011/05/14/stylesheet-limits-in-internet-explorer/

(Suggested here: https://twitter.com/gregwhitworth/status/916338494036783104)

antross:
## Pull request checklist

Make sure you:

Fix #570 by adding checks for known stylesheet limits in IE9 (4095 rules, 31 stylesheets, and 4 nested imports) as well as the larger limits in IE10+ (65534 rules and 4095 stylesheets).

Currently nested import checks only trigger in the chrome connector due to a jsdom bug which ignores import rules. I disabled related tests for jsdom and left a comment so we can re-enable once the jsdom bug is fixed. All other checks trigger in both connectors.

### DIFF
--- a/packages/rule-stylesheet-limits/README.md
+++ b/packages/rule-stylesheet-limits/README.md
@@ -4,37 +4,28 @@ Checks if CSS exceeds known stylesheet limits.
 
 ## Why is this important?
 
-Internet Explorer prior to version 10 have limits on the
-number of CSS stylesheets, imports, and rules which are relatively small
-compared to modern browsers. Once these limits are exceeded, additional
-stylesheets, imports, and rules are ignored. Much larger limits exist in more
-recent versions that can also cause styles to be ignored if exceeded. For more
-details see [Stylesheet limits in Internet Explorer][stylesheet limits].
+Internet Explorer 9 and below has limits on the number of CSS stylesheets,
+imports, and rules which are relatively small compared to modern browsers.
+Once these limits are exceeded, additional stylesheets, imports, and rules
+are ignored. For more details see
+[Stylesheet limits in Internet Explorer][stylesheet limits].
 
 ## What does the rule check?
 
-When targeting versions of Internet Explorer less than 10, this rule checks if
+When targeting versions of Internet Explorer 9 and below, this rule checks if
 one of the following [limits][stylesheet limits] is exceeded:
 
 * 4095 rules
 * 31 stylesheets
 * 4 levels of imports
 
-When targeting modern browsers, this rule checks if one of the following
-[limits in Internet Explorer 10+][stylesheet limits] is exceeded:
-
-* 65535 rules
-* 4095 stylesheets
-
 ### Examples that **trigger** the rule
 
 * A page targeting Internet Explorer 9 containing 4096 or more CSS rules
-* A page targeting Internet Explorer 10 containing 65535 or more CSS rules
 
 ### Examples that **pass** the rule
 
 * A page targeting Internet Explorer 9 with fewer than 4096 CSS rules
-* A page targeting Internet Explorer 10 and up with fewer than 65535 rules
 
 ## Can the rule be configured?
 
--- a/packages/rule-stylesheet-limits/README.md
+++ b/packages/rule-stylesheet-limits/README.md
@@ -1,59 +1,40 @@
 # stylesheet-limits (`@sonarwhal/rule-stylesheet-limits`)
 
-Checks if CSS exceeds known stylesheet limits
+Checks if CSS exceeds known stylesheet limits.
 
 ## Why is this important?
 
-IE versions prior to IE 10 have limits on the number of CSS stylesheets, imports, and rules which are relatively small compared to modern browsers. Once these limits are exceeded, additional stylesheets, imports, and rules are ignored. Much larger limits exist in more recent versions, but can also cause styles to be ignored if exceeded. For more details see [Stylesheet limits in Internet Explorer][stylesheet limits].
-
-## How to use this rule?
-
-To use it you will have to install it via `npm`:
-
-```bash
-npm install @sonarwhal/rule-stylesheet-limits
-```
-
-Note: You can make `npm` install it as a `devDependency` using the `--save-dev`
-parameter, or to install it globally, you can use the `-g` parameter. For
-other options see
-[`npm`'s documentation](https://docs.npmjs.com/cli/install).
-
-And then activate it via the [`.sonarwhalrc`][sonarwhalrc]
-configuration file:
-
-```json
-{
-    "connector": {...},
-    "formatters": [...],
-    "parsers": [...],
-    "rules": {
-        "stylesheet-limits": "error"
-    },
-    ...
-}
-```
+Internet Explorer versions prior to Internet Explorer 10 have limits on the
+number of CSS stylesheets, imports, and rules which are relatively small
+compared to modern browsers. Once these limits are exceeded, additional
+stylesheets, imports, and rules are ignored. Much larger limits exist in more
+recent versions, but can also cause styles to be ignored if exceeded. For more
+details see [Stylesheet limits in Internet Explorer][stylesheet limits].
 
 ## What does the rule check?
 
-When targeting versions of IE less than 10, this rule checks if one of the following limits is exceeded:
+When targeting versions of Internet Explorer less than 10, this rule checks if
+one of the following limits is exceeded:
+
 * 4095 rules
 * 31 stylesheets
 * 4 levels of imports
 
-When targeting modern browsers, this rule checks if one of the following limits in IE10+ is exceeded:
+When targeting modern browsers, this rule checks if one of the following
+limits in Internet Explorer 10+ is exceeded:
+
 * 65535 rules
 * 4095 stylesheets
 
 ### Examples that **trigger** the rule
 
-* A page targeting IE 9 containing 4096 or more CSS rules
-* A page targeting IE 10 containing 65535 or more CSS rules
+* A page targeting Internet Explorer 9 containing 4096 or more CSS rules
+* A page targeting Internet Explorer 10 containing 65535 or more CSS rules
 
 ### Examples that **pass** the rule
 
-* A page targeting IE 9 with fewer than 4096 CSS rules
-* A page targeting browsers newer than IE 9 with fewer than 65534 rules
+* A page targeting Internet Explorer 9 with fewer than 4096 CSS rules
+* A page targeting Internet Explorer 10 and up with fewer than 65534 rules
 
 ## Can the rule be configured?
 
@@ -78,7 +59,34 @@ In the [`.sonarwhalrc`][sonarwhalrc] file:
 }
 ```
 
-## Further Reading
+## How to use this rule?
+
+To use it you will have to install it via `npm`:
+
+```bash
+npm install @sonarwhal/rule-stylesheet-limits
+```
+
+Note: You can make `npm` install it as a `devDependency` using the `--save-dev`
+parameter, or to install it globally, you can use the `-g` parameter. For
+other options see
+[`npm`'s documentation](https://docs.npmjs.com/cli/install).
+
+And then activate it via the [`.sonarwhalrc`][sonarwhalrc]
+configuration file:
+
+```json
+{
+    "connector": {...},
+    "formatters": [...],
+    "parsers": [...],
+    "rules": {
+        "stylesheet-limits": "error",
+        ...
+    },
+    ...
+}
+```
 
 <!-- Link labels: -->
 
--- a/packages/rule-stylesheet-limits/README.md
+++ b/packages/rule-stylesheet-limits/README.md
@@ -1,0 +1,81 @@
+# stylesheet-limits (`@sonarwhal/rule-stylesheet-limits`)
+
+Checks if CSS exceeds known stylesheet limits
+
+## Why is this important?
+
+IE versions prior to IE 10 have limits on the number of CSS stylesheets, imports, and rules which are relatively small compared to modern browsers. Once these limits are exceeded, additional stylesheets, imports, and rules are ignored. Much larger limits exist in more recent versions, but can also cause styles to be ignored if exceeded. For more details see [Stylesheet limits in Internet Explorer][stylesheet limits].
+
+## How to use this rule?
+
+To use it you will have to install it via `npm`:
+
+```bash
+npm install @sonarwhal/rule-stylesheet-limits
+```
+
+Note: You can make `npm` install it as a `devDependency` using the `--save-dev`
+parameter, or to install it globally, you can use the `-g` parameter. For
+other options see
+[`npm`'s documentation](https://docs.npmjs.com/cli/install).
+
+And then activate it via the [`.sonarwhalrc`][sonarwhalrc]
+configuration file:
+
+```json
+{
+    "connector": {...},
+    "formatters": [...],
+    "parsers": [...],
+    "rules": {
+        "stylesheet-limits": "error"
+    },
+    ...
+}
+```
+
+## What does the rule check?
+
+When targeting versions of IE less than 10, this rule checks if one of the following limit is exceeded:
+* 4095 rules
+
+When targeting modern browsers, this rule checks if one of the following limit in IE10+ is exceeded:
+* 65535 rules
+
+### Examples that **trigger** the rule
+
+* A page targeting IE 9 containing 4096 or more CSS rules
+* A page targeting IE 10 containing 65535 or more CSS rules
+
+### Examples that **pass** the rule
+
+* A page targeting IE 9 with fewer than 4096 CSS rules
+* A page targeting browsers newer than IE 9 with fewer than 65534 rules
+
+## Can the rule be configured?
+
+You can overwrite the defaults by specifying custom values for the
+number of CSS rules to allow.
+
+In the [`.sonarwhalrc`][sonarwhalrc] file:
+
+```json
+{
+    "connector": {...},
+    "formatters": [...],
+    "rules": {
+        "stylesheet-limit": ["error", {
+            "max-css-rules": 1000
+        }],
+        ...
+    },
+    ...
+}
+```
+
+## Further Reading
+
+<!-- Link labels: -->
+
+[sonarwhalrc]: https://sonarwhal.com/docs/user-guide/further-configuration/sonarwhalrc-formats/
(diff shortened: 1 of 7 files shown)


packages/rule-stylesheet-limits/README.md
```diff
@@ -0,0 +1,86 @@
+# stylesheet-limits (`@sonarwhal/rule-stylesheet-limits`)
+
+Checks if CSS exceeds known stylesheet limits
```
molant: Missing `.` at the end

packages/rule-stylesheet-limits/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Why is this important?
+
+IE versions prior to IE 10 have limits on the number of CSS stylesheets, imports, and rules which are relatively small compared to modern browsers. Once these limits are exceeded, additional stylesheets, imports, and rules are ignored. Much larger limits exist in more recent versions, but can also cause styles to be ignored if exceeded. For more details see [Stylesheet limits in Internet Explorer][stylesheet limits].
+
+## How to use this rule?
```
molant: We've moved this towards the end of the file. Check the other rules to see where it is now. We should update the `new-rule` generator 🙄

packages/rule-stylesheet-limits/README.md
```diff
@@ -0,0 +1,86 @@
...
+            "maxSheets": 10,
+            "maxImports": 2
+        }],
+        ...
+    },
+    ...
+}
+```
+
+## Further Reading
```
molant: If there isn't any article about this (other than the previously linked) we can delete this section

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,141 @@
+/**
+ * @fileoverview Checks if CSS exceeds known stylesheet limits
+ */
+
+import { Category } from 'sonarwhal/dist/src/lib/enums/category';
+import { RuleContext } from 'sonarwhal/dist/src/lib/rule-context';
+// The list of types depends on the events you want to capture.
```
molant: Remove this comment

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+        schema: [{
+            additionalProperties: false,
+            properties: {
+                maxRules: { type: 'number' },
+                maxSheets: { type: 'number' },
+                maxImports: { type: 'number' }
+            },
+            type: ['object', 'null']
+        }],
+        scope: RuleScope.any
```
molant: Looking at how you get the data I believe the scope should be `browser`.

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+    public constructor(context: RuleContext) {
+
+        const includesIE9 = context.targetedBrowsers.includes('ie 9');
+
+        let maxRules = includesIE9 ? 4095 : 65534;
+        let maxSheets = includesIE9 ? 31 : 4095;
+        let maxImports = includesIE9 ? 4 : 0;
+
+        //  allow limits to be overridden by rule options
+        if (context.ruleOptions) {
```
molant: I'm not sure about this approach. I can see someone targetting IE9 increasing the number of `maxRules` to 5000 which is not great instead of removing the browser from their `browserslist`.
Maybe we should do a `Math.min()`?

alrra:
@antross Did you have the time to look into what limitations other browsers have?

antross:
@molant I'll fix the linting errors, but is there a way to have `yarn` run the linter in watch mode to catch these issues earlier next time? I was running `yarn watch` while developing the rule.

@alrra I ran a few searches looking for limits in other browsers, but nothing as well documented popped up. There was one post referencing a limit to the length of an individual CSS selector in Chrome, but with disagreement as to the actual limit (https://stackoverflow.com/questions/20828995/how-long-can-a-css-selector-be). I can build a test page to run cross-browser if we'd like to explore this a bit further.

molant:
> is there a way to have yarn run the linter in watch mode to catch these issues earlier next time?

What I usually do is have 2 terminals with `yarn watch:ts` and `yarn watch:test` but that doesn't cover the linting part. We could add a `yarn watch:test-all` or something similar that runs the tests and linting although I'm not sure the output would be very legible (but we can try).

packages/rule-stylesheet-limits/src/rule.ts
antross: I see `any`, `local`, and `site` and options for `RuleScope`. Were you referring to `site`?

I should also note that aside from the `@import` limitation mentioned in the PR description, `jsdom` does parse and populate the `document.styleSheets` collection meaning much of the testing can still be done in the `local` scope if I'm understanding the environment correctly.

packages/rule-stylesheet-limits/tests/tests.ts
```diff
@@ -0,0 +1,101 @@
...
+                name: `Page${label} contains ${maxImports} nested imports`,
+                reports: [{ message: `Maximum of ${maxImports} nested imports reached (${maxImports})` }],
+                serverConfig: generateImports(maxImports)
+            }
+
+        /*
+         * Exclude `jsdom` since it currently ignores `@import` rules.
+         * https://github.com/jsdom/jsdom/issues/2124
+         */
+        ], { ...configs, ignoredConnectors: ['jsdom'] });
```
antross: @molant @alrra Looks like my use of the object spread syntax here is causing failures in CI. I can change this to something different if you'd like or we can dig down to try and figure out why this isn't getting transpiled in that context.

packages/rule-stylesheet-limits/src/rule.ts
molant: You are correct, it's `site`.

`local` reads files individually from the disk so it will not work for now unless we add a css parser.

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,153 @@
...
+            category: Category.interoperability,
+            description: `Checks if CSS exceeds known stylesheet limits.`
+        },
+        id: 'stylesheet-limits',
+        schema: [{
+            additionalProperties: false,
+            properties: {
+                maxImports: { type: 'number' },
+                maxRules: { type: 'number' },
+                maxSheets: { type: 'number' }
```
alrra: The schema should specify that the number is positive and it's an integer/whole number,

alrra: Maybe use something such as:

```js
{
    additionalProperties: false,
    definitions: {
        number: { 
            minimum: 0,
            type: 'integer'
        }
    },    
    properties: {
        maxImports: { "$ref": "#/definitions/number" },                 
        maxRules: { "$ref": "#/definitions/number" },                 
        maxSheets: { "$ref": "#/definitions/number" }               
    }
    type: ['object', 'null']
}
```

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,153 @@
...
+            };
+
+            const combinedResults = {
+                imports: 0,
+                rules: 0,
+                sheets: 0
+            };
+
+            // Get the recursive count of rules for all stylesheets in the page.
+            Array.from(document.styleSheets).forEach((sheet) => {
```
alrra: Do rules in `<style></style>` count?

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,153 @@
...
+            return combinedResults;
+        };
+
+        const validateScanEnd = async () => {
+
+            const results = await context.evaluate(`(${injectedCode})()`);
+
+            // Report once we hit a limit to support flagging on platforms which will drop subsequent rules.
+
+            if (maxImports && results.imports >= maxImports) {
```
alrra: Why check for `maxImports` ?

alrra:
>  I can build a test page to run cross-browser if we'd like to explore this a bit further.

I think it would be great to digg a bit into it, but don't waste to much time.

Also, are there any [file size limitations](http://joshua.perina.com/africa/gambia/fajara/post/internet-explorer-css-file-size-limit)?

packages/rule-stylesheet-limits/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+* 4095 rules
+* 31 stylesheets
+* 4 levels of imports
+
+When targeting modern browsers, this rule checks if one of the following
+limits in Internet Explorer 10+ is exceeded:
+
+* 65535 rules
+* 4095 stylesheets
```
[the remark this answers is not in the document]
antross: The post linked in the "Why is this important" section covers both the IE9 and IE10 limits(https://blogs.msdn.microsoft.com/ieinternals/2011/05/14/stylesheet-limits-in-internet-explorer/), but I can reference it again here if you'd prefer.

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,153 @@
...
+                maxSheets: { type: 'number' }
+            },
+            type: ['object', 'null']
+        }],
+        scope: RuleScope.any
+    }
+
+    public constructor(context: RuleContext) {
+
+        const includesIE9 = context.targetedBrowsers.includes('ie 9');
```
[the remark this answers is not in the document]
antross: This seemed sufficient from a practical perspective, but I can certainly change it to be exhaustive. Unless I'm missing something, it looks like I'll need to individually check for each prior version in the `targetBrowsers` collection. How far back do `sonarwhal` rules typically cover in these cases? IE6? IE7?

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,153 @@
...
+
+        let maxImports = includesIE9 ? 4 : 0;
+        let maxRules = includesIE9 ? 4095 : 65534;
+        let maxSheets = includesIE9 ? 31 : 4095;
+
+        // Allow limits to be overridden by rule options.
+        const options = context.ruleOptions;
+
+        if (options) {
+            if (options.maxImports && (maxImports === 0 || options.maxImports < maxImports)) {
```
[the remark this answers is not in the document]
antross: To make sure the `ruleOptions` specified value is still respected if we don't have a default `maxImports` limit (is `0`). I'll add a code comment here to clarify.

packages/rule-stylesheet-limits/src/rule.ts
antross: In IE10 there's no limit (other than the max number of stylesheets), so maxImports is 0 to indicate it's unspecified. I'll add a code comment to clarify.

packages/rule-stylesheet-limits/src/rule.ts
antross: Yes, all rules affecting the document count whether they came from an inline style element, a linked stylesheet, or an imported stylesheet. As written I believe this catches all of the above.

packages/rule-stylesheet-limits/src/rule.ts
antross: Sounds good. Thanks for the example!

antross:
> Also, are there any file size limitations?

@alrra I thought I ran across mention of some in older versions of Chrome, but can't seem to find the reference again at the moment. Would you like to include the IE limits in this PR or leave it as a separate task?

packages/rule-stylesheet-limits/README.md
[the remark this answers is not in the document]
alrra: Reference it again. Thanks!

packages/rule-stylesheet-limits/src/rule.ts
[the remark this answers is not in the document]
alrra: >  I'll need to individually check for each prior version in the targetBrowsers collection.

You can do something like [this](https://github.com/sonarwhal/sonarwhal/blob/cb524ca3b20a0bb7205561aa813b92b448c9ee8a/packages/rule-highest-available-document-mode/src/rule.ts#L202-L204).

> ow far back do sonarwhal rules typically cover in these cases? IE6? IE7?

We usually go as far as we can/know. IE 6 and up should be fine.

packages/rule-stylesheet-limits/README.md
```diff
@@ -0,0 +1,94 @@
+# stylesheet-limits (`@sonarwhal/rule-stylesheet-limits`)
```
alrra: Needs better title.

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+                    combinedResults.sheets += subResults.sheets;
+                }
+            });
+
+            return combinedResults;
+        };
+
+        const validateScanEnd = async () => {
+
+            const results = await context.evaluate(`(${injectedCode})()`);
```
sarvaje: Does these limits apply to other browsers or just to IE?

antross: Just IE. I'm sure other browsers have limits somewhere, but they haven't been well documented and online reports are inconsistent. The higher limits in IE10+ have been documented so I included them for completeness.

sarvaje: Should we check then the target browsers before evaluate? If they are not targeting IE10, then the rules shouldn't be executed, right?

antross: Sorry, I meant just IE and Edge. I tested and confirmed the higher IE10 limits still apply to the latest Edge version.

sarvaje: ok!

antross:
FYI: due to some performance limitations in JSDOM's handling of large numbers of stylesheets, I'm going to update this PR to focus on just the IE9 limits and leave out the IE10+ limits for now.

packages/rule-stylesheet-limits/tests/tests.ts
```diff
@@ -0,0 +1,96 @@
+import { generateHTMLPage } from 'sonarwhal/dist/tests/helpers/misc';
+import * as ruleRunner from 'sonarwhal/dist/tests/helpers/rule-runner';
+
+const ruleName = 'stylesheet-limits';
```
alrra: Use:

```ts
import { getRuleName } from 'sonarwhal/dist/src/lib/utils/rule-helpers';

const ruleName = getRuleName(__dirname);
```

packages/rule-stylesheet-limits/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+    "compilerOptions": {
+        "outDir": "dist"
```
sarvaje: Can you add the option `"strict": true`? if there is errors just in your rule, fix them and keep the option, if there is a lot of errors in `sonarwhal` that are not easy to fix, then remove it.

The idea is adding the strict option step by step until the whole project use it.

Thanks!!!

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,177 @@
...
```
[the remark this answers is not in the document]
antross: Ok, I'll take a look. Thanks for figuring this out!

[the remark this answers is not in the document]
antross: Digging in I see that `injectedCode` when run via `yarn test` has additional statements like `cov_l3mx4wetg.s[46]++;` included by the code coverage utility `nyc`. These generate a script error when the code is actually injected since it runs in a different context.

Adding `/* istanbul ignore next */` before the definition of `injectedCode` fixes the issue. I'll update the PR shortly.

packages/rule-stylesheet-limits/tsconfig.json
antross: Yes, I'll add this in my next update.

packages/rule-stylesheet-limits/src/rule.ts
```diff
@@ -0,0 +1,178 @@
...
+        });
+
+        if (!options && !includesOldIE) {
+            // Exit if we don't have any limits to test.
+            return;
+        }
+
+        let hasImportLimit = includesOldIE;
+        let hasRuleLimit = includesOldIE;
+        let hasSheetLimit = includesOldIE;
```
alrra: Why not just use `includesOldIE`?

packages/rule-stylesheet-limits/README.md
```diff
@@ -0,0 +1,87 @@
...
+### Examples that **pass** the rule
+
+* A page targeting Internet Explorer 9 with fewer than 4096 CSS rules
+* A page not targeting Internet Explorer 9 or below regardless of the number
+  of CSS rules
+
+## Can the rule be configured?
+
+You can overwrite the defaults by specifying custom values for the
+number of CSS rules to allow.
```
alrra: Add note if about what happens if the custom values are below the IE values.

packages/rule-stylesheet-limits/package.json
```diff
@@ -0,0 +1,70 @@
...
+  "description": "Checks if CSS exceeds known stylesheet limits",
+  "devDependencies": {
+    "@types/debug": "0.0.30",
+    "@types/node": "8.0.14",
+    "ava": "^0.25.0",
+    "cpx": "^1.5.0",
+    "eslint": "^4.19.1",
+    "eslint-plugin-markdown": "^1.0.0-beta.7",
+    "eslint-plugin-typescript": "^0.12.0",
+    "markdownlint-cli": "^0.8.1",
```
alrra: Some of the  `devDependencies` are outdated compare to master.

(I will update them before merging)

packages/rule-stylesheet-limits/README.md
```diff
@@ -0,0 +1,97 @@
...
+    },
+    ...
+}
+```
+
+<!-- Link labels: -->
+
+[sonarwhalrc]: https://sonarwhal.com/docs/user-guide/further-configuration/sonarwhalrc-formats/
+[stylesheet limits]: https://blogs.msdn.microsoft.com/ieinternals/2011/05/14/stylesheet-limits-in-internet-explorer/
+[chrome limits]: https://stackoverflow.com/questions/20828995/how-long-can-a-css-selector-be]
```
alrra: For consistency with the other rules, these should be ordered alphabetically.

(I will fix this before merging).

alrra:
@antross The rule is now [published!](https://www.npmjs.com/package/@sonarwhal/rule-stylesheet-limits).

Issue #1074, opened 2018-05-16
Split error messages in `http-cache` rule [0.1]


ramonTC:
Analyzing http-cache performance issues it seems that http-cache rule do not work properly in that case.

![image](https://user-images.githubusercontent.com/16465474/40105029-1eeff954-58f2-11e8-9dda-e74a9d6ee344.png)

Checking out the response headers cache-control, it is really setting the correct values. See below:

![image](https://user-images.githubusercontent.com/16465474/40105184-8c19e846-58f2-11e8-8ff7-02e385358fd7.png)

molant:
Currently the `http-cache` will show the same message if you don't have a long `max-age` and/or you are missing the `immutable` directive.

To avoid confussion I think we should split this in 2 messages:

* Static resources should have a long cache value (31536000)
* Static resources should use the immutable directive

@sonarwhal/core, @kshyju, @qzhou1607 what do you think?

kshyju:
TIL, chrome does not support immutable!  😆

PR #1112, opened 2018-06-01
Add an example of what an immutable cache control header might look like

Munter:
Docs: Add example of immutable cache header to error message

When seeing the below error message about far future expiring cache headers, it's not very actionable. Add an example of how to fix the error

image

### DIFF
--- a/packages/rule-http-cache/src/rule.ts
+++ b/packages/rule-http-cache/src/rule.ts
@@ -381,7 +381,7 @@ export default class HttpCacheRule implements IRule {
 
             // We want long caches with "immutable" for static resources
             if (usedDirectives.has('no-cache') || !(longCache && immutable)) {
-                const message: string = `Static resources should have a long cache value (${maxAgeResource}) and use the immutable directive:\n${header}`;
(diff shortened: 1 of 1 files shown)


alrra:
> Add an example of how to fix the error

@Munter The examples are in the [documentation](https://sonarwhal.com/docs/user-guide/rules/rule-http-cache/#examples-that-pass-the-rule) (even [how to configure some servers](https://sonarwhal.com/docs/user-guide/rules/rule-http-cache/#how-to-configure-the-server-to-pass-this-rule))?



Do you feel like the documentation link is not clear enough? 

---

Note: If you think something can be improved related to the docs, [this is the file](https://github.com/sonarwhal/sonarwhal/blob/e9623e3c2a09f259d6fc6fb5bbd2fcb029a51151/packages/rule-http-cache/README.md) that should be changed.

molant:
Maybe in this case we could make the error message clearer and have 2 different messages? One for long cache assets and another for immutable.

molant:
BTW, the split message is tracked in #1074
Closing this PR in favor of the work in that issue.

PR #1122, opened 2018-06-06
Fix: Remove `pluralize` dependency

alrra:
## Pull request checklist

Make sure you:

### DIFF
--- a/packages/formatter-codeframe/src/formatter.ts
+++ b/packages/formatter-codeframe/src/formatter.ts
@@ -15,7 +15,6 @@ import chalk from 'chalk';
 import * as groupBy from 'lodash.groupby';
 import * as sortBy from 'lodash.sortby';
 import * as forEach from 'lodash.foreach';
-import * as pluralize from 'pluralize';
 import * as logSymbols from 'log-symbols';
 
 import { cutString } from 'sonarwhal/dist/src/lib/utils/misc';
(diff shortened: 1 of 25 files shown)


packages/formatter-codeframe/src/formatter.ts
```diff
@@ -160,6 +159,6 @@ export default class CodeframeFormatter implements IFormatter {
 
         const color: typeof chalk = totalErrors > 0 ? chalk.red : chalk.yellow;
 
-        logger.log(color.bold(`${logSymbols.error} Found a total of ${totalErrors} ${pluralize('error', totalErrors)} and ${totalWarnings} ${pluralize('warning', totalWarnings)}`));
+        logger.log(color.bold(`${logSymbols.error} Found a total of ${totalErrors} ${totalErrors ? 'errors' : 'error'} and ${totalWarnings} ${totalWarnings !== 1 ? 'warnings' : 'warnings'}`));
```
molant: You are using `'warnings'` in both expressions.

Also for all the conditions I think we should use
```
=== 1 ? 'warning' : 'warnings' // or equivalent
```
instead of 
```
!== 1 ? 'warnings' : 'warning'
```

The reason is that dealing with negations is overhead that we can avoid. It took me a couple seconds to know exactly what expression was executed in each case and it would have been straightforward the other way.

packages/formatter-summary/src/formatter.ts
```diff
@@ -43,7 +42,7 @@ export default class SummaryFormatter implements IFormatter {
         }
 
         const buildMessage = (count, type) => {
-            return `${count} ${pluralize(type, count)}`;
+            return `${count} ${count !== 1 ? `${type}s` : type}`;
```
molant: I don't like the idea of template literals inside template literals in this case. I think the following is easier to understand:

```js
${count} ${type}${count === 1 : '' ? 's'}
```

packages/formatter-codeframe/src/formatter.ts
alrra: (Plural was usually the more common case, so it was a micro optimization.)

I changed things to `=== 1`.

packages/formatter-summary/src/formatter.ts
alrra: Done.

alrra:
> Why don't we have our own util to pluralize?

@sarvaje `eslint` depends on it.

alrra:
> Why don't we have our own util to pluralize?

I don't think we need yet, but we can consider it in the future?

(sorry I misread your comment the first time)

PR #1125, opened 2018-06-07
Fix: Load dependencies by path in configuration

sarvaje:
## Pull request checklist

Make sure you:

### DIFF
--- a/packages/sonarwhal/src/lib/sonarwhal.ts
+++ b/packages/sonarwhal/src/lib/sonarwhal.ts
@@ -9,6 +9,8 @@
  * ------------------------------------------------------------------------------
  */
 
+import * as fs from 'fs';
+import * as path from 'path';
 import * as url from 'url';
 
 import * as browserslist from 'browserslist';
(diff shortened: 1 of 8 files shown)


sarvaje:
@molant @alrra, If you can take a look to see if it is a good solution it would be great :)

packages/sonarwhal/src/lib/utils/resource-loader.ts
```diff
@@ -230,33 +231,21 @@ const generateConfigPathsToResources = (configurations: Array<string>, name: str
...
- * 4. external rules
- *
- */
-export const loadResource = (name: string, type: ResourceType, configurations: Array<string> = [], verifyVersion = false) => {
-    debug(`Searching ${name}…`);
+type ResourceInfo = {
+    packageName: string;
+    resourceName: string;
+    key: string;
+    sources: Array<string>;
```
alrra: Can you document what these fields mean? Thanks!

packages/sonarwhal/src/lib/sonarwhal.ts
```diff
@@ -85,6 +89,33 @@ export class Sonarwhal extends EventEmitter {
         });
     }
 
+    private resolveRule(key: string): string {
```
molant: Not a fan of this part. It's weird to have to do all of this and it will slow down the start up process. We need to find another way.
Maybe `resources` should use the value in `.sonarwhalrc` in the values it returns. We can access the `id` via `meta.id` and do the matching from there.

packages/sonarwhal/src/lib/utils/resource-loader.ts
molant: Should we move this to the `types` folder?

alrra:
Closing in favor of #1134.

PR #1134, opened 2018-06-13
New: Use path to load resources in `.sonarwhalrc`

### LINKED ISSUE #1123: Support for paths to resources on .sonarwhalrc
*Original issue at the end.* 

On top of using package names, we should allow paths.

The scenarios that this unlocks are:

* **Shared configurations in the same project** There are some rules that apply to all moments of the development process, but some others that are only valid when the site is in staging or in production. For example: I could test my website locally, but I'm probably not going to compress my assets using brotli all the time. This way we can have a common file that we can extend from and have those configurations as part of the project:

```json
{
  "extends": "./common/.sonarwhalrc",
  "connector": "local"
}
```

* **Custom resources for the project** Having to create a package for a custom rule/parser is some overhead that not everybody is willing to do (and TBH, not worth it if it is very specific). This way there's no need to publish a package to create it (we might want to improve the wizard and just create a single file with no testing files for this scenario?):

```json
{
  ...
  "rules": {
    "./rules/my-custom-rule.js": "error"
  }
}
```

- - - - - - - - - - - - - - - - - - -

Steps to reproduce:
1. Go to rule-no-vulnerable-javascript-libraries
1. Run npm install
1. Run npm test

molant:
Enable loading a resource using a path instead of the ID of the package.
This is useful when you have multiple configuration files with some
common parts or when you have a locally developed rule that you don't
want to publish on a package.

Fix #1123

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/sonarwhal/src/lib/utils/resource-loader.ts
+++ b/packages/sonarwhal/src/lib/utils/resource-loader.ts
@@ -12,6 +12,7 @@
  */
 
 import * as path from 'path';
+import * as fs from 'fs';
 
 import * as globby from 'globby';
 import * as semver from 'semver';
@@ -244,11 +245,13 @@ const generateConfigPathsToResources = (configurations: Array<string>, name: str
  */
 export const loadResource = (name: string, type: ResourceType, configurations: Array<string> = [], verifyVersion = false) => {
     debug(`Searching ${name}…`);
-
+    const isSource = fs.existsSync(name); // eslint-disable-line no-sync
     const nameSplitted = name.split('/');
 
     const packageName = nameSplitted[0];
-    const resourceName = nameSplitted[1] || packageName;
+    const resourceName = isSource ?
+        name :
+        nameSplitted[1] || packageName;
 
     const key: string = `${type}-${name}`;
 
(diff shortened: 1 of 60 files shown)


molant:
Going to add tests and change some stuff in rule-runner, but you can take a look at this. If we are ok with this approach then #1125 can be closed.

molant:
Pending things:
* [ ] Update documentation to explain how to things work with a path
* [ ] Update `rule-runner` to use a path so the release script works OK
* [ ] Add tests

packages/sonarwhal/src/lib/utils/resource-loader.ts
```diff
@@ -244,11 +245,13 @@ const generateConfigPathsToResources = (configurations: Array<string>, name: str
  */
 export const loadResource = (name: string, type: ResourceType, configurations: Array<string> = [], verifyVersion = false) => {
     debug(`Searching ${name}…`);
-
+    const isSource = fs.existsSync(name); // eslint-disable-line no-sync
```
sarvaje: if name is a relative path, this is not going to work, isn't it?

packages/sonarwhal/src/lib/utils/resource-loader.ts
```diff
@@ -244,11 +245,13 @@ const generateConfigPathsToResources = (configurations: Array<string>, name: str
...
     const nameSplitted = name.split('/');
 
     const packageName = nameSplitted[0];
-    const resourceName = nameSplitted[1] || packageName;
+    const resourceName = isSource ?
+        name :
+        nameSplitted[1] || packageName;
 
     const key: string = `${type}-${name}`;
```
sarvaje: key should be something like 
```
const key: string = isSource ? name : `${type}-${name}`
```

packages/sonarwhal/src/lib/utils/resource-loader.ts
```diff
@@ -322,7 +334,12 @@ export const loadResource = (name: string, type: ResourceType, configurations: A
         throw new ResourceError(`Resource ${name} not found`, ResourceErrorStatus.NotFound);
     }
 
-    resources.set(key, resource);
+    if (isSource) {
```
sarvaje: You will no longer need this if you do the `key` change

packages/sonarwhal/tests/lib/utils/rule-helpers.ts
```diff
@@ -143,15 +143,15 @@ test.serial(`getRuleName - returns an empty string if it can't determine the rul
     t.deepEqual(ruleName, '');
 });
 
-test.serial(`getRuleName - returns a combined string if a package name is passed`, (t) => {
-    const sandbox = sinon.createSandbox();
+// test.serial(`getRuleName - returns a combined string if a package name is passed`, (t) => {
```
sarvaje: uncomment or remove it :)

sarvaje:
@molant should I close https://github.com/sonarwhal/sonarwhal/pull/1125 in favor of this PR?

packages/sonarwhal/src/lib/utils/resource-loader.ts
molant: It was working on my local tests. IIRC `fs. exists` uses `process.cwd()` so it should be ok?
I guess the problem is if we are calling from another place different place than where the `.sonarwhalrc` file is located (e.g.: when using the file in the user's home folder)

packages/sonarwhal/tests/lib/utils/rule-helpers.ts
molant: I have to rewrite all the `getRuleName` tests. None work after this and some of the previous logic didn't make any sense.

molant:
TL;DR: this is going to take longer than I expected if we want to do it right

Longer version:
Supporting paths in the initial configuration is relatively easy (the current code should handle it). The problem is if we want to support paths in "extended" configurations and those configurations are in different paths.
Resources should be loaded relatively to the config that declares them. The way I'm currently working on is to change that to absolute paths but I'll have to change more things in `UserConfig` and `SonarwhalConfig`. Problem is that `UserConfig` knows about paths but not extended configuration and `SonarwhalConfig` is the opposite, so I'll have to do a few changes around both 😔

molant:
New push that loads the path resources relatively to where the configuration file is.

molant:
Closing in favor of #1140 that should be ready to review 😊

PR #1191, opened 2018-07-24
Fix: Allow additional keywords in `rel` value

### LINKED ISSUE #1188: Feature Request: Require "X-Content-Type-Options: nosniff" HTTP response header for alternate stylesheets
I noticed Sonarwhal generates a Security Error when alternate stylesheets are served with the "X-Content-Type-Options: nosniff" HTTP response header (eg. "'x-content-type-options' header is not needed"). However, I see no reason alternate stylesheets should be treated differently than regular stylesheets; the MIME-type confusion attack vector is the same as browsers (or extensions) may allow users to enable alternate stylesheets at their discretion.

In other words, I believe the "X-Content-Type-Options: nosniff" HTTP header _should_ be required of both stylesheets and alternate stylesheets and alternate stylesheets served with this header _should not_ trigger the "header is not needed" Security Error.

alrra:
## Pull request checklist

Make sure you:

This change makes it so that cases such as `rel="alternate stylesheet"` are allowed.

See also: https://html.spec.whatwg.org/multipage/links.html#rel-alternate

Fix #1188

### DIFF
--- a/packages/hint-x-content-type-options/src/hint.ts
+++ b/packages/hint-x-content-type-options/src/hint.ts
@@ -39,6 +39,34 @@ export default class XContentTypeOptionsHint implements IHint {
 
     public constructor(context: HintContext) {
 
+        const isHeaderRequired = (element: IAsyncHTMLElement): boolean => {
+            if (!element) {
+                return false;
+            }
+
+            const nodeName = normalizeString(element.nodeName);
+
+            /*
+             * See:
+             *
+             *  * https://github.com/whatwg/fetch/issues/395
+             *  * https://fetch.spec.whatwg.org/#x-content-type-options-header
+             */
+
+            if (nodeName === 'script') {
+                return true;
+
+            }
+
+            if (nodeName === 'link') {
+                const relValues = (normalizeString(element.getAttribute('rel'), '')).split(' ');
+
+                return relValues.includes('stylesheet');
+            }
+
+            return false;
+        };
+
         const validate = async (fetchEnd: FetchEnd) => {
             const { element, resource, response }: { element: IAsyncHTMLElement, resource: string, response: Response } = fetchEnd;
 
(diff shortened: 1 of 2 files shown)


packages/hint-x-content-type-options/src/hint.ts
```diff
@@ -63,7 +63,12 @@ export default class XContentTypeOptionsHint implements IHint {
              */
 
             if (nodeName === 'script' ||
-                (nodeName === 'link' && normalizeString(element.getAttribute('rel')) === 'stylesheet')) {
+                (nodeName === 'link' &&
+                    (normalizeString(element.getAttribute('rel'), ''))
```
sarvaje: I think you should move this to an `isStylesheet` method.

alrra: @sarvaje I refactored the code a bit.

packages/hint-x-content-type-options/src/hint.ts
```diff
@@ -39,6 +39,34 @@ export default class XContentTypeOptionsHint implements IHint {
...
+             *  * https://fetch.spec.whatwg.org/#x-content-type-options-header
+             */
+
+            if (nodeName === 'script') {
+                return true;
+
+            }
+
+            if (nodeName === 'link') {
+                const relValues = (normalizeString(element.getAttribute('rel'), '')).split(' ');
```
sarvaje: I have a question, is the `split` necessary?

packages/hint-x-content-type-options/src/hint.ts
sarvaje: @alrra I like the refactor :)

packages/hint-x-content-type-options/src/hint.ts
alrra: @sarvaje Yes, because of cases such as `<link rel="  alternate stylesheet " href="test.css">'`

sarvaje: but the result will be the same if you do a ' alternate stylesheet '.includes('stylesheet'), without splitting it.

alrra: @sarvaje 'stylesheetx '.includes('stylesheet') => `true` (should be `false`)

sarvaje: ok!

PR #1194, opened 2018-07-25
New: CSS Parser

### LINKED ISSUE #1115: Add a CSS parser
Including a CSS parser would help in the creation of rules focused on interoperability in the CSS space. We can do some of this now using script injection by walking the CSSOM, but this has limitations such as only being able to be run in a browser environment. Having a CSS parser would enable analysis to be done statically in addition to run-time and enable checks for patterns that may be ignored and therefore inaccessible via CSSOM.

antross:
## Pull request checklist

Make sure you:

Fix #1115. Built on top of PostCSS. The `ast` property on `StyleParse`
references the PostCSS AST types to enable type validation and
autocomplete when writing rules. Currently only supports CSS, but
support for LESS and SASS can be added in the future using
existing PostCSS plugins.

### DIFF
--- a/packages/parser-css/package.json
+++ b/packages/parser-css/package.json
@@ -1,0 +1,70 @@
+{
+  "ava": {
+    "failFast": false,
+    "files": [
+      "dist/tests/**/*.js"
+    ],
+    "timeout": "1m"
+  },
+  "dependencies": {
+    "postcss": "^7.0.1"
+  },
+  "description": "webhint parser needed to analyze CSS files",
+  "devDependencies": {
+    "ava": "^0.25.0",
+    "cpx": "^1.5.0",
+    "eslint-plugin-import": "^2.13.0",
+    "eslint-plugin-markdown": "^1.0.0-beta.7",
+    "eslint-plugin-typescript": "^0.12.0",
+    "hint": "^3.0.0-beta.4",
+    "markdownlint-cli": "^0.11.0",
+    "npm-link-check": "^2.0.0",
+    "npm-run-all": "^4.1.2",
+    "nyc": "^12.0.2",
+    "proxyquire": "2.0.0",
+    "rimraf": "^2.6.2",
+    "sinon": "^6.1.4",
+    "typescript": "^2.9.2",
+    "typescript-eslint-parser": "^16.0.1"
+  },
+  "engines": {
+    "node": ">=8.0.0"
+  },
+  "files": [
+    "dist/src",
+    "npm-shrinkwrap.json"
+  ],
+  "homepage": "https://webhint.io/",
+  "keywords": [
+    "webhint",
+    "webhint-parser",
+    "css"
+  ],
+  "license": "Apache-2.0",
+  "main": "dist/src/parser.js",
+  "name": "@hint/parser-css",
+  "nyc": {
+    "extends": "../../.nycrc"
+  },
+  "peerDependencies": {
+    "hint": "^3.0.0-beta.4"
+  },
+  "repository": "webhintio/hint",
+  "scripts": {
+    "build": "npm run clean && npm-run-all build:*",
+    "build-release": "npm run clean && npm run build:assets && tsc --inlineSourceMap false --removeComments true",
+    "build:assets": "cpx \"./{src,tests}/**/{!(*.ts),.!(ts)}\" dist",
+    "build:ts": "tsc",
+    "clean": "rimraf dist",
+    "lint": "npm-run-all lint:*",
+    "lint:js": "eslint . --cache --ext js --ext md --ext ts --ignore-path ../../.eslintignore --report-unused-disable-directives",
+    "lint:md": "markdownlint --ignore CHANGELOG.md *.md",
+    "test": "npm run lint && npm run build && npm run test-only",
+    "test-only": "nyc ava",
+    "watch": "npm run build && npm-run-all --parallel -c watch:*",
+    "watch:assets": "npm run build:assets -- -w --no-initial",
+    "watch:test": "ava --watch",
+    "watch:ts": "npm run build:ts -- --watch"
+  },
+  "version": "1.0.0-beta.0"
+}
--- a/packages/parser-css/package.json
+++ b/packages/parser-css/package.json
@@ -49,6 +49,7 @@
   "peerDependencies": {
     "hint": "^3.0.0-beta.4"
   },
+  "private": true,
   "repository": "webhintio/hint",
   "scripts": {
     "build": "npm run clean && npm-run-all build:*",
@@ -66,5 +67,5 @@
     "watch:test": "ava --watch",
     "watch:ts": "npm run build:ts -- --watch"
   },
-  "version": "1.0.0-beta.0"
+  "version": "1.0.0"
 }
--- a/packages/parser-css/src/parser.ts
+++ b/packages/parser-css/src/parser.ts
@@ -39,7 +39,6 @@ export default class CSSParser extends Parser {
         await this.emitCSS(code, resource);
     }
 
-
     private isCSSType(element: IAsyncHTMLElement) {
         const type = element.getAttribute('type');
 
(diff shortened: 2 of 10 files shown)


packages/parser-css/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "lint:js": "eslint . --cache --ext js --ext md --ext ts --ignore-path ../../.eslintignore --report-unused-disable-directives",
+    "lint:md": "markdownlint --ignore CHANGELOG.md *.md",
+    "test": "npm run lint && npm run build && npm run test-only",
+    "test-only": "nyc ava",
+    "watch": "npm run build && npm-run-all --parallel -c watch:*",
+    "watch:assets": "npm run build:assets -- -w --no-initial",
+    "watch:test": "ava --watch",
+    "watch:ts": "npm run build:ts -- --watch"
+  },
+  "version": "1.0.0-beta.0"
```
molant: I think you need a `private: true`. Not sure about the version. @alrra?
Did you use `npm create parser` to do the scaffolding?

packages/parser-css/src/parser.ts
```diff
@@ -0,0 +1,68 @@
...
+            const styleData: StyleParse = {
+                ast,
+                code,
+                resource
+            };
+
+            await this.engine.emitAsync(`parse::${this.name}::end`, styleData);
+
+        } catch (err) {
+            console.log(`Error parsing code: ${code} - ${err}`);
```
molant: We should be using `log.error` from utils or `debug`

packages/parser-css/src/parser.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private async parseCSS(fetchEnd: FetchEnd) {
+        const code = fetchEnd.response.body.content;
+        const resource = fetchEnd.resource;
+
+        await this.emitCSS(code, resource);
+    }
+
+
```
molant: Remove extra empty line

packages/parser-css/package.json
antross: @molant Thanks. I didn't use `npm create parser`, is that documented somewhere? Also, is that the correct command? It didn't do anything when I just tried to run it. Instead I used the current parser-javascript as a base and built out from there.

Also, is parser-webpack-config the most recently created parser? It seems to be the only one with `private: true` and a different version (`0.0.1`). The rest of the existing parsers seem to match what I did here. Happy to make the changes, but wanted to double-check due to the inconsistencies.

packages/parser-css/src/parser.ts
antross: Ok. FYI parser-javascript does the same. Should we update that as well (possibly as a separate PR) or just leave it for now and fix this one?

packages/parser-css/package.json
alrra: > I think you need a private: true. Not sure about the version. @alrra?

If we don't want a beta version published, `private: true` and `v1.0.0`.

molant: @antross this is what I get when I run the command:

image

Maybe you have an old version of npm? Try with `npm init parser` just in case.

@alrra doesn't your script need to have the `private:true` initially to know it is going to be the first time it publishes something and create the changelog?

packages/parser-css/src/parser.ts
molant: Separate PR I think should be fine. Not sure if we should use `debug` or `log.error` though. I think I'm more inclined towards `debug` but don't have a really strong opinion about it. @webhintio/core ?

packages/parser-css/package.json
alrra: > doesn't your script need to have the private:true initially to know it is going to be the first time it publishes something and create the changelog?

Yes, but more importantly the first tag, and not search for previous tags.

alrra: > npm init parser just in case.

or `npm create parser`. :D

alrra: >  I didn't use `npm create parser`, is that documented somewhere? 

`create` should just be an alias of `init` (I'm using `npm` `v6.2.0`):

```
$ npm init -h

npm init [--force|-f|--yes|-y|--scope]
npm init <@scope> (same as `npx <@scope>/create`)
npm init [<@scope>/]<name> (same as `npx [<@scope>/]create-<name>`)

aliases: create, innit  <===========
```

antross: Yep, I needed to upgrade my npm. That said, running through the wizard only gave me a `.npmrc` and `LICENSE.text` file in the created folder... seems like it's not working right...

Apart from that, I think we should document this command here: https://webhint.io/docs/contributor-guide/how-to/parser/. Happy to spin up a PR for that change if you agree.

alrra: >  seems like it's not working right..

Yes, it seems to be a bug:

```
Files copied
(node:5576) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/Users/xxxxx/.npm/_npx/5572/lib/node_modules/create-parser/dist/src/templates/new-parser/script.hbs'
(node:5576) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:5576) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

> Apart from that, I think we should document this command here: https://webhint.io/docs/contributor-guide/how-to/parser/. Happy to spin up a PR for that change if you agree.

Do it!

molant: > Apart from that, I think we should document this command here: 

That site is not updated. The latest version of the documentation is in the staging environment until we have the final homepage design. The documentation for the parser is in https://sonarwhal-staging.azurewebsites.net/docs/contributor-guide/how-to/parser/

molant: > Yes, there seems to be a bug:

I fixed that bug a while ago but there's another one with one of the templates not being properly formed. I'm working on a fix.

molant: And the [fix is in master](https://github.com/webhintio/hint/commit/887ce5c086c1a0741e5702f4d7caf556aeb51fdb)

packages/parser-css/src/parser.ts
```diff
@@ -0,0 +1,68 @@
+import * as postcss from 'postcss';
+
+import * as logger from 'hint/dist/src/lib/utils/logging';
+import { IAsyncHTMLElement, ElementFound, FetchEnd, Parser } from 'hint/dist/src/lib/types';
+import { StyleParse } from './types';
+import { Engine } from 'hint/dist/src/lib/engine';
+
+const styleContentRegex: RegExp = /^<style[^>]*>([\s\S]*)<\/style>$/;
```
alrra: Something such as:

```
<style>

</style     >
         ⬆️ spaces
```

should also match, because: [`4. After the tag name, there may be one or more ASCII whitespace.`](https://html.spec.whatwg.org/#end-tags)

packages/parser-css/src/parser.ts
```diff
@@ -0,0 +1,68 @@
...
+        const code = fetchEnd.response.body.content;
+        const resource = fetchEnd.resource;
+
+        await this.emitCSS(code, resource);
+    }
+
+    private isCSSType(element: IAsyncHTMLElement) {
+        const type = element.getAttribute('type');
+
+        return !type || type === 'text/css';
```
alrra: Can you provide a link to the spec here? Thanks!

packages/parser-css/src/parser.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    private async parseCSS(fetchEnd: FetchEnd) {
+        const code = fetchEnd.response.body.content;
+        const resource = fetchEnd.resource;
+
+        await this.emitCSS(code, resource);
+    }
+
+    private isCSSType(element: IAsyncHTMLElement) {
+        const type = element.getAttribute('type');
```
alrra: Use `normalizeString(element.getAttribute('type'))` because you can have things such as: `TEXT/CSS`.

PR #1227, opened 2018-08-10
Docs: Fix typo for `npm hint`, should be `npx hint`

cheeaun:
`npm hint` doesn't work, should be `npx hint`.


alrra:
@cheeaun Good catch, thanks!

PR #1243, opened 2018-08-17
Fix: Handle base[href] tags

chrisguttandin:
## Pull request checklist

Make sure you:

I discovered a bug in the "no-broken-links" hint, which causes it to ignore a base tag with an href value when resolving relative URLs.

I put together a fix which works when testing it with a site hosted on GitHub pages. However I was not able to add any tests as I couldn't get the tests to work in the first place. I followed the guide on the website. (https://webhint.io/docs/contributor-guide/getting-started/development-environment/)

Are the tests currently broken or am I doing something wrong?

### DIFF
--- a/packages/hint-no-broken-links/src/hint.ts
+++ b/packages/hint-no-broken-links/src/hint.ts
@@ -3,7 +3,7 @@
  * uses are available online. Checks for 404, 410, 500 or 503 status
  */
 
-import * as URL from 'url';
+import { URL } from 'url';
 import { Category } from 'hint/dist/src/lib/enums/category';
 import { HintContext } from 'hint/dist/src/lib/hint-context';
 import {
@@ -146,13 +146,17 @@ export default class NoBrokenLinksHint implements IHint {
             fetchedUrls.push({ statusCode: fetchEnd.response.statusCode, url: fetchEnd.resource });
         };
 
-        const createReports = (element: IAsyncHTMLElement, urls: Array<string>, resource: string, hrefAttribute: null | string): Array<Promise<void>> => {
-            return urls.map((url) => {
-                const isRelativeUrl = (!url.startsWith('/') && URL.parse(url).hostname === null);
-                const fullUrl = (isRelativeUrl && hrefAttribute !== null) ?
-                    URL.resolve(resource, hrefAttribute + url) :
-                    URL.resolve(resource, url);
+        const createResourceUrl = async (resource: string) => {
+            const pageDOM: IAsyncHTMLDocument = context.pageDOM as IAsyncHTMLDocument;
+            const baseTags: Array<IAsyncHTMLElement> = await pageDOM.querySelectorAll('base');
+            const hrefAttribute = (baseTags.length === 0) ? null : baseTags[0].getAttribute('href');
 
+            return (hrefAttribute === null) ? new URL(resource) : new URL(hrefAttribute, new URL(resource));
+        };
+
+        const createReports = (element: IAsyncHTMLElement, urls: Array<string>, resourceUrl: URL): Array<Promise<void>> => {
+            return urls.map((url) => {
+                const fullUrl = (new URL(url, resourceUrl)).toString();
                 const fetched = getFetchedUrl(fullUrl);
 
                 if (fetched) {
@@ -178,13 +182,10 @@ export default class NoBrokenLinksHint implements IHint {
         };
 
         const validateCollectedUrls = async (event: TraverseEnd) => {
-            const { resource } = event;
-            const pageDOM: IAsyncHTMLDocument = context.pageDOM as IAsyncHTMLDocument;
-            const baseTags: Array<IAsyncHTMLElement> = await pageDOM.querySelectorAll('base');
-            const hrefAttribute = (baseTags.length === 0) ? null : baseTags[0].getAttribute('href');
+            const resourceUrl = await createResourceUrl(event.resource);
 
             const reports: Array<Promise<void>> = collectedElementsWithUrls.reduce<Promise<void>[]>((accumulatedReports, [element, urls]) => {
-                return [...accumulatedReports, ...createReports(element, urls, resource, hrefAttribute)];
+                return [...accumulatedReports, ...createReports(element, urls, resourceUrl)];
             }, []);
 
             await Promise.all(reports);
--- a/packages/hint-no-broken-links/src/hint.ts
+++ b/packages/hint-no-broken-links/src/hint.ts
@@ -146,6 +146,37 @@ export default class NoBrokenLinksHint implements IHint {
             fetchedUrls.push({ statusCode: fetchEnd.response.statusCode, url: fetchEnd.resource });
         };
 
+        const createReports = (element: IAsyncHTMLElement, urls: Array<string>, resource: string, hrefAttribute: null | string): Array<Promise<void>> => {
+            return urls.map((url) => {
+                const isRelativeUrl = (!url.startsWith('/') && URL.parse(url).hostname === null);
+                const fullUrl = (isRelativeUrl && hrefAttribute !== null) ?
+                    URL.resolve(resource, hrefAttribute + url) :
+                    URL.resolve(resource, url);
+
+                const fetched = getFetchedUrl(fullUrl);
+
+                if (fetched) {
+                    const statusIndex = brokenStatusCodes.indexOf(fetched.statusCode);
+
+                    if (statusIndex > -1) {
+                        return context.report(fullUrl, null, `Broken link found (${brokenStatusCodes[statusIndex]} response).`);
+                    }
+                } else {
+                    // An element which was not present in the fetch end results
+                    return requester
+                        .get(fullUrl)
+                        .then((value: NetworkData) => {
+                            return handleSuccess(value, fullUrl, element);
+                        })
+                        .catch((error: any) => {
+                            return handleRejection(error, fullUrl, element);
+                        });
+                }
+
+                return Promise.resolve();
+            });
+        };
+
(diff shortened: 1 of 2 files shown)


alrra:
> Are the tests currently broken or am I doing something wrong?

Did you run `yarn test` from the root of the project? And if so, what exactly was the error? Thanks!

chrisguttandin:
Sorry, I forgot to run `yarn` first inside of the directory of the package. Running `yarn test` works in the root directory and in the directory of the package.

I will try to add some tests for my changes.

alrra:
> Sorry, I forgot to run yarn first inside of the directory of the package

Sorry for the confusion, we need to improve those docs.

> I will try to add some tests for my changes.

@chrisguttandin Thanks!

chrisguttandin:
I refactored my changes a bit and also added two tests.

The rule does now cache all discovered elements with their urls. Once `traverse::end` is fired it will look for the first base tag and will use this (if present) to check all the cached urls. I tried to write that code in a similar way as the respective part of the `hint-meta-charset-utf-8` rule is written.

This behavior relies on the `element::*` events to always happen before the `traverse::end` event and also requires the `traverse::end` event to always fire. Is that guaranteed or should the code be written in way which accounts for a different order of events.

molant:
Thanks for your PR @chrisguttandin ! I'll review it today.

> This behavior relies on the `element::*` events to always happen before the `traverse::end` event and also requires the traverse::end event to always fire. Is that guaranteed or should the code be written in way which accounts for a different order of events.

If there's an HTML then that's the right order of events so this should be fine.

packages/hint-no-broken-links/src/hint.ts
```diff
@@ -169,15 +146,56 @@ export default class NoBrokenLinksHint implements IHint {
...
-        context.on('element::object', validateElementSrcs);
+        const validateCollectedUrls = async (event: TraverseEnd) => {
+            const { resource } = event;
+            const pageDOM: IAsyncHTMLDocument = context.pageDOM as IAsyncHTMLDocument;
+            const baseTags: Array<IAsyncHTMLElement> = await pageDOM.querySelectorAll('base');
+            const hrefAttribute = (baseTags.length === 0) ? null : baseTags[0].getAttribute('href');
+
+            const reports: Array<Promise<void>> = collectedElementsWithUrls.reduce<Promise<void>[]>((accumulatedReports, [element, urls]) => {
+                return [...accumulatedReports, ...urls.map(async (url) => {
+                    const isRelativeUrl = (!url.startsWith('/') && URL.parse(url).hostname === null);
```
molant: I believe this is part of the [Legacy URL API](https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_legacy_url_api). It seems we didn't catch this on the original PR ☹

packages/hint-no-broken-links/src/hint.ts
```diff
@@ -169,15 +146,56 @@ export default class NoBrokenLinksHint implements IHint {
...
-        context.on('element::track', validateElementSrcs);
-        context.on('element::object', validateElementSrcs);
+        const validateCollectedUrls = async (event: TraverseEnd) => {
+            const { resource } = event;
+            const pageDOM: IAsyncHTMLDocument = context.pageDOM as IAsyncHTMLDocument;
+            const baseTags: Array<IAsyncHTMLElement> = await pageDOM.querySelectorAll('base');
+            const hrefAttribute = (baseTags.length === 0) ? null : baseTags[0].getAttribute('href');
+
+            const reports: Array<Promise<void>> = collectedElementsWithUrls.reduce<Promise<void>[]>((accumulatedReports, [element, urls]) => {
+                return [...accumulatedReports, ...urls.map(async (url) => {
```
molant: Is there any particular reason of using spreading instead of `concat`? Stylistic is a good answer btw.

packages/hint-no-broken-links/src/hint.ts
```diff
@@ -169,15 +146,56 @@ export default class NoBrokenLinksHint implements IHint {
...
+
+                    if (fetched) {
+                        const statusIndex = brokenStatusCodes.indexOf(fetched.statusCode);
+
+                        if (statusIndex > -1) {
+                            return context.report(fullUrl, null, `Broken link found (${brokenStatusCodes[statusIndex]} response).`);
+                        }
+                    } else {
+                        // An element which was not present in the fetch end results
+                        return await requester
```
molant: What do you think about removing the `async/await` from this function. I know it's in the original code but I think it can be safely removed.

packages/hint-no-broken-links/src/hint.ts
```diff
@@ -42,6 +42,9 @@ export default class NoBrokenLinksHint implements IHint {
         const requester = new Requester(options);
         const brokenStatusCodes = [404, 410, 500, 503];
 
+        /** Stores the elements with their urls which have been collected while traversing the page. */
```
molant: It should be `URLs` (you can see @alrra complaining about this in virtually any PR we do 😅). Can you please update it in here and any other comment instance in this file? I think our policy for `URL` in variables isn't clear. @alrra, do you want to update the code? Things like `fetchedUrls`, `urls` (eslint will complain about upper case and constructors)?

packages/hint-no-broken-links/tests/tests.ts
```diff
@@ -71,6 +75,20 @@ const tests: Array<HintTest> = [
...
     },
+    {
+        name: `This test should pass as it has links with valid href values and a base tag which gets not used`,
+        serverConfig: {
+            '/': {content: generateHTMLPage('<base href="nested/">', bodyWithValidLinks)},
+            '/about': {content: 'My about page content'}
+        }
+    },
+    {
+        name: `This test should pass as it has a link a valid link (when resolved with the base tag)`,
```
molant: I think the name for this test is not correct. Can you please update it?

packages/hint-no-broken-links/tests/tests.ts
```diff
@@ -71,6 +75,20 @@ const tests: Array<HintTest> = [
...
+        serverConfig: {
+            '/': {content: generateHTMLPage('<base href="nested/">', bodyWithValidRelativeLink)},
+            '/nested/about': {content: 'My about page content'}
+        }
+    },
```
molant: Maybe add a test that fails? Something like:

```js
{
  name: `This test should fail because resolved URL with base doesn't exist `,
  reports: [{message: `Broken link found (404 response).` }],
  serverConfig: {
    '/': {content: generateHTMLPage('<base href="nested/">', bodyWithValidRelativeLink)},
    '/nested/about': {status: 404}
  }
}
```

I haven't tested the code above so there might be some modifications to do. There's more information in the [contributor guide](https://webhint.io/docs/contributor-guide/how-to/test-rules/). Also feel free to change the name of the test 😊

packages/hint-no-broken-links/src/hint.ts
alrra: Maybe start using `somethingURL` / `urlSomething`?

packages/hint-no-broken-links/src/hint.ts
sarvaje: IMO `urls.map` should be move to its own variable/method and then `return [...accumulatedReports, ...reports]` or even `return [...accumulatedReport, ...getReports(urls)]`.
I think it would be more readable.

packages/hint-no-broken-links/src/hint.ts
chrisguttandin: I changed the code now to use the new WHATWG URL API instead. I initially thought I should better keep it in the way it was written before.

packages/hint-no-broken-links/src/hint.ts
chrisguttandin: That was just a stylistic decision. Let me know if you want me to change it to use `concat()` instead.

packages/hint-no-broken-links/src/hint.ts
chrisguttandin: Yes, you're right. There seems to be no reason for awaiting that promise.

packages/hint-no-broken-links/src/hint.ts
chrisguttandin: I'm a bit confused. Should I change the spelling of "URL" only in the comments or should I rename all variables?

alrra: > I change the spelling of "URL" only in the comments or should I rename all variables?

Both would be great! Thanks!

chrisguttandin: Ok, no problem. And what should be the new name of the variable called `url` and `urls`? Should they be renamed to something like `uRL`?

alrra: > variable called `url` and `urls`

@chrisguttandin They should remain the same. Thanks again!

packages/hint-no-broken-links/src/hint.ts
molant: And you were right on doing that. On the original PR for this rule we didn't realize we were using the old API. Sorry about that and thanks for updating it!

packages/hint-no-broken-links/src/hint.ts
molant: Maybe just do what @sarvaje proposes of extracting `urls.map` to its own variable to make it a bit clearer 😊

alrra:
@chrisguttandin Fix is now available in [`@hint/hint-no-broken-links v1.0.3`](https://www.npmjs.com/package/@hint/hint-no-broken-links). 🎉

chrisguttandin:
@alrra Thanks a lot.

PR #1254, opened 2018-08-22
New: Add extension for VSCode

antross:
## Pull request checklist

Make sure you:

- [x] Signed the [Contributor License Agreement](https://cla.js.foundation/webhintio/hint)
- [x] Followed the [commit message guidelines](https://webhint.io/docs/contributor-guide/getting-started/pull-requests/#commit-messages)

For non-trivial changes, please make sure you also:

- [x] Added/Updated related documentation.
- [x] Added/Updated related tests.

## Short description of the change(s)

Fixes #1011.

Key pieces are in `extension-vscode/src/extension.ts` and `extension-vscode/src/server.ts`.

Installing the test package:
1. Download and unzip [vscode-webhint-1.0.0.zip](https://github.com/webhintio/hint/files/2402773/vscode-webhint-1.0.0.zip)
2. Open Visual Studio Code
3. Go to "View" > "Extensions"
4. Click "..." in the upper-right of the panel
5. Choose "Install from VSIX..."
6. Select the unzipped `*.vsix`

Issues:
* [x] Fix `connector-local` to parse and emit events for HTML (PR #1277)
* [x] Fix `fetchContent` in `connector-local` to correctly handle `file:` URLs (PR #1311)
* [x] Fix `engine` and `connector-local` to accept file contents directly (PR #1268) 
* [x] Fix `engine` to clear internal list of messages after `executeOn` e4ab0b83bfd17ea26e71a979d8e4132ab5d48193
* [x] Use workspace copy of `hint` (if present) e8735e5bdf1fd3cc79e1a111cc4a374dde4a7791
* [x] Load `.hintrc` from the workspace (if present) 54b0d3ba085922826816d31dea5d88683901870e
* [x] Improve character/line information for some hints (e.g. `hint-manifest-app-name`) (PR #1297)
* [x] Add a JSON parser with character/line information (PR #1297)
* [ ] Link to documentation for errors (blocked on Microsoft/vscode#11847)
* [x] Refresh hints if `.hintrc` changes 7601f281fbdb486ce3cb817656cb667196d547b6
* [x] Fix bug where hints can appear in wrong file due to parallel execution a229b918bdb14767bcd85804ba7d3a13cd5a7d49

Issue #1245, opened 2018-08-17
Add documentation for the `local` connector


alrra:
There is currently no documentation in [`connectors`](https://webhint.io/docs/user-guide/concepts/connectors/) about the [`local`](https://github.com/webhintio/hint/tree/master/packages/connector-local) connector.

PR #1268, opened 2018-08-24
New: Pass content to `engine.executeOn`

antross:
## Pull request checklist

Make sure you:

Needed for #1254 to support validating files with unsaved changes.
Includes enhancing the local connector to use passed content.
Implemented via an options object to simplify consumption/maintenance.
New tests added for both `Engine` and `LocalConnector` to validate behavior.

### DIFF
--- a/packages/connector-local/src/connector.ts
+++ b/packages/connector-local/src/connector.ts
@@ -31,6 +31,7 @@ import * as logger from 'hint/dist/src/lib/utils/logging';
 
 import {
     IConnector,
+    IFetchOptions,
     Event, FetchEnd, ScanEnd, NetworkData
 } from 'hint/dist/src/lib/types';
 import { Engine } from 'hint/dist/src/lib/engine';
@@ -84,7 +85,7 @@ export default class LocalConnector implements IConnector {
         return [pattern];
     }
 
-    private async fetch(target: string, source?: string) {
+    private async fetch(target: string, options?: IFetchOptions) {
         /*
          * target can have one of these forms:
          *   - /path/to/file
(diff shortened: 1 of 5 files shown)


antross:
Also, do we want to add documentation for the local connector (#1245) as part of this PR, or leave it for a separate one?

alrra:
>  add documentation for the local connector (#1245) as part of this PR, or leave it for a separate one?

We can do it here in a separate commit, or in another PR. Whatever works best for you, thanks!

antross:
Ok, I'll leave the local connector documentation for another PR (likely once we've fixed it to parse HTML).

packages/connector-local/src/connector.ts
```diff
@@ -97,7 +98,7 @@ export default class LocalConnector implements IConnector {
          */
         const uri: url.URL = getAsUri(target);
         const filePath: string = asPathString(uri);
-        const content: NetworkData = await this.fetchContent(filePath);
+        const content: NetworkData = await this.fetchContent(filePath, options);
```
sarvaje: `await this.fetchContent(filePath, null, options);`

otherwise, when you check in line # 253 `options.content` is going to be always null.

packages/connector-local/src/connector.ts
```diff
@@ -300,9 +301,16 @@ export default class LocalConnector implements IConnector {
                 dot: true,
                 gitignore: true
             } as any));
+
+            // Ignore options.content when matching multiple files
+            if (options && options.content) {
+                delete options.content;
```
sarvaje: why don't set just to null?

packages/connector-local/src/connector.ts
antross: Good catch, thanks!

packages/connector-local/src/connector.ts
antross: No reason other than preference. I'll change this to set `content` to `null` instead.

packages/hint/tests/lib/engine.ts
```diff
@@ -874,3 +874,46 @@ test.serial('executeOn should return all messages', async (t) => {
 
     t.is(result.length, 2);
 });
+
+test.serial('executeOn should forward content if provided', async (t) => {
```
alrra: We started using `'` around coding related things such as `'executeOn'`, so let's use that in the future. 

I'll make this change before merging

PR #1277, opened 2018-08-28
New: Add a standalone `parser-html`

antross:
## Pull request checklist

Make sure you:

For use by `connector-local` in support of #1254 so we can expand the 
set of hints which run in the VSCode extension.

Uses `jsdom` (with scripts and resource loading disabled) to maximize
support for existing hints (e.g. `hint-axe`).


packages/connector-local/src/connector.ts
```diff
@@ -242,6 +245,17 @@ export default class LocalConnector implements IConnector {
         });
     }
 
+    private watchForWindow() {
```
sarvaje: This will work if there is more than one html file?

packages/hint/src/lib/types/jsdom-async-html.ts
```diff
@@ -1,4 +1,5 @@
-import { IAsyncHTMLDocument, IAsyncHTMLElement } from './async-html'; //eslint-disable-line
+import { IAsyncHTMLDocument, IAsyncHTMLElement, IAsyncWindow } from './async-html'; //eslint-disable-line
```
sarvaje: What does it happen if you remove `//eslint-disable-line`?

packages/parser-html/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "lint:js": "eslint . --cache --ext js --ext md --ext ts --ignore-path ../../.eslintignore --report-unused-disable-directives",
+    "lint:md": "markdownlint --ignore CHANGELOG.md *.md",
+    "test": "npm run lint && npm run build && nyc ava",
+    "init": "npm install && npm run build",
+    "watch": "npm run build && npm-run-all --parallel -c watch:*",
+    "watch:assets": "npm run build:assets -- -w --no-initial",
+    "watch:test": "ava --watch",
+    "watch:ts": "npm run build:ts -- --watch"
+  },
+  "version": "0.0.1"
```
alrra: `1.0.0`

packages/connector-local/src/connector.ts
antross: This will only populate the connector for the target HTML file as the connector APIs are tied to that singleton as far as I could tell.

packages/connector-local/package.json
```diff
@@ -7,6 +7,7 @@
     "timeout": "1m"
   },
   "dependencies": {
+    "@hint/parser-html": "^0.0.1",
```
alrra: Note to self: Move this into a different commit.

packages/hint/src/lib/types/jsdom-async-html.ts
antross: Nothing it seems. Want me to remove it? I was originally only making necessary changes (as this comment was already in the file), but it's an easy change so I can include the removal in this PR.

sarvaje: Yes please. I saw it wasn't your change, but it would be nice to remove it :)

packages/connector-local/package.json
antross: Honestly that should be a dev dependency as it's only used for types (specifically the `HTMLParse` event). The only functional tie `connector-local` has to the `parser-html` is via the `parse::html::end` event. I'll fix it (though that probably doesn't change your need to add even the dev dependency in a separate commit).

alrra: I was referring a different commit, not anything about coding changing.

That is for the release process.

packages/connector-local/src/connector.ts
```diff
@@ -242,6 +245,17 @@ export default class LocalConnector implements IConnector {
...
+        this._window = null;
+
+        const setWindow = (event: HTMLParse) => {
+            this._window = event.window;
```
antross: This also needs to emit a `can-evaluate::script` event here or `hint-axe` and friends won't work. I'll update the PR for this in a bit (need to grab a bite to eat).

packages/connector-local/src/connector.ts
sarvaje: Yep, the problem will be with a glob pattern.

I think you need to touch a little bit how the connector works to support multiple HTMLs.

Right now, the fetch is in parallel, but maybe you can refactor the code to emit the `fetch::end::html` serially so you can set the properties in the connector for each html, this way, the connector will support the `parse-html` for all the HTML files. Do you know what I mean?

The idea is when the connector store all the HTMLs in an array, without emits the `fetch::end::html`. Then, iterate that array serially doing:
1. emit the `fetch::end::html`
1. in the event `parse::html` set `this._window`
1. wait until the analysis of the html finish.
1. clean `this._window`
1. go to the next html.

antross: Makes sense and doesn't look too difficult from glancing at the code. I'll see if I can incorporate this tomorrow.

antross:
Now when I'm trying to run the `connector-local` tests on my machine, they're all failing with the message:

`'Attempted to wrap default which is already wrapped'`

in

 `\hint\node_modules\sinon\lib\sinon\util\core\wrap-method.js:38`.

Any idea what might be up?

molant:
That happens if you are trying to create a spy (or similar) twice on the same method without restoring before.

Ive seen that happening before when a test fails and the `after` portion where the restore happens doesn't get executed (or it gets executed but fails for some reason before doing it). The next test will fail with that message.

antross:
FYI, I merged and tested this with the VSCode extension PR and it worked (although error positions will need a bit of work as expected). Even got `hint-axe` running after changing its `HintScope` to `any`.

molant:
> Even got `hint-axe` running after changing its `HintScope` to `any`.

🎉🎉🎉

molant:
Everything is failing because `nyc` can't be found 🤔

```text
Error: Cannot find module 'C:\projects\hint\packages\parser-html\node_modules\nyc\bin\nyc.js'
    at Function.Module._resolveFilename (module.js:547:15)
    at Function.Module._load (module.js:474:25)
    at Function.Module.runMain (module.js:693:10)
```

packages/parser-html/README.md
```diff
@@ -0,0 +1,60 @@
+# HTML parser (`@hint/parser-html`)
+
+The `HTML` parser is built on top of `[jsdom][jsdom]` so hints can
+analyze `HTML` files.
+
+Note: This parser is not needed if using a [connector][connectors]
+which connects to a browser as the browser's DOM will be used to
+generate events instead.
```
molant: The `jsdom` connector doesn't need it neither. Maybe reformulate to say it's only required by `local`?

packages/parser-html/src/parser.ts
```diff
@@ -0,0 +1,64 @@
...
+        super(engine, 'html');
+
+        engine.on('fetch::end::html', this.onFetchEndHtml.bind(this));
+    }
+
+    private async onFetchEndHtml(fetchEnd: FetchEnd) {
+        const resource = this._url = fetchEnd.response.url;
+
+        const html = fetchEnd.response.body.content;
+        const dom = new JSDOM(html, { runScripts: 'outside-only' });
```
molant: Maybe add a comment here explaining why we want `outside-only` for future us?

packages/parser-html/src/parser.ts
```diff
@@ -0,0 +1,64 @@
...
+        await this.engine.emitAsync(`traverse::down`, traverseEvent);
+
+        // Recursively traverse child elements.
+        for (let i = 0; i < element.children.length; i++) {
+            await this.traverseAndNotify(element.children[i] as HTMLElement);
+        }
+
+        await this.engine.emitAsync(`traverse::up`, traverseEvent);
+
+        return Promise.resolve();
```
molant: You are already using `await` through this method so you shouldn't need to return a `Promise`. Is TS complaining?

packages/connector-local/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "description": "hint local connector",
   "devDependencies": {
+    "@hint/parser-html": "^1.0.0",
```
alrra: Does `connector-local` not depend on this?

antross: `connector-local` only depends on receiving the `parse::html::end` event (and will work without it, just obviously without HTML support). The dev dependency is just for the `HTMLParse` type to make TypeScript happy.

packages/parser-html/package.json
```diff
@@ -0,0 +1,77 @@
...
+      "dist/tests/**/*.js"
+    ],
+    "timeout": "1m"
+  },
+  "dependencies": {
+    "jsdom": "^12.0.0"
+  },
+  "description": "webhint parser needed to analyze HTML files",
+  "devDependencies": {
+    "@types/cheerio": "^0.22.9",
```
alrra: This is no longer needed?

packages/parser-html/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "@types/debug": "0.0.30",
+    "@types/jsdom": "^11.0.6",
+    "@types/node": "8.0.14",
```
alrra: Remove types that are not necessary.

(we try to keep them to a minimum, because they often break our builds 😔)

packages/parser-html/tests/parser.ts
```diff
@@ -0,0 +1,117 @@
...
+
+test.beforeEach((t) => {
+    t.context.engine = new EventEmitter2({
+        delimiter: '::',
+        maxListeners: 0,
+        wildcard: true
+    });
+});
+
+test.serial('If fetch::end::html is received, then we should parse the code and emit a parse::html::end event', async (t) => {
```
alrra:
```
If 'fetch::end::html' is received, then the code should be parsed and the `parse::html::end` event emitted
```

packages/parser-html/package.json
alrra: Also, after you do these changes, can you run `yarn` in the root so that `yarn.lock` file gets updated. Thanks!

antross:
@alrra do you want me to rebase and resolve the `yarn.lock` conflicts?

alrra:
> do you want me to rebase and resolve the yarn.lock conflicts?

Yes, thanks!

---

Just remove the  `yarn.lock` and run `yarn`, don't do it manually.

packages/create-parser/src/shared-templates/package.hbs
```diff
@@ -78,7 +78,8 @@
     "lint:js": "eslint . --cache --ext js --ext md --ext ts --ignore-path ../../.eslintignore --report-unused-disable-directives",{{else}}
     "lint:js": "eslint . --cache --ext js --ext md --ext ts --report-unused-disable-directives",{{/if}}
     "lint:md": "markdownlint --ignore CHANGELOG.md *.md",
-    "test": "npm run lint && npm run build && nyc ava",
+    "test": "npm run lint && npm run build && nyc ava",{{#if official}}
+    "test-only": "nyc ava",{{/if}}
```
alrra: Added this in a separate commit.

alrra:
[Package has been published](https://www.npmjs.com/package/@hint/parser-html). 🎉

PR #1297, opened 2018-09-07
New: Add line/col information for JSON

### LINKED ISSUE #1282: Provide line/col numbers for hints in JSON files
Currently webhint has a number of different parsers for various JSON-based file types (e.g. `parser-manifest` and `parser-typescript-config`). However all of these use `JSON.parse` under the covers and thus don't provide any location information about parsed objects to hints for these file types.

In working on #1254, I came across [jsonc-parser](https://github.com/Microsoft/node-jsonc-parser) which provides this information (in addition to optionally supporting JSON-with-comments). I propose adding either a utility or new parser based on this library for the other parsers and hints to consume. Then updating the JSON-based hints themselves to provide this information in errors/warnings.

antross:
## Pull request checklist

Make sure you:

Fixes #1282. Also updates these packages with line/col data:

* `hint-babel-config`
* `hint-manifest-app-name`
* `hint-manifest-is-valid`
* `hint-typescript-config`
* `parser-babel-config`
* `parser-manifest`
* `parser-typescript-config`

### DIFF
--- a/packages/hint-manifest-app-name/src/hint.ts
+++ b/packages/hint-manifest-app-name/src/hint.ts
@@ -13,7 +13,7 @@ import { ucs2 } from 'punycode';
 
 import { Category } from 'hint/dist/src/lib/enums/category';
 import { IHint, HintMetadata, IJSONLocationFunction } from 'hint/dist/src/lib/types';
-import { ManifestParsed } from '@hint/parser-manifest/dist/src/types';
+import { Manifest, ManifestParsed } from '@hint/parser-manifest/dist/src/types';
 import { HintContext } from 'hint/dist/src/lib/hint-context';
 import { HintScope } from 'hint/dist/src/lib/enums/hintscope';
 
@@ -60,7 +60,7 @@ export default class ManifestAppNameHint implements IHint {
         };
 
         const validate = async (manifestParsed: ManifestParsed) => {
-            const { getLocation, parsedContent: manifest, resource } = manifestParsed;
+            const { getLocation, parsedContent: manifest, resource }: { getLocation: IJSONLocationFunction, parsedContent: Manifest, resource: string } = manifestParsed;
 
             const name = manifest.name;
 
--- a/packages/parser-manifest/src/parser.ts
+++ b/packages/parser-manifest/src/parser.ts
@@ -8,7 +8,8 @@ import {
     FetchStart,
     IJSONResult,
     NetworkData,
-    Parser
+    Parser,
+    SchemaValidationResult
 } from 'hint/dist/src/lib/types';
 
 import isHTTP from 'hint/dist/src/lib/utils/network/is-http';
@@ -171,7 +172,7 @@ export default class ManifestParser extends Parser {
          * is a valid acording to the schema.
          */
 
-        const validationResult = validate(this.schema, result.data, result.getLocation);
+        const validationResult: SchemaValidationResult = validate(this.schema, result.data, result.getLocation);
 
         if (!validationResult.valid) {
             const manifestInvalidSchemaEvent: ManifestInvalidSchema = {


antross:
@alrra, @molant Do we want to update the remaining existing JSON-based parsers and hints as part of this PR? I think only `parser-babel-config` and `parser-typescript-config` along with their associated hints are left.

molant:
> Do we want to update the remaining existing JSON-based parsers and hints as part of this PR?

@antross yes please!

packages/hint-manifest-app-name/src/hint.ts
```diff
@@ -63,7 +60,7 @@ export default class ManifestAppNameHint implements IHint {
         };
 
         const validate = async (manifestParsed: ManifestParsed) => {
-            const { parsedContent: manifest, resource }: { parsedContent: Manifest, resource: string } = manifestParsed;
+            const { getLocation, parsedContent: manifest, resource } = manifestParsed;
```
molant: Can you please add the types back? We are trying to go full strict (#576) and I think this is needed.

packages/parser-manifest/src/parser.ts
```diff
@@ -170,7 +171,7 @@ export default class ManifestParser extends Parser {
          * is a valid acording to the schema.
          */
 
-        const validationResult: SchemaValidationResult = validate(this.schema, jsonContent);
+        const validationResult = validate(this.schema, result.data, result.getLocation);
```
molant: Can you add a type here?

packages/hint-manifest-app-name/src/hint.ts
antross: Really? TypeScript fully infers those types correctly so I'm surprised that they'd be required... 

I might try a local test case first just to make sure this is necessary in full strict mode - I hate unnecessarily redundant type definitions. :)

molant: We've had some problems in the past believing we were getting one type when it was actually another, or when we changed the return type and nothing complained and then hell happen...

packages/hint-typescript-config/src/is-valid.ts
```diff
@@ -39,12 +39,15 @@ export default class TypeScriptConfigIsValid implements IHint {
...
 
         const invalidSchema = async (fetchEnd: TypeScriptConfigInvalidSchema) => {
-            const { prettifiedErrors, resource } = fetchEnd;
+            const { errors, prettifiedErrors, resource } = fetchEnd;
 
             debug(`parse::typescript-config::error::schema received`);
 
-            for (const error of prettifiedErrors) {
-                await context.report(resource, null, error);
+            for (let i = 0; i < errors.length; i++) {
```
sarvaje: Why a `for` instead of the `for.. of`? I'm just curious.

antross: I needed an index to access the corresponding items from both `prettifiedErrors` and `errors` in order to get the message text from the former and `ProblemLocation` from the latter.

packages/hint-babel-config/tests/is-valid.ts
```diff
@@ -55,7 +55,7 @@ const tests: Array<HintLocalTest> = [
         name: 'If .babelrc contains an invalid extends, it should fail',
         path: path.join(__dirname, 'fixtures', 'invalid-extends'),
         reports: [
-            { message: `Unexpected token ' in JSON at position 191` }
+            { message: `Unexpected token ' in JSON at position 202` }
```
antross: @alrra and @molant, I changed this because the position was off on my machine, but it looks like the original value was expected in CI. Upon further review, it seems this position is different when running on Windows. I suspect this is due to Windows using `\r\n` instead of `\n` for newlines as the difference is exactly the same as the number of lines before the token.

I can (1) simply put this back to the old value to pass CI, (2) update the test (and similar tests) to not depend on the exact position, or (3) update how we're loading the test fixture files so we always use `\n`, making the offsets match cross-OS. 

My preference is (3), but I'd like your opinions.

molant: Maybe you are checking out your files with `\r\n` in your machine?
What do you think of:

- Make sure those filed don't have `\r\n`
- Change your options to "Checkout as-is, commit Unix-style": `git config --global core.autocrlf input` if [I understand the guide correctly](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace) although maybe there's a way to enforce this in the project's configuration in `.gitattributes`?

I don't think that should cause any issue with VS Code and will not need any code changes on your side. We might

packages/hint-typescript-config/src/is-valid.ts
sarvaje: That make sense hehehe, I miss the part where you use the index hehehe.

antross:
Don't know why travis-ci disappeared from the checks, but the last build passed:
https://travis-ci.org/webhintio/hint/builds/426836878

molant:
@alrra do you want to take a look or are you ok?

packages/hint-babel-config/tests/is-valid.ts
antross: > Don't use eol=lf it will bite users in the future.

Ok, I'll remove it.

I'll have to use something slightly different than `os.EOL` to avoid breaking the test case for @molant who already has user settings to checkout with LF even on Windows...

I can load the actual JSON test file separately and check for `\r\n` to alter the position if that's fine with you.

molant:
> Don't use eol=lf it will bite users in the future.

Why would that be the case?


packages/hint-babel-config/tests/is-valid.ts
molant: > My recommendation is to not use `eol=lf` as it will bite users in the future.

I'm curious about this. Why do you think this can bite users in the future? All Windows editors (including notepad) support this line ending and not having this option has already bite us.

alrra: > All Windows editors (including notepad) support this line ending and not having this option has already bite us.

Ok.

antross:
>> All Windows editors (including notepad) support this line ending and not having this option has already bite us.

> Ok.

@alrra I'll take that as we're keeping `eol=lf`. Anything else you need from me on this PR?

packages/hint-babel-config/tests/is-valid.ts
alrra: >  I'll take that as we're keeping `eol=lf`

Yes.

packages/hint/src/lib/utils/schema-validator.ts
```diff
@@ -6,7 +6,7 @@ import {
     without
 } from 'lodash';
 
-import { SchemaValidationResult } from '../types';
+import { ProblemLocation, IJSONLocationFunction, ISchemaValidationError, SchemaValidationResult } from '../types';
```
alrra: `ProblemLocation` is not used.

Fixed in https://github.com/webhintio/hint/commit/a1e1e51c658e7bcbba7bb4d5db56c95c5731e494.

Issue #773, opened 2018-01-23
Connectors Configuration:"defaultProfile" option not working


ricardogaldino:
Hi, is that correct? It did not work for me.

*File ".sonarwhalrc"
```json
{
    "browserslist": [],
    "connector": {
        "name": "chrome",
        "options": {
            "waitFor": 1000,
            "defaultProfile": true
        }
    },
    "formatters": [
        "codeframe"
    ],
    "ignoredUrls": [],
    "rules": {
        "amp-validator": "off",
        "apple-touch-icons": "error",
        "axe": "error",
        "content-type": "error",
        "disown-opener": "error",
        "highest-available-document-mode": "error",
        "html-checker": "error",
        "http-cache": "error",
        "http-compression": "error",
        "image-optimization-cloudinary": "off",
        "manifest-app-name": "error",
        "manifest-exists": "error",
        "manifest-file-extension": "error",
        "manifest-is-valid": "error",
        "meta-charset-utf-8": "error",
        "meta-viewport": "error",
        "no-disallowed-headers": "error",
        "no-friendly-error-pages": "error",
        "no-html-only-headers": "error",
        "no-http-redirects": "error",
        "no-protocol-relative-urls": "error",
        "no-vulnerable-javascript-libraries": "error",
        "ssllabs": "off",
        "strict-transport-security": "error",
        "validate-set-cookie-header": "error",
        "x-content-type-options": "error"
    },
    "rulesTimeout": 120000
}
```

molant:
Hi @ricardogaldino 

Do you mind following the issue template? The `json` you've shared has a few issues (missing `}`, doesn't have a `rules` section, it isn't indented...). That will help us investigate what's going on.

Also, in order for `defaultProfile` we will need to know a bit more about what type of authentication system the site uses and how long the authentication state lasts. A link to the site will be helpful. In the meantime you can try the following:

* If the site remembers you from one session to another, make sure that you are already authenticated.
* If the site only authenticates you for a session (i.e.: when you close the browser you are logged out), then open the site, log in, and then run `sonarwhal` without closing the browser.

ricardogaldino:
Hello @molant , 
When I assign "defaultProfile": true it opens a new tab with a blank page in the same open browser where I already authenticated before.

A red "X" appears on the console and it stops.

When it is set to false it only checks the login page.

molant:
Can you please fill the template with all the requested information. It will be useful to have the debug output for this.

alrra:
[...]

PR #1305, opened 2018-09-12
Docs: Add note about `defaultProfile` and chrome launcher issue

sarvaje:
## Pull request checklist

Make sure you:

### DIFF
--- a/packages/connector-chrome/README.md
+++ b/packages/connector-chrome/README.md
@@ -39,7 +39,9 @@ The set of settings supported by Chrome connector are:
   default profile or create a new one. By default the value is `false`
   so a new one is created. You might want to set it to `true` if you
   want `webhint` to have access to pages where the default profile is
-  already authenticated.
+  already authenticated. NOTE: Due to an issue in
+  [`chrome-launcher`][chrome-launcher-issue], all the instances of
+  the browser need to be closed before run `webhint`.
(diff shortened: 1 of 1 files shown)


alrra:
@sarvaje In the future, add `Fix <issue_number>` at the end of the commit. Thanks!

sarvaje:
@alrra in this case it was intentionally, IMO, this doesn't fix the issue, just add a note in the documentation.

But you are right, most of the time I forgot to add it hehehe.

alrra:
> this doesn't fix the issue

Yes and no. Practically, until a new version of `chrome-launcher` is released, there is nothing we can do about it in `webhint` except inform users about the limitation. 

---

@webhintio/core Feel free to reopen #773 if you think it's worth having it opened.

sarvaje:
>Yes and no. Practically, until a new version of `chrome-launcher` is released, there is nothing we can do about it in `webhint` except inform users about the limitation. 

That's true :)

molant:
> until a new version of `chrome-launcher` is released

Actually is more of a limitation of chrome. You need to launch it with a special parameter to enable the debugging protocol. If you run it normally the browser doesn't listen into any port. If the browser is already opened when you use `chrome-launcher` with the option to use the default profile then it will not be able to connect to it and thus the reason of time outs.

PR #1337, opened 2018-09-24
Fix/multiple things

molant:
## Pull request checklist

Make sure you:

* Convert to text base64 text based responses
* Fix a location issue with `jsdom` depending on the version pulled
* Some coding improvements in `debugging-protocol`: some types, prefix private properties with `_` 
* Fix truncated responses in `debugging-protocol`

### DIFF
--- a/packages/hint/src/lib/types/jsdom-async-html.ts
+++ b/packages/hint/src/lib/types/jsdom-async-html.ts
@@ -63,11 +63,16 @@ export class JSDOMAsyncHTMLElement implements IAsyncHTMLElement {
     /* istanbul ignore next */
     public getLocation(): ProblemLocation {
         try {
+            /*
+             * TODO: Depending on the install (yarn vs npm) we get a different version of `jsdom`
+             * that has a different version of `parse5`. This takes care of this issue but should
+             * be fixed once we migrate completely to jsdom 12 (https://github.com/webhintio/hint/pull/1274)
+             */
             const location = this._dom && this._dom.nodeLocation(this._htmlelement);
 
             return location && {
-                column: location.startTag.col,
-                line: location.startTag.line - 1
+                column: location.startTag.col || (location.startTag as any).startCol,
+                line: (location.startTag.line || (location.startTag as any).startLine) - 1
             } || null;
         } catch (e) {
             // JSDOM throws an exception if `includeNodeLocations` wasn't set.
--- a/packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
+++ b/packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
@@ -17,6 +17,7 @@ import { promisify } from 'util';
 import * as cdp from 'chrome-remote-interface';
 import { compact, filter } from 'lodash';
 import { atob } from 'abab';
+import { Crdp } from 'chrome-remote-debug-protocol';
 
 import { CDPAsyncHTMLDocument, AsyncHTMLElement } from './cdp-async-html';
 import { getContentTypeData, getType } from 'hint/dist/src/lib/utils/content-type';
@@ -53,7 +54,7 @@ export class Connector implements IConnector {
     /** The instance of hint that is using this connector. */
     private _server: Engine;
     /** The client to talk to the browser. */
-    private _client;
+    private _client: Crdp.CrdpClient;
     /** A set of requests done by the connector to retrieve initial information more easily. */
     private _requests: Map<string, any>;
     /** Indicates if there has been an error loading the page (e.g.: it doesn't exists). */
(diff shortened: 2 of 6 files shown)


molant:
Love the PR number 😎

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -63,20 +65,22 @@ export class Connector implements IConnector {
...
     private _tabs = [];
     /** Tells if a favicon of a page has been downloaded from a link tag. */
     private _faviconLoaded: boolean = false;
     /** The amount of time before an event is going to be timedout. */
     private _timeout: number;
     /** Browser PID */
-    private pid: number;
+    private _pid: number;
     private _targetNetworkData: NetworkData;
     private launcher: ILauncher;
```
antross: Any reason this isn't `_launcher` since you've added leading underscores to the other private properties?

packages/hint/src/lib/types/jsdom-async-html.ts
```diff
@@ -63,11 +63,16 @@ export class JSDOMAsyncHTMLElement implements IAsyncHTMLElement {
     /* istanbul ignore next */
     public getLocation(): ProblemLocation {
         try {
-            const location = this._dom && this._dom.nodeLocation(this._htmlelement);
+            /*
+             * TODO: Depending on the install (yarn vs npm) we get a different version of `jsdom`
+             * that has a different version of `parse5`. This takes care of this issue but should
+             * be fixed once we migrate completely to jsdom 12 (https://github.com/webhintio/hint/pull/1274)
+             */
+            const location: any = this._dom && this._dom.nodeLocation(this._htmlelement);
```
antross: Minor nit since this should be a temporary workaround, but I'd suggest pushing the explicit use of `any` as low as possible so we still get types and intellisense for supported properties. How about an explicit cast on lines 74 and 75?

```typescript
column: location.startTag.col || (location.startTag as any).startCol,
line: (location.startTag.line || (location.startTag as any).startLine) - 1
```

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -302,7 +307,19 @@ export class Connector implements IConnector {
         }
     }
 
-    private async getResponseBody(cdpResponse): Promise<{ content: string, rawContent: Buffer, rawResponse(): Promise<Buffer> }> {
+    /** Wait until the given `requestId` request has loaded all the content. */
+    // TODO: remove `any` from return type
+    private waitForContentLoaded(requestId: string): Promise<any> {
+        if (this._finishedRequests.has(requestId)) {
+            return Promise.resolve();
```
antross: Since we add items to `_finishedRequests` before they're resolved, couldn't this incorrectly cause the calling code to think the request is complete when it's still pending? Perhaps the `Promise` created on line 317 should be included in the `Map` and returned here instead?

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
molant: Probably a leftover from when I was cherry-picking. It should have a `_`

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
molant: I think the problem is the name of the variable. `_finishedRequests` isn't the best one but couldn't find a better one. Let's see if I can explain how this works (it's a bit circular) and we can come out with a better one:

**Option 1:** Request the body after the full response is available:

1. the `onLoadingFinished` (which is fired only when the full response of a request is available) event handler adds an item to the `_finishedRequests` whose key is the `requestId`.
1. `waitingForContentLoaded` is called from `getResponseBody`. It checks if `_finishedRequests` has an item for that `requestId`. It does so it knows the content is available and can actually request it via `Network.getResponseBody`

**Option 2:** Request the body before the full response is available:

1. `waitingForContentLoaded` is called from `getResponseBody`. `_finishedRequests` doesn't have an item with the right `requestId` so it adds a promise to the value to be called when it is available.
1. `onLoadingFinished` receives the vent for the previous `requestId`. Because the item is already there, it calls `resolve` so the execution can continue in the other side

Let me know if this makes more sense. If you have a better idea on how to do it let me know. This is the first thing I could come up (and we are using a similar pattern for something similar here too). Happy to document this also in the code.

molant:
@antross I've updated the code, hopefully now it's clearer what's going on. Let me know if not!

antross:
> hopefully now it's clearer what's going on
Yes, that helped. Thanks!

PR #1341, opened 2018-09-25
Chore: Enable strict mode across packages

antross:
## Pull request checklist

Make sure you:

- [ ] ~Added/Updated related documentation.~
- [ ] ~Added/Updated related tests.~

Refs #576.

Each commit targets a package so this can be merged piece-by-piece if desired.

### DIFF
--- a/packages/connector-local/src/connector.ts
+++ b/packages/connector-local/src/connector.ts
@@ -103,7 +103,7 @@ export default class LocalConnector implements IConnector {
     }
 
     private async fetchData(target: string, options?: IFetchOptions): Promise<FetchEnd> {
-        const content: NetworkData = await this.fetchContent(target, void 0, options);
+        const content: NetworkData = await this.fetchContent(target, undefined, options);
 
         return {
             element: null,
@@ -163,7 +163,7 @@ export default class LocalConnector implements IConnector {
 
             this.watcher = chokidar.watch(target, {
                 /* istanbul ignore next */
-                cwd: !isF ? targetString : void 0,
+                cwd: !isF ? targetString : undefined,
                 ignored: ignored.concat(['.git/']),
                 ignoreInitial: true,
                 /*
@@ -326,7 +326,7 @@ export default class LocalConnector implements IConnector {
 
             // Ignore options.content when matching multiple files
             if (options && options.content) {
-                options.content = void 0;
+                options.content = undefined;
             }
         }
 
--- a/packages/connector-local/src/connector.ts
+++ b/packages/connector-local/src/connector.ts
@@ -49,12 +49,12 @@ const debug: debug.IDebugger = d(__filename);
 const defaultOptions = {};
 
 export default class LocalConnector implements IConnector {
-    private _window: IAsyncWindow = null;
+    private _window: IAsyncWindow | undefined;
     private _options: any;
     private engine: Engine;
     private _href: string = '';
     private filesPattern: Array<string>;
-    private watcher: chokidar.FSWatcher = null;
+    private watcher: chokidar.FSWatcher | null = null;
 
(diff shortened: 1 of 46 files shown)


packages/connector-local/src/connector.ts
```diff
@@ -103,7 +103,7 @@ export default class LocalConnector implements IConnector {
     }
 
     private async fetchData(target: string, options?: IFetchOptions): Promise<FetchEnd> {
-        const content: NetworkData = await this.fetchContent(target, null, options);
+        const content: NetworkData = await this.fetchContent(target, void 0, options);
```
alrra: Is there a reason why `void 0` is used?

packages/extension-vscode/src/server.ts
```diff
@@ -21,7 +21,7 @@ import * as config from 'hint/dist/src/lib/config';
 import * as loader from 'hint/dist/src/lib/utils/resource-loader'; // eslint-disable-line
 import { Problem, Severity, UserConfig } from 'hint/dist/src/lib/types';
 
-let workspace: string;
+let workspace = '';
```
alrra: Why not `let workspace: string = ''` ?

packages/connector-local/src/connector.ts
antross: The current types required this to be `undefined` as opposed to `null`. Per our ESLint rules we shouldn't use `undefined` directly since it can be overridden, so `void 0` is the recommended approach in this case. 

Alternatively we could plumb through support for `null` instead if you'd prefer.

packages/extension-vscode/src/server.ts
antross: Simply because it's redundant.

packages/connector-local/src/connector.ts
alrra: > shouldn't use undefined directly since it can be overridden

Not as of ES5. 

> Alternatively we could plumb through support for null instead if you'd prefer.

@webhintio/core Thoughts?

molant: We could disable the `undefined` rule. I think we discussed this previously.

I don't have a strong opinion about `null` or not.

antross: > We could disable the `undefined` rule

@alrra does this sound good to you? If so I'll make the change to `.eslintrc.json` in this PR and switch my `void 0` usages to `undefined` instead.

alrra: Yes, go for it.

packages/extension-vscode/src/server.ts
```diff
@@ -150,7 +155,7 @@ const loadWebHint = async (directory: string): Promise<hint.Engine> => {
      * `iltorb` if it was compiled for a different version of `node` and the
      * `local` connector doesn't support it anyway.
      */
-    userConfig.hints['http-compression'] = 'off';
+    (userConfig.hints as HintsConfigObject)['http-compression'] = 'off';
```
molant: Why do you need to do an `as HintConfigObject`?

molant: Never mind, I just checked the definition of `UserConfig`

molant:
Do we want to merge what you currently have or want to wait until everything is done? The more we wait the more the possibilities of conflicts if we add new code (although with the current tasks we are doing I think the risk is low).

alrra:
> merge what you currently have

I'm fine with doing that.

antross:
@molant if you want to merge now would be a good time. I'm knee-deep in `packages/hint` because I started hitting some hints which needed the core package itself to be more strongly typed.

molant:
@alrra @antross let's merge it then!

PR #1373, opened 2018-10-05
Chore/enable strict mode

molant:
## Pull request checklist

Make sure you:

- [x] Signed the [Contributor License Agreement](https://cla.js.foundation/webhintio/hint)
- [x] Followed the [commit message guidelines](https://webhint.io/docs/contributor-guide/getting-started/pull-requests/#commit-messages)

For non-trivial changes, please make sure you also:

- ~~Added/Updated related documentation.~~
- [x] Added/Updated related tests.

## Short description of the change(s)

Enable `strict` mode in several packages (`hints`, `create-`).

PR #1372, opened 2018-10-05
Chore: Enable strict mode in remaining packages

antross:
## Pull request checklist

Make sure you:

- [x] Signed the [Contributor License Agreement](https://cla.js.foundation/webhintio/hint)
- [x] Followed the [commit message guidelines](https://webhint.io/docs/contributor-guide/getting-started/pull-requests/#commit-messages)

For non-trivial changes, please make sure you also:

- ~Added/Updated related documentation.~
- [x] Added/Updated related tests.

## Short description of the change(s)

Refs #576 and covers packages not included in #1371.

PR #1382, opened 2018-10-08
Chore: Fully enable strict mode for the project

### LINKED ISSUE #576: Enable strict mode of TypeScript

antross:
## Pull request checklist

Make sure you:

Fixes #576.

Combines PRs #1372, #1373, and #1377.


packages/hint-typescript-config/tests/import-helpers.ts
```diff
@@ -1,20 +1,23 @@
 import * as path from 'path';
 
+import { GenericTestContext, Context } from 'ava';
 import * as sinon from 'sinon';
 
 import { getHintPath } from 'hint/dist/src/lib/utils/hint-helpers';
 import * as hintRunner from '@hint/utils-tests-helpers/dist/src/hint-runner';
 import { HintLocalTest } from '@hint/utils-tests-helpers/dist/src/hint-test-type';
+
+
```
sarvaje: remove extra line.

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -706,13 +707,13 @@ export class Connector implements IConnector {
...
-            Network.clearBrowserCache(),
-            Network.setCacheDisabled({ cacheDisabled: true }),
+            Network.clearBrowserCache!(),
+            Network.setCacheDisabled!({ cacheDisabled: true }),
             // The typings we use for CDP aren't 100% compatible with our libarary
-            Network['requestWillBeSent'](this.onRequestWillBeSent.bind(this)), // eslint-disable-line dot-notation
-            Network['responseReceived'](this.onResponseReceived.bind(this)), // eslint-disable-line dot-notation
-            Network['loadingFinished'](this.onLoadingFinished.bind(this)), // eslint-disable-line dot-notation
-            Network['loadingFailed'](this.onLoadingFailed.bind(this)) // eslint-disable-line dot-notation
+            (Network as any)['requestWillBeSent'](this.onRequestWillBeSent.bind(this)), // eslint-disable-line dot-notation
```
sarvaje: Why are we using ` (Network as any)['requestWillBeSent'](this.onRequestWillBeSent.bind(this))` instead of `Network.requestWillBeSent`?

packages/hint-x-content-type-options/src/hint.ts
```diff
@@ -78,7 +79,7 @@ export default class XContentTypeOptionsHint implements IHint {
                 return;
             }
 
-            const headerValue: string = normalizeString(response.headers && response.headers['x-content-type-options']);
+            const headerValue: string = normalizeString(response.headers && response.headers['x-content-type-options']) || '';
```
antross: This needs to be `headerValue: string | null` instead of falling back to the empty string to avoid altering behavior (note the explicit check for `if (headerValue === null)` a few lines later. I'll update the PR shortly.

scripts/release.ts
```diff
@@ -44,6 +46,35 @@ type GitHub = {
     password?: string;
 };
 
+interface ITaskContext {
```
sarvaje: can we use `type` instead of `interface`?. IIRC, we are using `interface`s just when a class implemented it.

Same with `interface ITask`

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
molant: Because the types we have aren't 100% compatible with the library we are using for CDP. The types provided by the chrome team use `onRequestWillBeSent` but [`chrome-remote-interface`](https://github.com/cyrus-and/chrome-remote-interface) has them as `requestWillbeSent`.

sarvaje: ok!

packages/connector-jsdom/src/resource-loader.ts
```diff
@@ -19,14 +19,14 @@ export default class CustomResourceLoader extends ResourceLoader {
         this._connector = connector;
     }
 
-    public fetch(url: string, options: { element: HTMLElement }) {
+    public async fetch(url: string, options: { element: HTMLElement }): Promise<Buffer | null> {
```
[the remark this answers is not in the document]
antross: It cleans up the `return null` further down in the function to actually return a `Promise<null>`  which was my interpretation of the intent (please correct me if I'm wrong):
https://github.com/webhintio/hint/blob/482a09d557f28e22853a37603f77265443bb0a96/packages/connector-jsdom/src/resource-loader.ts#L43

I could have explicitly returned a new `Promise`, but this felt like a more complete fix to avoid having an unwrapped `null` returned from the function accidentally in the future.

scripts/release.ts
antross: Sure. Any particular reason for the preference?

packages/connector-jsdom/src/resource-loader.ts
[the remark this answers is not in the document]
sarvaje: ohhh, I see, you are rigth, that was an error.
Anyway, you don't need to return a new promise you can just `return Promise.resolve()`

scripts/release.ts
sarvaje: We were mixing `type`s and `interface`s all over the code and we didn't have any criteria to use one or another, so, because interface and class in other languages are related we agree to use `type` if there is no class that implement the interface.

PR #1389, opened 2018-10-09
Build: Use TypeScript project references

antross:
## Pull request checklist

Make sure you:

Reduce build time for the project by ~50% and automatically
build dependencies for sub-packages. E.g. building
`connector-local` will build both `hint` and `parser-html` if needed.

The build order of sub-packages no longer needs to be specified in
the root `package.json` as `tsc -b` will use the provided references
to automatically determine the build order instead.

A new script, `update-tsconfig-references.ts` is included to create
references in each package's `tsconfig.json` file based on
dependencies in the associated `package.json`. This can be built
and run from the project root using `yarn update:references`, but
only needs to be run when new packages are added or cross-
package dependencies change.

Issues:
Currently `build:assets` which copies static dependences doesn't
get run for references when building a sub-package (as this is
done by `cpx`, not `tsc`). This can be worked around by running
`yarn build:assets` once from the root of the project, then
running `yarn build` in the sub-package, which is still much
faster than a full build.


packages/connector-edge/tsconfig.json
```diff
@@ -11,5 +11,8 @@
     "include": [
         "src/**/*.ts",
         "tests/**/*.ts"
+    ],
+    "references": [
+        { "path": "../hint" }
```
sarvaje: I thinks that `utils-debugging-protocol-common` is also needed here.

antross: It's not listed in `dependencies` or `devDependencies` of `packages/connector-edge/package.json`.

Is there a way it can be included that I'm missing?

scripts/update-tsconfig-references.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    writeContent(content).to(filePath);
+};
+
+const main = () => {
+
+    // Include all sub-packages (except configurations and connector-edge)
+    const subPackages = shell.ls('packages')
+        .filter((packageName: string) => {
+            return !packageName.startsWith('configuration-') && packageName !== 'connector-edge';
```
sarvaje: Why are you ignoring `connector-edge`?

antross: Because it only builds on Windows (and even then we don't typically build it as part of the full build when doing local development). It still builds individually.

packages/connector-edge/tsconfig.json
sarvaje: > It's not listed in dependencies or devDependencies 

Then that is an issue, it should be hehehe.

scripts/update-tsconfig-references.ts
sarvaje: ok! :)

sarvaje:
> Reduce build time for the project by ~50%

🎉🎈🎊🎉🎈🎊💃💃💃

packages/connector-edge/tsconfig.json
sarvaje: @antross can you add it to the package.json, it should be a dependency, like in `chrome-connector`

antross: Will do.

antross:
In the future we may get to drop explicitly creating project references. The TypeScript team is discussing automatically figuring out monorepo internal dependencies by looking at package.json and/or symlink structure: https://github.com/Microsoft/TypeScript/issues/25376

molant:
I quickly synced with @antross this morning and on top of rebasing and fix conflicts he's going to look also into the `peerDependencies` field when generating the references.

I've added the `workInProgress` label so this doesn't get merged by mistake.

antross:
@molant this has been rebased and now includes `peerDependencies` so removing the `work-in-progress` label. 

On a side note, did someone recently re-write history in `master`? It looked like a dedicated commit to update `lodash` got either replaced or merged with a different commit when I rebased (which appears to be the reason for the merge conflict in the first place).

package.json
```diff
@@ -46,27 +46,12 @@
...
-    "build:connector-local": "cd packages/connector-local && yarn build",
-    "build:connectors": "yarn build:connector-chrome && yarn build:connector-jsdom && yarn build:connector-local",
-    "build:hint": "cd packages/hint && yarn build",
-    "build:parser:babel": "cd packages/parser-babel-config && yarn build",
-    "build:parser:html": "cd packages/parser-html && yarn build",
-    "build:parser:js": "cd packages/parser-javascript && yarn build",
-    "build:parser:manifest": "cd packages/parser-manifest && yarn build",
-    "build:parser:ts": "cd packages/parser-typescript-config && yarn build",
-    "build:parser:webpack": "cd packages/parser-webpack-config && yarn build",
+    "build": "yarn clean && yarn build:assets && yarn build:ts",
```
alrra: Maybe this should be `yarn clean && yarn update:references && ...`  to ensure that everything is up-to-date?

antross: Sounds good to me.

molant:
🎉

PR #1399, opened 2018-10-15
Feature/hint valid doctype

CKGrafico:
## Pull request checklist

Make sure you:

Pull Request for https://github.com/webhintio/rfcs/issues/33


packages/hint-valid-doctype/src/hint.ts
```diff
@@ -0,0 +1,114 @@
...
+            description: `Hint to validate if the doctype is correct`
+        },
+        id: 'valid-doctype',
+        schema: [],
+        scope: HintScope.any
+    }
+
+    public constructor(context: HintContext) {
+        const validDoctype = '<!doctype html>';
+        const doctypeRegex = /(<!doctype ).+(>)(.+)?/;
```
sarvaje: This variable is not used anywhere.

packages/hint-valid-doctype/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+    "compilerOptions": {
+        "outDir": "dist"
```
sarvaje: add the `strict: true`

sarvaje:
@CKGrafico there is also few eslint errors: https://travis-ci.org/webhintio/hint/jobs/441776934

packages/hint-valid-doctype/tsconfig.json
molant: See https://github.com/webhintio/hint/issues/1400

packages/hint-valid-doctype/src/hint.ts
```diff
@@ -0,0 +1,114 @@
...
+            const matched = content.match(doctypeRegexFactory('gi'));
+
+            if (!matched || matched.length < 1) {
+                await context.report(resource, element, `The file does not contain a doctype tag`);
+                return;
+            }
+
+            const [contentDoctype] = matched;
+
+            if (contentDoctype.toLowerCase() !== validDoctype) {
```
alrra: `<!doctype   html   >` is also valid and should be allowed.

(see: https://html.spec.whatwg.org/multipage/syntax.html#the-doctype).

alrra: We don't do that for other hints. If the value is valid and the intended one, we don't complain (e.g.: [1](https://github.com/webhintio/hint/blob/e7887d4068f20caf8a55b534a3d1e0e72e2f5c1f/packages/hint-meta-charset-utf-8/src/hint.ts#L108)).

CKGrafico:
We are implementing the feedback, thanks!

PS: We have different result of eslint in our local env than in travis, we are investigating why.

packages/hint-doctype/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "files": [
+      "dist/tests/**/*.js"
+    ],
+    "timeout": "1m"
+  },
+  "description": "hint to validate if the doctype is correct",
+  "devDependencies": {
+    "@hint/utils-tests-helpers": "^1.0.0",
+    "@types/debug": "0.0.30",
+    "@types/node": "8.0.14",
```
molant: Now that we are in strict mode we can update this version. Can you make it match the one we have in the rest of the project? I don't think it's the latest one though.

packages/hint-doctype/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "cpx": "^1.5.0",
+    "eslint": "^5.5.0",
+    "eslint-plugin-markdown": "^1.0.0-beta.7",
+    "eslint-plugin-typescript": "^0.12.0",
+    "markdownlint-cli": "^0.13.0",
+    "npm-link-check": "^2.0.0",
+    "npm-run-all": "^4.1.2",
+    "nyc": "^13.0.1",
+    "rimraf": "^2.6.2",
+    "hint": "^3.3.2",
```
molant: Update to the latest versions of all packages. E.g.: hint is 3.4.10, TypeScript 3.1.1, etc.

packages/hint-doctype/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "doctype-hint"
+  ],
+  "license": "Apache-2.0",
+  "main": "dist/src/hint.js",
+  "name": "@hint/hint-doctype",
+  "nyc": {
+    "extends": "../../.nycrc"
+  },
+  "peerDependencies": {
+   "hint": "^3.3.2"
```
molant: Don't forget to update to the latest version

packages/hint-doctype/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "lint:md": "markdownlint --ignore CHANGELOG.md *.md",
+    "test": "npm run lint && npm run build && npm run test-only",
+    "test-only": "nyc ava",
+    "init": "npm install && npm run build",
+    "watch": "npm run build && npm-run-all --parallel -c watch:*",
+    "watch:assets": "npm run build:assets -- -w --no-initial",
+    "watch:test": "ava --watch",
+    "watch:ts": "npm run build:ts -- --watch"
+  },
+  "version": "0.0.1"
```
molant: Version 1.0.0

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+            return new RegExp(/(<!doctype\s(html)\s*?)(>)(.+)?/, flags ? flags : 'g');
+        };
+
+        const checkDoctypeIsValid = async (resource: string, element: IAsyncHTMLElement | null, content: string): Promise<boolean> => {
+            debug(`Checking if the doctype is valid.`);
+
+            const matched = content.match(doctypeRegexFactory('gi'));
+
+            if (!matched || matched.length < 1) {
+                await context.report(resource, element, `The file does not contain a valid doctype tag`);
```
molant: > The resource does not contain a valid DOCTYPE.

packages/hint-doctype/package.json
```diff
@@ -0,0 +1,70 @@
...
+  "peerDependencies": {
+   "hint": "^3.3.2"
+  },
+  "private": true,
+  "repository": "webhintio/hint",
+  "scripts": {
+    "build": "npm run clean && npm-run-all build:*",
+    "build-release": "npm run clean && npm run build:assets && tsc --inlineSourceMap false --removeComments true",
+    "build:assets": "cpx \"./{src,tests}/**/{!(*.ts),.!(ts)}\" dist",
+    "build:ts": "tsc",
```
sarvaje: @CKGrafico can you double check these scripts with the one in any other hint?

at least `build:ts` has to be `tsc -b`.

This is something we need probably change in the templates.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        const checkDoctypeFirstLine = async (resource: string, element: IAsyncHTMLElement | null, content: string): Promise<void> => {
+            debug(`Checking if the doctype is in the first line.`);
+
+            const firstLine = content.split(/\r|\n/)[0];
+            const matched = firstLine.match(doctypeRegexFactory('gi'));
+
+
+            if (!matched || matched.length < 1) {
+                await context.report(resource, element, `The first line does not contain a valid doctype tag.`);
```
molant: > DOCTYPE is not in the first line.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+        },
+        id: 'doctype',
+        schema: [],
+        scope: HintScope.any
+    }
+
+    public constructor(context: HintContext) {
+
+        const doctypeRegexFactory = (flags?: string) => {
+            return new RegExp(/(<!doctype\s(html)\s*?)(>)(.+)?/, flags ? flags : 'g');
```
sarvaje: it should be `/(<!doctype\s+(html)\s*?)(>)(.+)?/`

packages/hint-doctype/tests/tests.ts
```diff
@@ -0,0 +1,109 @@
+import { getHintPath } from 'hint/dist/src/lib/utils/hint-helpers';
+import { HintTest } from '@hint/utils-tests-helpers/dist/src/hint-test-type';
+import * as hintRunner from '@hint/utils-tests-helpers/dist/src/hint-runner';
+
+const hintPath = getHintPath(__filename);
+
+const tests: Array<HintTest> = [
```
sarvaje: Add some test with multiple whitespaces:
`<doctype  html>` `doctype html >`

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+                await context.report(resource, element, `The first line does not contain a valid doctype tag.`);
+
+                return;
+            }
+
+            if (matched) {
+                const cleaned = matched[0].trim().split('');
+
+                if (cleaned[cleaned.length-1] !== '>') {
+                    await context.report(resource, element, `There is additional information on the line with the doctype tag`);
```
molant: What is this testing exactly? Can you put a comment with an example of something that will fail?

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+            }
+        };
+
+        const checkDoctypeLowercase = async (resource: string, element: IAsyncHTMLElement | null, content: string): Promise<void> => {
+            debug(`Checking that the doctype is in lowercase`);
+
+            const matched = content.match(doctypeRegexFactory());
+
+            if (!matched) {
+                await context.report(resource, element, `The doctype should be in lowercase`);
```
molant: > DOCTYPE should be lowercase.

Also we should talk about enforcing this or not. The reason why we say lowercase is for (marginally) better compression but even [the spec uses `<!DOCTYPE html>`](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype). I believe this is going to cause more noise than anything else so maybe we should just ignore the casing? @webhintio/contributors thoughts?

Also the spec mentions another valid DOCTYPE: `<!DOCTYPE html SYSTEM "about:legacy-compat">`. That I don't believe it's covered in here or the tests.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+
+                return;
+            }
+        };
+
+        const checkNoDuplicateDoctype = async (resource: string, element: IAsyncHTMLElement | null, content: string): Promise<void> => {
+            debug(`Checking that there is only one doctype tag in the document`);
```
molant: Missing `.` at the end.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
+/**
```
molant: Some comments for the messages for `debug` and `report`:

* Use DOCTYPE instead of doctype, that's how the spec mentions it
* Make sure there's always an ending `.`. You've missed a couple.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+
+                return;
+            }
+        };
+
+        const onFetchEndHTML = async (fetchEnd: FetchEnd): Promise<void> => {
+            const { resource, element, response } = fetchEnd;
+
+            if (!response || !response.body || !response.body.content) {
+                await context.report(resource, element, 'Content has no body');
```
molant: > Resource has no content.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+
+                return;
+            }
+
+            const { body } = response;
+            const { content } = body;
+
+            /*
+             * If doctype is not valid, don not run more tests
```
molant: Nit: You can collapse this into one line comment: `// If DOCTYPE is not valid, do not run more tests` . Also you have a typo in
> do~~n~~ not

packages/hint-doctype/package.json
CKGrafico: I think that the newest version that I've seen in other packages is typescript@3.0.3 what is better? Update our packages to the latest version as others have in the project, or update to the latest version using some tools like `ncu` ?

molant: @greenkeeper[bot] should have updated those 🤔
I think it's best if you use the latest version in this package and we can take care of updating in a separate PR.

packages/hint-doctype/package.json
CKGrafico: Ok thanks, I think that the template that I've used was not updated. Now I'm using `hint-minified-js` as the reference.

packages/hint-doctype/tests/tests.ts
CKGrafico: Yes!

packages/hint-doctype/package.json
CKGrafico: Ok, I'm using as reference `hint-minified-js`

packages/hint-doctype/package.json
CKGrafico: Ups, thanks!

packages/hint-doctype/src/hint.ts
CKGrafico: Yes!

packages/hint-doctype/src/hint.ts
antross: I agree we should ignore the casing. Although webhint is configurable, we should be careful about flagging issues where there's limited benefit in a change. Developers should be able to trust that webhint identifies *important* issues in their code.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,125 @@
...
+
+            if (matched && matched.length > 1) {
+                await context.report(resource, element, `There is more than one doctype tag in the document`);
+
+                return;
+            }
+        };
+
+        const onFetchEndHTML = async (fetchEnd: FetchEnd): Promise<void> => {
+            const { resource, element, response } = fetchEnd;
```
antross: The `element` here will be the one that triggered the request, not the `doctype` in this `resource`. We'll need to either search for the actual node for the `doctype` to pass it on to `context.report` (which I'm not sure is possible in all our connectors) or explicitly pass a `ProblemLocation` to `context.report` based on where the issue was found.

molant: > (which I'm not sure is possible in all our connectors) or explicitly pass a `ProblemLocation` to `context.report` based on where the issue was found.

Probably calculating the `ProblemLocation` is going to be the more robust solution assuming the doctype is going to be always near the top...

packages/hint-doctype/src/hint.ts
molant: The documentation needs to be updated too.

packages/hint-doctype/src/hint.ts
molant: The `DOCTYPE` comment from above also applies to the `README.md`

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,72 @@
+# Check that the Doctype tag of the document is valid
+
+This hint checks if there is a valid Doctype in the document.
```
molant: What do you think of:
> This hint checks if the HTML is using the most modern DOCTYPE.

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Why is this important?
+
+A valid doctype will give the browser the correct information
```
molant: I like MDN's explanation. Maybe we can just put it as a comment and link the source:

> In HTML, the doctype is the required "`<!DOCTYPE html>`" preamble found at the top of all documents. Its sole purpose is to prevent a browser from switching into so-called “quirks mode” when rendering a document; that is, the "`<!DOCTYPE html>`" doctype ensures that the browser makes a best-effort attempt at following the relevant specifications, rather than using a different rendering mode that is incompatible with some specifications.

https://developer.mozilla.org/en-US/docs/Glossary/Doctype

packages/hint-doctype/src/hint.ts
CKGrafico: Do you want to configure that the ProblemLocation always on the first line?
Something like:
```
export type ProblemLocation = {
    column: 0;
    line: 0;
};
```
Or do you prefer to search for the doctype tag and if this is in the 10th line calculate that the problem location is in the 10th line?

antross: > Or do you prefer to search for the doctype tag and if this is in the 10th line calculate that the problem location is in the 10th line?

I'd go with this unless we don't find a doctype at all, in which case the first line is fine. This should make it clearer to developers what needs to change, especially in cases where there's multiple doctypes or content after the doctype.

CKGrafico: Yes, thanks.
Only to be sure, this is how you expect the report for now?
```
await context.report(resource, null, 'Resource has no content.', undefined, defaultProblemLocation);
```
We didn't find how to do it without the null and the undefined.

CKGrafico:
We've also found that some websites have a `linebreak` on the first line.

linebreak example

We are not checking it, do we have to throw and error or it's ok and we have to remove the first character?

packages/hint-doctype/src/hint.ts
antross: Yes, we probably need to rework the available signatures for `report` a bit, but that's correct given the way things stand today.

I filed https://github.com/webhintio/rfcs/issues/47 to track making this better in the future.

antross:
> or it's ok and we have to remove the first character?

Per [the HTML spec](https://html.spec.whatwg.org/multipage/parsing.html#the-initial-insertion-mode), both whitespace and comments are allowed to occur before the doctype without making it invalid: 

> When the user agent is to apply the rules for the "initial" insertion mode, the user agent must handle the token as follows:
> * A character token that is one of U+0009 CHARACTER TABULATION, U+000A LINE FEED (LF), U+000C FORM FEED (FF), U+000D CARRIAGE RETURN (CR), or U+0020 SPACE
>   * Ignore the token.
> * A comment token
>   * Insert a comment as the last child of the Document object.
> * A DOCTYPE token
> ...

molant:
> and comments are allowed to occur before the doctype

I think we should flag comments before doctype. I want to remember that old versions of IE will switch to quirks mode (or any other compat mode) if they find them. Do you remember anything like that from your old days?

molant:
> Yes, but conditionally.

I think is a good practice to not put anything before the doctype and we should flag regardless of the browser support matrix. We know that things break prior IE10, but we don't know about other browsers and I don't want to investigate which break or not. This is similar to `no-bom`. Some browsers do it right, some other don't  but the general consensus is to not use it.

antross:
> we should flag regardless of the browser support matrix

Fine with me if we want to err on the side of caution here. When we document this, we can call out that "some browsers, including versions of IE prior to IE10, trigger quirks mode if a comment occurs before the DOCTYPE."

packages/hint-valid-doctype/README.md
```diff
@@ -0,0 +1,60 @@
...
+### Examples that **pass** the hint
+
+A doctype in the first line, without other information in that line.
+
+## How to use this hint?
+
+To use it you will have to install it via `npm`:
+
+```bash
+npm install @hint/hint-valid-doctype
```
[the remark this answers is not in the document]
CKGrafico: I think that this is resolved. Can you confirm?

packages/hint-doctype/src/hint.ts
CKGrafico: Updated with this

packages/hint-doctype/README.md
CKGrafico: Can you confirm that this is right ?

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,73 @@
+# Check that the DOCTYPE tag of the document is valid
+
+This hint checks if the HTML is using the most modern DOCTYPE.
+
+## Why is this important?
+
+In HTML, the doctype is the required "<!DOCTYPE html>" preamble
```
molant: Because this is  a quote from another website, we should put it as such. The aXe hint has an example on how we do this: https://github.com/webhintio/hint/blob/master/packages/hint-axe/README.md 
Can you mimic it?

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,73 @@
...
+found at the top of all documents. Its sole purpose is to prevent
+a browser from switching into so-called “quirks mode” when rendering
+a document; that is, the "<!DOCTYPE html>" doctype ensures that the
+browser makes a best-effort attempt at following the relevant specifications,
+rather than using a different rendering mode that is incompatible
+with some specifications.
+
+## What does the hint check?
+
+This hint checks if the HTML is using the most modern DOCTYPE.
```
molant: We should expand this a bit more. Say that we are checking that it uses the most modern and give an example, that there's nothing before the doctype of the document, and any other check that is done.

packages/hint-doctype/package.json
molant: @CKGrafico `@types/node` is still an old version.

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,106 @@
...
+This hint checks if the HTML is using the most modern and valid DOCTYPE.
+
+It checks that the DOCTYPE is in the first line and that there are no other
+lines before the DOCTYPE. This is important as some browsers,
+including versions of IE prior to IE10, trigger quirks mode if a comment
+occurs before the DOCTYPE.
+
+It checks that there is no additional info apart from the DOCTYPE on the same
+line and that it is not duplicated elsewhere in the document. For compatibility
+with older browsers, it also accepts the legacy-compat DOCTYPE. Examples:
```
antross: The legacy-compat DOCTYPE is actually there to support older HTML content generators - not older browsers.

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,106 @@
...
+<!--all content below the DOCTYPE tag-->
+```
+
+A DOCTYPE in multiple lines.
+
+```html
+<!DOCTYPE html>
+<!--some content below the DOCTYPE tag-->
+<!DOCTYPE html>
+<!--all content below the second DOCTYPE tag-->
```
antross: Multiple DOCTYPEs should be under the examples that **trigger** the hint.

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+`<!DOCTYPE html>`
+
+`<!doctype html>`
+
+`<!DOCTYPE html SYSTEM "about:legacy-compat">`
+
+### Examples that **trigger** the hint
+
+Also the hint will trigger if the DOCTYPE is not in the first line, for example
```
molant: > ~~Also~~ The hint will trigger...

The hint will also trigger if you use an old DOCTYPE:
```
<!DOCTYPE html PUBLIC
  "-//W3C//DTD XHTML 1.0 Transitional//EN"
  "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,106 @@
...
+<!DOCTYPE html SYSTEM "about:legacy-compat">
+<!--all content below the DOCTYPE tag-->
+```
+
+A DOCTYPE in multiple lines.
+
+```html
+<!DOCTYPE html>
+<!--some content below the DOCTYPE tag-->
+<!DOCTYPE html>
```
molant: Is having 2 DOCTYPE declarations valid?

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+                matches: text.match(doctypeRegExpFlexible)
+            };
+        };
+
+        const checkNoDoctypeInContent = async (matchInfo: MatchInformation, resource: string): Promise<boolean> => {
+            if (matchInfo.matches && matchInfo.matches.length > 0) {
+                return true;
+            }
+
+            await report(resource, 'The resource does not contain a valid DOCTYPE.');
```
antross: Nit: I think including an example would help someone fix the problem faster.
```suggestion
            await report(resource, 'The resource does not contain a valid DOCTYPE (e.g. `<!doctype html>`).');
```

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+            const { body } = response;
+            const { content } = body;
+
+            const globalMatch = getMatchInformation(content);
+
+            if (!checkNoDoctypeInContent(globalMatch, resource)) {
+                return;
+            }
+
+            checkNoDoctypeInCorrectLine(globalMatch, resource);
```
molant: This methods are async but we are not awaiting for them so there could be some problems. What do you think of doing:

```js

return Promise.all([
  checkNoDoctypeInCorrectLine(globalMatch, resource),
  checkDoctypeHasMoreThanValidInfo(globalMatch, resource),
  checkDoctypeIsDuplicated(globalMatch, resource)
]);

File: packages/hint-doctype/src/hint.ts [hunk 31]
```diff
@@ -0,0 +1,150 @@
...
+
+                return;
+            }
+
+            const { body } = response;
+            const { content } = body;
+
+            const globalMatch = getMatchInformation(content);
+
+            if (!checkNoDoctypeInContent(globalMatch, resource)) {
```
molant: I think you need to `await` here to get the right result. Otherwise I think you are just casting a Promise to a boolean.

packages/hint-doctype/tests/tests.ts
```diff
@@ -0,0 +1,238 @@
...
```
antross: Nit: we recently cut over to using the `[]` syntax for `Array` types.
```suggestion
const tests: HintTest[] = [
```

packages/hint-doctype/src/hint.ts
isabellachen: Is it ok if the function onFetchEndHTML returns `Promise<any>` instead of `Promise<[void, void, void]>` for this concrete case?

isabellachen: Ok, we're now returning void because of the linting, 'consistent-return'.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+            let found = false;
+
+            lines.forEach((line: string, i: number): void => {
+                if (found) {
+                    return;
+                }
+
+                const matched = doctypeRegExpFlexible.exec(line);
+
+                if (matched){
```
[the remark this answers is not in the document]
molant: @sarvaje how do we accept this change directly?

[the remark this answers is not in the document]
sarvaje: I don't see any button, so I think it is the author of the PR the one that has to accept the change.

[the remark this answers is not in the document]
sarvaje: ok, if you read the info button. Only PR author and assignees can commit the change.

molant:
LGTM, @sarvaje @antross want to take a last look?

packages/hint-doctype/tests/tests.ts
```diff
@@ -0,0 +1,238 @@
...
+        serverConfig: {
+            '/': {
+                content: '',
+                headers: { 'Content-Type': 'text/html' }
+            }
+        }
+    },
+    {
+        name: 'DOCTYPE not found should fail',
+        reports: [{ message: `The resource does not contain a valid DOCTYPE (e.g. \`<!doctype html>\`).` }],
```
antross: One last thing: it doesn't look like any of the tests validate the provided `ProblemLocation`. It can be added as an optional `position` property on the report:

```ts
reports: [{ message '...', position: { line: 0, col: 0 }]
```

I'd be particularly interested in having some validation for the more complicated positions (e.g. the test `DOCTYPE appearing more than once should fail`).

CKGrafico:
🎉

molant:
@alrra want to take a look or can I squash and merge?

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+            description: `This hint checks if the HTML is using the most modern DOCTYPE.`
+        },
+        id: 'doctype',
+        schema: [],
+        scope: HintScope.any
+    }
+
+    public constructor(context: HintContext) {
+        const regExpFactory = (isFlexible = false) => {
+            return new RegExp(`(<!doctype\\s+(html)\\s*(\\s+system\\s+"about:legacy-compat")?\\s*?>)${isFlexible ? '(.+)?' : ''}`, 'gi');
```
alrra: > 'about:legacy-compat'

Both the use of `"` and `'` are valid (see: https://html.spec.whatwg.org/multipage/syntax.html#the-doctype).

alrra:
> want to take a look or can I squash and merge?

I will review it more properly in the next few days.

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,121 @@
...
+It will trigger if there are multiple DOCTYPEs:
+
+```html
+<!DOCTYPE html>
+<!--some content below the DOCTYPE-->
+<!DOCTYPE html>
+<!--more content below the second DOCTYPE-->
+```
+
+It will trigger if there is additional information on the same line as the DOCTYPE:
```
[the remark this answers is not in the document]
CKGrafico: @molant This has sense for me too, if all you are ok we are going to implement this tomorrow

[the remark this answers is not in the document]
molant: Yes please, do the change and let's get it merged!

packages/hint-doctype/src/hint.ts
molant: @CKGrafico this too ⬆

isabellachen:
The feedback's been applied :)

packages/hint-doctype/src/hint.ts
```diff
@@ -30,13 +30,12 @@ export default class implements IHint {
     }
 
     public constructor(context: HintContext) {
-        const regExpFactory = (isFlexible = false) => {
-            return new RegExp(`(<!doctype\\s+(html)\\s*(\\s+system\\s+"about:legacy-compat")?\\s*?>)${isFlexible ? '(.+)?' : ''}`, 'gi');
+        const regExpFactory = () => {
```
sarvaje: Is this still necessary? You are using it just once, so maybe it is better to have something like `const doctypeRegExp = /myregex/gi`

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,114 @@
+# Check that the DOCTYPE tag of the document is valid
+
+This hint checks if the HTML is using the most modern DOCTYPE.
```
alrra: > is using the most modern document type declaration (a.k.a. DOCTYPE).

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,114 @@
...
+a document; that is, the "<!DOCTYPE html>" doctype ensures that the
+browser makes a best-effort attempt at following the relevant specifications,
+rather than using a different rendering mode that is incompatible
+with some specifications.
+
+***From [MDN glossary for DOCTYPE specification][docmdn].***
+
+## What does the hint check?
+
+This hint checks if the HTML is using the most modern and valid DOCTYPE.
```
alrra: DOCTYPE => document type declaration

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+It checks that the DOCTYPE is in the first line and that there are no other
```
alrra: > It checks that the DOCTYPE is in the first line and that there are no other
lines before the DOCTYPE.

If there are only whitespace characters before the doctype it's fine and everything should work as expected.

The problem with enforcing this is that some back-end languages generate HTML with a lot of space (from experience I know Java and PHP do this - e.g.: Look at the source code for: https://www.java.com/en/).

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,114 @@
...
+lines before the DOCTYPE. This is important as some browsers,
+including versions of IE prior to IE10, trigger quirks mode if a comment
```
alrra: For consistency, use:

> Internet Explorer prior to version 10...

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,114 @@
...
+<!DOCTYPE html SYSTEM "about:legacy-compat">
+<!--all content below the DOCTYPE-->
+```
+
+## How to use this hint?
+
+To use it you will have to install it via `npm`:
+
+```bash
+npm install @hint/hint-DOCTYPE
```
alrra: > hint-doctype

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Examples that **pass** the hint
+
+A DOCTYPE in the first line, without other information in that line.
+
+```html
+<!DOCTYPE html>
+<!--all content below the DOCTYPE-->
+```
+
```
alrra: I think we should change all example to use `<!doctype html>` and add here 2 examples:

```html
<!doctype html>
<!--all content below the DOCTYPE-->
```

```html
<!DOCTYPE html>
<!--all content below the DOCTYPE-->
```

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,114 @@
...
+It checks that there is no additional info apart from the DOCTYPE on the same
+line and that it is not duplicated elsewhere in the document. To support older
+HTML content generators, it also accepts the legacy-compat DOCTYPE. Examples:
+
+`<!DOCTYPE html>`
+
+`<!doctype html>`
+
+`<!DOCTYPE html SYSTEM "about:legacy-compat">`
```
alrra: I personally don't think we should recommend this. This is mostly for things such as [older outdated tools](http://bugzilla.bluegriffon.org/show_bug.cgi?id=634#c0) to not complain, not for browsers.

Also from: https://dev.w3.org/html5/html-author/#doctype-declaration

> Note: The term "legacy-compat" refers to compatibility with legacy producers only. In particular, **it does not refer to compatibility with legacy browsers, which, in practice, ignore SYSTEM identifiers and DTDs.**

Thoughts? 

---

See also: https://hsivonen.fi/doctype/#handling

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,114 @@
+# Check that the DOCTYPE tag of the document is valid
```
alrra: > Check if the page has the most modern document type declaration

packages/hint-doctype/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "extends": "../../.nycrc"
+  },
+  "peerDependencies": {
+    "hint": "^3.4.10"
+  },
+  "private": true,
+  "repository": "webhintio/hint",
+  "scripts": {
+    "build": "npm run clean && npm-run-all build:*",
+    "build-release": "npm run clean && npm run build:assets && tsc -b --inlineSourceMap false --removeComments true",
```
alrra: > tsc -b --inlineSourceMap false --removeComments true"

Remove `-b` (see: https://github.com/webhintio/hint/pull/1411).

packages/hint-doctype/README.md
isabellachen: @molant what do you think? That was a feedback added recently, but @alrra seems right.

isabellachen:
The latest feedback has been applied :)

packages/hint-doctype/README.md
alrra: Cc @webhintio/contributors

packages/hint-doctype/README.md
alrra: Cc @webhintio/contributors

packages/hint-doctype/README.md
alrra: @CKGrafico @isabellachen Can you also change the code for this? Thanks!

packages/hint-doctype/README.md
alrra: @CKGrafico @isabellachen Can you also change the code for this? Thanks!

packages/hint-doctype/README.md
CKGrafico: Yes, today we are going to fix this. Thanks!

isabellachen: The changes have been made, docs updated :)

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,126 @@
...
+            column: 0,
+            line: 0
+        };
+
+        const report = async (resource: string, message: string, problemLocation = defaultProblemLocation): Promise<void> => {
+            await context.report(resource, null, message, undefined, problemLocation);
+        };
+
+        const getCurrentDoctypeProblemLocation = (text: string): ProblemLocation[] => {
+            const lines = text.split('\n');
```
alrra: Should we use [`os.EOL`](https://nodejs.org/api/os.html#os_os_eol)? 

Or, am I missing something and we are normalizing things at some point?

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,126 @@
...
+            return false;
+        };
+
+        const checkNoDoctypeInCorrectLine = async (matchInfo: MatchInformation, resource: string): Promise<void> => {
+
+            if (matchInfo.locations[0].line === correctLine) {
+                return;
+            }
+
+            await report(resource, 'DOCTYPE is not in the first line.', matchInfo.locations[0]);
```
alrra: For consistency with the other error messages:

> 'doctype' should be specified before anything else.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,126 @@
...
+                matches: text.match(doctypeRegExp)
+            };
+        };
+
+        const checkNoDoctypeInContent = async (matchInfo: MatchInformation, resource: string): Promise<boolean> => {
+            if (matchInfo.matches && matchInfo.matches.length > 0) {
+                return true;
+            }
+
+            await report(resource, 'The resource does not contain a valid DOCTYPE (e.g. `<!doctype html>`).');
```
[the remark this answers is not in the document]
isabellachen: We have changed this a little bit (inspired in apple-touch-icons `'apple-touch-icon' link element should be specified in the '<head>'.`).

If there is no valid doctype but we find a tag with doctype word inside.
`'DOCTYPE' should be specified as '<!doctype html>'.`
else
`'DOCTYPE' was not specified.`

packages/hint-doctype/src/hint.ts
isabellachen: Thanks for the feedback, because of this bug with template strings https://github.com/nodejs/node/issues/3121 we had to normalize the content, but only in the tests, which use template strings.

We were also looking for examples using os.EOL in the code and found instead that '/n' was used a lot instead, so maybe that should be looked into.

alrra: Ok. 🤷🏻‍♂️

molant: I find your lack of faith... disturbing

CKGrafico: @molant the feedback was applied yesterday, do you want to reset it?

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,129 @@
...
+including versions of Internet Explorer prior to version 10,
+trigger quirks mode if a comment occurs before the DOCTYPE.
+
+It also checks that the DOCTYPE is not duplicated elsewhere in the document.
+
+Although an alternative legacy compatibility DOCTYPE is available,
+this hint does not recommend it. It is a common misconception that
+the legacy compatibility DOCTYPE refers to compatibility with
+legacy browsers, when, in fact, it is used to deal with [compatibility issues with
+outdated XML tools](http://bugzilla.bluegriffon.org/show_bug.cgi?id=634#c0).
```
alrra: Use link label for the URL.

(I will take care of this before merging)

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,129 @@
...
+To use it you will have to install it via `npm`:
+
+```bash
+npm install @hint/hint-doctype
+```
+
+Note: You can make `npm` install it as a `devDependency` using the `--save-dev`
+parameter, or to install it globally, you can use the `-g` parameter. For
+other options see
+[**npm's** documentation](https://docs.npmjs.com/cli/install).
```
alrra: Use link label for the URL.

(I will take care of this before merging)

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,129 @@
...
+And then activate it via the `.hintrc` or `hintrc`
+configuration file:
+
+```json
+{
+    "connector": {...},
+    "formatters": [...],
+    "parsers": [...],
+    "hints": {
+        "doctype": "error"
```
alrra: >  "doctype": "error",
>  ...

packages/hint-doctype/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Further Reading
+
+* [DOCTYPE (Wikipedia)][docwiki]
+* [DOCTYPE (MDN)][docmdn]
+* [Activating Browser Modes with Doctype][hsivonen]
+
+<!-- Link labels: -->
+
+[docwiki]: https://en.wikipedia.org/wiki/Document_type_declaration
```
alrra: Order these alphabetically.

packages/hint-doctype/src/hint.ts
```diff
@@ -0,0 +1,131 @@
...
+                return true;
+            }
+
+            if (!doctypeFlexibleRegExp.exec(content)) {
+                await report(resource, `'DOCTYPE' was not specified.`);
+
+                return false;
+            }
+
+            await report(resource, `'DOCTYPE' should be specified as '<!doctype html>'.`);
```
alrra: For consistency with other error messages, let' use `doctype` instead of `DOCTYPE` here and everywhere else.

molant:
image

🎉

molant:
Congrats to @isabellachen and @CKGrafico on their first merged PR 🎉

packages/hint-doctype/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "files": [
+      "dist/tests/**/*.js"
+    ],
+    "timeout": "1m"
+  },
+  "description": "this hint checks if the HTML is using the most modern DOCTYPE.",
+  "devDependencies": {
+    "@hint/utils-tests-helpers": "^1.0.0",
+    "@types/debug": "0.0.31",
+    "@types/node": "10.9.1",
```
alrra: This and some other ones need to be updated.

(I will take care of this once I merge)

PR #1402, opened 2018-10-15
Fix: Ignore domains when they don't support https

### LINKED ISSUE #1401: `strict-transport-security` can make slower the analysis of an url.
Some sites doesn't support `https` at all, and when we have active the hint `strict-transport-security`, we try to get the `https` version of all the resource. This petition can throw a `ETIMEOUT` after a while. If the site we are analyzing has a lot of resources, this analysis can take forever.

sarvaje:
Fix #1401

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint-strict-transport-security/src/hint.ts
+++ b/packages/hint-strict-transport-security/src/hint.ts
@@ -213,9 +213,7 @@ export default class StrictTransportSecurityHint implements IHint {
                      * If the https resource can't be fetched, add
                      * the domain to the unsupported list
                      */
-                    if (err && err.error && err.error.code === 'ETIMEDOUT') {
-                        unsupportedDomains.add(urlObject.host);
-                    }
+                    unsupportedDomains.add(urlObject.host);
                 }
 
                 return;
--- a/packages/hint-strict-transport-security/src/hint.ts
+++ b/packages/hint-strict-transport-security/src/hint.ts
@@ -46,6 +46,8 @@ export default class StrictTransportSecurityHint implements IHint {
         const statusApiEndPoint = `https://hstspreload.org/api/v2/status?domain=`;
         /** Endpoint to verify that the domain name is qualified to be preloaded */
         const preloadableApiEndPoint = `https://hstspreload.org/api/v2/preloadable?domain=`;
+        /** Set of unsupported domains to avoid make unnecessary requests. */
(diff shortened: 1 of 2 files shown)


packages/hint-strict-transport-security/src/hint.ts
```diff
@@ -198,6 +208,14 @@ export default class StrictTransportSecurityHint implements IHint {
                 } catch (err) {
                     // HTTPS site can't be fetched, do nothing.
                     debug(`${resource} doesn't support HTTPS`);
+
+                    /*
+                     * If the https resource can't be fetched, add
+                     * the domain to the unsupported list
+                     */
+                    if (err && err.error && err.error.code === 'ETIMEDOUT') {
```
molant: Can we receive other errors than `ETIMEDOUT`.
Maybe it's worth adding some tests to verify that we are not doing any more requests if the first time it fails and that way we can check if there are other type of errors coming from the server?

sarvaje: >Can we receive other errors than ETIMEDOUT?

No, looks like some other erros can be received, like `ECONNRESET`.

packages/hint-strict-transport-security/src/hint.ts
```diff
@@ -178,7 +180,15 @@ export default class StrictTransportSecurityHint implements IHint {
             }
 
             if (!isHTTPS(resource) && !headerValue) {
-                const httpsResource = url.format(Object.assign(new URL(resource), { protocol: `https` }));
+                const urlObject = new URL(resource);
+
+                if (unsupportedDomains.has(urlObject.host)) {
+                    debug(`${resource} ignored because the domain ${urlObject.host} doesn't support HTTPS`);
```
alrra: For consistency, add `.` at the end and use `does not`.

packages/hint-strict-transport-security/src/hint.ts
```diff
@@ -198,6 +208,12 @@ export default class StrictTransportSecurityHint implements IHint {
                 } catch (err) {
                     // HTTPS site can't be fetched, do nothing.
                     debug(`${resource} doesn't support HTTPS`);
+
+                    /*
+                     * If the https resource can't be fetched, add
+                     * the domain to the unsupported list
```
alrra: Use `HTTPS` and `.` at the end.

PR #1409, opened 2018-10-16
Fix: Refactor debugging protocol

### LINKED ISSUE #1398: Refactor Debugging Protocol
`utils-debugging-protocol` needs some ❤
The interaction with the debugging protocol is a bit complex and ~~we~~I've kept adding hacks during the last few months making the code messiers and messier. It's time to pay the debt.

Some examples of issues we have or had recently:

1. #1395 
2. https://github.com/webhintio/rfcs/issues/44
3. #1325
4. Also `rawResponse` can be empty in most cases (similar to #1395 but affecting `performance-budget`)
5. #1403

molant:
## Pull request checklist

Make sure you:

Refactor of the debugging protocol code. We were having a few issues and needed some clean up.

Fix #1398 

Keep in mind this is a work in progress. I still need to make all the test pass and I want to add some documentation on how the debugging protocol works because I will completely forget in 2 weeks.

### DIFF
--- a/packages/connector-chrome/tests/collect.ts
+++ b/packages/connector-chrome/tests/collect.ts
@@ -55,7 +55,7 @@ const runTest = async (t: GenericTestContext<Context<any>>, ConnectorConstructor
     await connector.close();
 };
 
-test(`[${name}] The HTML is downloaded is present in a 'link' element with 'rel' attribute set to 'icon' `, async (t) => {
+test(`[${name}] The HTML is downloaded and the right event emitted`, async (t) => {
     const serverConfig: ServerConfiguration = { '/': generateHTMLPage(`<title>Test</title>`) };
 
     await runTest(t, ChromeConnector, serverConfig);
--- a/packages/connector-chrome/tests/collect.ts
+++ b/packages/connector-chrome/tests/collect.ts
@@ -6,6 +6,7 @@ import { URL } from 'url';
 
 import * as sinon from 'sinon';
 import test, { GenericTestContext, Context } from 'ava';
+import * as isCI from 'is-ci';
 
 import { createServer, ServerConfiguration } from '@hint/utils-create-server';
 import { IConnector, IConnectorConstructor } from 'hint/dist/src/lib/types';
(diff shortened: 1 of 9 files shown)


packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -22,26 +22,257 @@ import { Crdp } from 'chrome-remote-debug-protocol';
...
 const debug: debug.IDebugger = d(__filename);
 
+enum RequestStatus {
+    willBeSent = 'willBeSent',
+    responseReceived = 'responseReceived',
+    loadingFinished = 'loadingFinished',
+    loadingFailed = 'loadingFailed'
+}
+
+type NetworkEvent = Crdp.Network.RequestWillBeSentEvent |
```
molant: @antross can you take a 👀 at this? Not sure this is the best option.

antross: Taking a look now...

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -22,26 +22,257 @@ import { Crdp } from 'chrome-remote-debug-protocol';
...
+    Crdp.Network.ResponseReceivedEvent |
+    Crdp.Network.LoadingFinishedEvent |
+    Crdp.Network.LoadingFailedEvent;
+
+let overrideInvalidCert: false;
```
sarvaje: IMO, this should be a private variable in each class, and it should be pass through the constructor.

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -22,26 +22,257 @@ import { Crdp } from 'chrome-remote-debug-protocol';
...
+
+class RequestResponse {
```
sarvaje: Each class should go to a different file.

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -22,26 +22,257 @@ import { Crdp } from 'chrome-remote-debug-protocol';
...
+            strictSSL: !overrideInvalidCert
+        };
+
+        const request: Requester = new Requester(options);
+
+        return request.get(href);
+    }
+
+    private _rawResponse: Buffer | undefined;
+    private getRawResponse() {
```
sarvaje: don't you want this to be `private get rawResponse()` as the other properties??

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -22,26 +22,257 @@ import { Crdp } from 'chrome-remote-debug-protocol';
...
+            if (key.startsWith(':')) {
+                return final;
+            }
+
+            final[key] = value;
+
+            return final;
+        }, {} as Crdp.Network.Headers);
+
+        return this.fetchContent(responseUrl, validHeaders)
```
sarvaje: any reason to not use `async/await`?

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -22,26 +22,257 @@ import { Crdp } from 'chrome-remote-debug-protocol';
...
+            })
+            .then((value) => {
+                that._rawResponse = value;
+
+                return value;
+            });
+    }
+
+    private _response: Response | undefined;
+    public getResponse(element: AsyncHTMLElement | null): Response | undefined {
```
sarvaje: `public get response()`?

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -295,14 +489,25 @@ export class Connector implements IConnector {
 
         // DOM is not ready so we queue up the event for later
         if (!this._dom) {
+            console.error(`(${params.requestId}) Queuing onLoadingFailed`);
```
sarvaje: use `logger.error` or remove `console.error`

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -503,7 +678,8 @@ export class Connector implements IConnector {
         if (!isTarget) {
             // DOM is not ready so we queue up the event for later
             if (!this._dom) {
-                this._pendingResponseReceived.push(this.onResponseReceived.bind(this, params));
+                console.error(`(${params.requestId}) Queuing onLoadingFinished`);
```
sarvaje: use `logger.error` or remove `console.error`

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
molant: I wish I could, but it has some `async` stuff and the return type is `NetworkData` instead of `Buffer`

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
molant: so the `catch` can be handle up by whoever is calling this method. This way I don't have to put a `try/catch` here.

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
molant: Remove `console.error`. Left over from testing.

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
sarvaje: ok!

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
sarvaje: Ahhh, ok! Maybe can you add the type to the function then. `private getRawResponse(): NetworkData`

molant:
@sarvaje @antross I should have addressed all the feedback you gave me in different channels and tests for `connector-chrome` are now passing in my machine so this should be ready for another review.

molant:
> now passing in my machine so this should be ready for another review

And for whatever reason the favicon tests fail in Travis 😭

molant:
> And for whatever reason the favicon tests fail in Travis 😭

@alrra could it be that because we are using `npm` I'm getting the previous version of `utils-debugging-protocol` and that could be messing things up?

alrra:
> we are using npm

We are using [`yarn`](https://travis-ci.org/webhintio/hint/jobs/442486032#L648). 

`npm` is only used when releasing packages.

molant:
The problem seems to be that chrome will not download the favicon when running in headless mode.

I think that I'll disable those tests via `is-ci` if needed. Any objections?

packages/utils-debugging-protocol-common/src/RequestResponse.ts
```diff
@@ -0,0 +1,271 @@
...
+    /** The final URL after redirects. Could be the same as `originalUrl`. */
+    public get finalUrl(): string {
+        // `willBeSent` gets updated with each redirect so this is the final requested url
+        return this.willBeSent.request.url;
+    }
+
+    private _willBeSent: Crdp.Network.RequestWillBeSentEvent | undefined;
+    /** The payload returned by the debugging protocol in the `RequestWillBeSent` event. */
+    public get willBeSent(): Crdp.Network.RequestWillBeSentEvent {
+        return this._willBeSent!;
```
antross: Is there a reason this one is assumed not to be `undefined` whereas the others may be?

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -136,7 +127,7 @@ export class Connector implements IConnector {
              * doesn't need to be escaped.
              */
             const query: string = `[src$="${basename}" i],[href$="${basename}" i],[src$="${decodeURIComponent(basename)}" i],[href$="${decodeURIComponent(basename)}" i]`;
-            const newElements: Array<AsyncHTMLElement> = await this._dom.querySelectorAll(query);
+            const newElements: Array<AsyncHTMLElement> = await this._dom!.querySelectorAll(query);
```
antross: Nit: `this._dom!` is technically safe as all callers (or callers of callers) ensured it existed before we got here. However, we're so far removed by this point it's hard to tell without reading a fair amount of code. 

What do you think about adding `dom: CDPAsyncHTMLDocument` as a parameter to this function and having it explicitly passed down from the callers that performed the check to ensure it existed?

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -311,214 +258,30 @@ export class Connector implements IConnector {
...
+                debug(`Error finding element for request ${requestResponse.requestId}. element will be null`);
             }
         }
 
-        const response: Response = await this.createResponse(params, element);
+        const response: Response = requestResponse.getResponse(element);
 
         const request: Request = {
-            headers: params.response.requestHeaders as HttpHeaders,
+            headers: normalizeHeaders(requestResponse.responseReceived!.response.requestHeaders) as HttpHeaders,
```
antross: Nit: similar to my earlier comment about `this._dom`, consider passing `responseReceived` down as a parameter from the calling functions that checked to ensure it existed to make the fact that this is not `undefined` clearer (and understandable by the compiler).

sarvaje:
>The problem seems to be that chrome will not download the favicon when running in headless mode.
>
>I think that I'll disable those tests via is-ci if needed. Any objections?

@molant can this be a problem in a scanner using chrome connector?

packages/connector-chrome/tests/events.ts
```diff
@@ -295,7 +295,7 @@ test(`[${name}] Events`, async (t) => {
     }
 
     // List of events that only have to be called once per execution
-    const singles = ['fetch::error', 'scan::start', 'scan::end', 'fetch::end::html'];
+    const singles = ['scan::start', 'scan::end', 'fetch::end::html'];
```
sarvaje: question: Why did you remove the `fetch::error`?

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -16,30 +16,30 @@ import { promisify } from 'util';
...
-import { atob } from 'abab';
+
 import { Crdp } from 'chrome-remote-debug-protocol';
 
 import { createCDPAsyncHTMLDocument, CDPAsyncHTMLDocument, AsyncHTMLElement } from './cdp-async-html';
-import { getContentTypeData, getType } from 'hint/dist/src/lib/utils/content-type';
+import { getType } from 'hint/dist/src/lib/utils/content-type';
 import { debug as d } from 'hint/dist/src/lib/utils/debug';
-import * as logger from 'hint/dist/src/lib/utils/logging';
+// import * as logger from 'hint/dist/src/lib/utils/logging';
```
sarvaje: remove comment.

packages/utils-debugging-protocol-common/src/RequestResponse.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    /** Update the payload received on the `LoadingFailed` event for this `requestId`. */
+    public updateLoadingFailed(event: Crdp.Network.LoadingFailedEvent) {
+        this._loadingFailed = event;
+
+        this._status = RequestStatus.loadingFailed;
+        this.logInfo(RequestStatus.loadingFailed);
+    }
+
+    public constructor(network: Crdp.NetworkClient, event: Crdp.Network.RequestWillBeSentEvent, overrideInvalidCert: boolean) {
```
sarvaje: IMO, `overrideInvalidCert` should be part of an `options` object. Just in case we need to pass something else in the future.

packages/utils-debugging-protocol-common/src/RequestResponse.ts
molant: Nope, left over from when I was doing tests.

packages/utils-debugging-protocol-common/src/RequestResponse.ts
molant: `RequestResponse` is meant to be used only by `utils-debugging-protocol`. If we need to add more things we can then switch to an `options` object without breaking.

packages/connector-chrome/tests/events.ts
molant: Because of https://github.com/webhintio/hint/issues/1398#issuecomment-429929512
`fetch::error` is no longer a single event, we receive it twice probably because of speculative downloading.

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -247,7 +246,7 @@ export class Connector implements IConnector {
 
         debug(`Error found loading ${resource}:\n%O`, params);
 
-        const element: AsyncHTMLElement = (await this.getElementFromRequest(params.requestId))!;
+        const element: AsyncHTMLElement = (await this.getElementFromRequest(params.requestId, this._dom))!;
```
antross: Ideally this should be `AsyncHTMLElement | null` (or simply omitted), but that requires updating the definition of `FetchError` to have `element: IAsyncHTMLElement | null`, which would be good to align it with the definition of `FetchEnd` that already allows `null`, but may be more than you want to do for this PR (as adding the `null` to `FetchError` could have ripple effects requiring changes in other packages).

molant: Yeah, I tried that before submitting the changes but it seemed to be more changes than needed. I'm working on fixing the perf budget tests but I can look into this later.

molant: I've fixed this.

molant:
All feedback should be in and hopefully test pass now. At the end I had to force the download of the favicon when running as `--headless` so there are a few more files changed to expose the flags used to open chrome.
If you can take another look it will be great.

packages/connector-chrome/tests/collect.ts
```diff
@@ -55,11 +55,19 @@ const runTest = async (t: GenericTestContext<Context<any>>, ConnectorConstructor
     await connector.close();
 };
 
+test(`[${name}] The HTML is downloaded is present in a 'link' element with 'rel' attribute set to 'icon' `, async (t) => {
```
sarvaje: >The HTML ~is~ downloaded is present`?

molant:
@antross, I've modified a few new files since you last approved it. Can you please take a look?
@alrra do you want to take a look? I'd like to have this merged and published tomorrow Friday and before merging any of the breaking changes so we can do another big run test.
I know you are traveling so I can take care of that if you can't.

antross:
>  I've modified a few new files since you last approved it. Can you please take a look?

@molant Still looks good to me.

packages/utils-debugging-protocol-common/src/RequestResponse.ts
```diff
@@ -0,0 +1,290 @@
...
+    willBeSent = 'willBeSent',
+    responseReceived = 'responseReceived',
+    loadingFinished = 'loadingFinished',
+    loadingFailed = 'loadingFailed'
+}
+
+export class RequestResponse {
+    private _overrideInvalidCert: boolean;
+
+    /** The debugging protocol Network Client used to download the response body */
```
alrra: For consistency add `.` at the end, and and everywhere else.

packages/utils-debugging-protocol-common/src/RequestResponse.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    private _originalUrl: string;
+    /** The initial URL that started the request process before any redirects. */
+    public get originalUrl() {
+        return this._originalUrl;
+    }
+
+    /** The final URL after redirects. Could be the same as `originalUrl`. */
+    public get finalUrl(): string {
+        // `willBeSent` gets updated with each redirect so this is the final requested url
```
alrra: > URL

packages/utils-debugging-protocol-common/src/RequestResponse.ts
```diff
@@ -0,0 +1,290 @@
...
+            let rawContent = Buffer.alloc(0);
+            let rBody = {
+                content: '',
+                rawContent,
+                rawResponse: () => {
+                    return Promise.resolve(Buffer.alloc(0));
+                }
+            };
+
+
```
alrra: Remove extra empty line.

PR #1415, opened 2018-10-17
Breaking: Change `context.report` to take an `options` object 

antross:
## Pull request checklist

Make sure you:

Fix webhintio/rfcs#47.

Now `report` only has two required parameters, a string `resource`
and a string `message`. All other parameters are now optionally
passed as part of a third `options` parameter of type
`ReportOptions`.

Old code like:
```ts
await context.report(
    resource,
    null,
    message,
    undefined,
    location
);
```

Can now be written as:
```ts
await context.report( resource, message, { location });
```


antross:
Marking `work-in-progress` as I still need to add official documentation for the full set of options available on `report`.

antross:
@molant although we haven't touched the package in 8 months this is also now failing when I run the test against master. 

Digging into the [amphtml-validator npm package](https://github.com/ampproject/amphtml/tree/master/validator/nodejs/) on GitHub, it looks like most of the logic is contained in a `validator.js` file that is dynamically fetched from `https://cdn.ampproject.org/v0/validator.js` instead of being included in the package. This file is built from the core [amphtml-validator folder](https://github.com/ampproject/amphtml/tree/master/validator) which is not published on npm. Checking the HTTP headers of this online, it has a last-modified header for yesterday: Wed, 17 Oct 2018 19:15:00 GMT.

For now I'm going to update the test (in a separate PR so we can merge it quickly as this affects all our branches due to being dynamically fetched), but I'd like your input on what we want to do with this test going forward. We have the option of passing our own `validator.js` to the call to `getInstance` so we could keep a downloaded copy in-box to avoid being surprised by future updates. That said, we'd need to manually update this so it may not be worth it unless we feel we need to eliminate the extra service call.

antross:
@molant, just saw you found and updated that last night as I rebased master - still curious about your thoughts for this package going forward.

molant:
> still curious about your thoughts for the hint-amp-validator package going forward.

We haven't had any problem in the last few months so I would do anything right now. That said, I think we should open an issue with low priority with what you propose in case someone wants to help with it. If it's documented well enough it could be a good first contribution.

molant:
@antross you have some linting issue in the `apple-touch-icon` hint that fails the build. Can you please fix it?

antross:
> you have some linting issue in the apple-touch-icon hint that fails the build. Can you please fix it?

🤦‍♂️ - on it

antross:
@alrra This is now up-to-date when you have time to take a look. CI is passing on the node versions for which canvas-prebuilt currently works.

packages/hint/docs/contributor-guide/how-to/hint.md
```diff
@@ -94,15 +94,29 @@ with the website and report errors.
...
 ```
 
 * `context.report()` is an asynchronous method, you should always `await`.
 * `resource` is the URL of what is being analyzed (HTML, JS, CSS, manifest,
   etc.)
-* `element` is the `IAsyncHTMLElement` that triggered the problem. Not always
-  necessary. In the case of an image, script, style, it will be an `img`,
-  `script`, `link`, etc.
+* `message` is the text to show to the user about the problem.
+* `element` is an optional `IAsyncHTMLElement` where the issue was found
```
molant: I think this is a bit confusing. The third parameter is the `options` object that can have an `element` property, right? I think we should make it clearer. Right now this splits the description between `element` and `options`. What about something like:

```ts
await context.report(resource, message, options);
```
* `context.report()` is an asynchronous method, you should always `await`.
* `resource` is the URL of what is being analyzed (HTML, JS, CSS, manifest,
  etc.)
* `message` is the text to show to the user about the problem.
* options is an (optional) object that can contain the following:
  * `element` is an optional `IAsyncHTMLElement` where the issue was found ...
  * `codeSnippet` is a string of source code to display (defaults to the
  `outerHTML` of `element`).
  * ...

alrra:
@antross Can you take care of the conflicts? Thanks!

antross:
Yep, I'll have them fixed shortly.

PR #1440, opened 2018-10-24
Chore: Replace ESLint `no-unused-vars`

molant:
## Pull request checklist

Make sure you:

Disable `no-unused-vars` and replace it with TypeScript's
`noUnusedLocals` compiler option.

Fix webhintio/rfcs#46


.eslintrc.json
```diff
@@ -263,7 +260,7 @@
         "typescript/no-namespace": "error",
         "typescript/no-parameter-properties": "error",
         "typescript/no-triple-slash-reference": "error",
-        "typescript/no-unused-vars": "error",
+        "typescript/no-unused-vars": "off",
         "typescript/no-use-before-define": "error",
```
antross: Slightly off-topic, but I notice we have a few of the same rules enabled for both in general and under the `typescript/` section (e.g. `no-use-before-define`). From my experience it seems both of these can flag in TypeScript files which feels a bit unnecessarily redundant. 

Is there a need to keep both or should we prefer having one enabled over the other?

packages/create-hint/src/no-official-files/.eslintrc.json
```diff
@@ -258,7 +255,7 @@
         "typescript/no-namespace": "error",
         "typescript/no-parameter-properties": "error",
         "typescript/no-triple-slash-reference": "error",
-        "typescript/no-unused-vars": "error",
+        "typescript/no-unused-vars": "off",
```
antross: If we're updating the templates, it seems we should also turn on the `noUnusedLocals` compiler flag in `shared-templates/tsconfig.json.hbs`.

packages/create-parser/src/no-official-files/.eslintrc.json
```diff
@@ -258,7 +255,7 @@
         "typescript/no-namespace": "error",
         "typescript/no-parameter-properties": "error",
         "typescript/no-triple-slash-reference": "error",
-        "typescript/no-unused-vars": "error",
+        "typescript/no-unused-vars": "off",
```
antross: Should also update `shared-templates/tsconfig.json.hbs` for this package.

.eslintrc.json
molant: I think this is because we have JavaScript files (`/scripts`) and the rest is TypeScript. Are you sure the JavaScript one will flag on TypeScript files too? If so we can probably disable the TS one.

packages/create-hint/src/no-official-files/.eslintrc.json
molant: I was sure you were going to say this 😁
I think I'll enable it on the path that's not `official`. The reason is that we are extending from the main `tsconfig.json`. And you will say:
- "But we have `strict: strue`! 
- Yes, but the other packages also have it and it was for consistency that we added it.

One thing I've realized is that we don't have `strict: true` in the `non-official` path though..

Does this sound ok?

.eslintrc.json
antross: > Are you sure the JavaScript one will flag on TypeScript files too?

No, I dug a bit deeper and according to the documentation for `eslint-plugin-typescript` the `typescript/*` rules should only flag TypeScript-specific constructs. Therefore any double-reporting I've seen should probably be considered a bug in that project (and may be due to enhancements to the ESLint core via the official `typescript-eslint-parser` covering more scenarios than when the rule was originally authored).

Issue #1445, opened 2018-10-25
Builds fail on `node` v11 because missing `canvas-prebuilt`


alrra:
Ref https://travis-ci.org/webhintio/hint/jobs/445942975#L1002-L1034

---

This issue is more of FYI then anything else.

alrra:
This issues is more of FYI then anything else.

molant:
I'll ping him.

molant:
Maybe have a fallback to compile `canvas`?

molant:
BTW<jsdom [will soon have canvas 2.0 support](https://github.com/jsdom/jsdom/pull/1964)

sarvaje:
> BTW, jsdom will soon have canvas 2.0 support

In fact, it is already published: https://github.com/jsdom/jsdom/blob/master/Changelog.md

molant:
Some updates about this.

* `canvas` will automatically download the binaries under release of `canvas-prebuilt`. `canvas-prebuilt` will not have another release on npm
* `jsdom` 13 drops support for `canvas` < 2
* there are binaries for Linux and Windows for node 11 and `canvas` 2.0.1
![image.png](https://images.zenhubusercontent.com/56d5e96d6caf56c9d38d18fe/a2f8ec36-db39-48b0-b827-82815e3f8621)

I'm upgrading version in #1465 but I'm starting to think that `canvas` should be an optional dependency and we should try to better handle when loading fails. E.g.: the vscode extension will complain if we've downloaded a version that is not for node 8 (current version used by vscode). This might involve doing a PR to jsdom to change how it's being loaded among other things.

@webhintio/core what do you think?

antross:
What about dropping `canvas` entirely? If we don't include it, `jsdom` won't try to use it.

I understand this would provide a less realistic experience when using `connector-jsdom`, but I'm starting to think this isn't worth the trouble for supporting a few extra APIs - especially when a bunch of other visual APIs don't work in `jsdom` anyway.

molant:
> What about dropping `canvas` entirely? If we don't include it, `jsdom` won't try to use it.

IIRC images will not be downloaded and thus provide less accurate results. @sarvaje can you confirm? Maybe this is no longer the case with the new `ResourceLoader`?

In any case I've opened https://github.com/jsdom/jsdom/issues/2410

sarvaje:
>IIRC images will not be downloaded and thus provide less accurate results.

[...]

PR #1466, opened 2018-10-30
Build: Allow node v11 builds to fail for now

alrra:
## Pull request checklist

Make sure you:

Allow node v11 builds to fail for now as `canvas-prebuilt` will be missing.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Ref #1445

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,6 +87,19 @@ install:
       fi
 
 jobs:
+
+  # Allow node v11 builds to fail for now as `canvas-prebuilt`
+  # will be missing.
+  #
+  # Ref: https://github.com/webhintio/hint/issues/1445
+
+  allow_failures:
+    - node_js: node
+
+  fast_finish: true
+
+  # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
   include:
 
     # Stage used for checking for broken links.


alrra:
> Are we planning on dropping support for node 8?

Are we? 

Note: Even our site can currently only be built using node 8.

See also: https://github.com/nodejs/Release#release-schedule

> see us explicitly target node 8, 10 and add 11 when canvas-prebuilt has a binary

The problem with explicitly targeting and adding the next version of node manually is that we will forget to update and then we will discover problems such as the one with `canvas-prebuilt` when users encounter them, so too late. I personally rather know what fails as soon as possible.

---

@molant This was meant only as a temporary solution.

molant:
> Even our site can currently only be built using node 8.

That's because another package and the related issues has some ideas on how to solve it (BTW, I'm waiting on feedback about them).

> The problem with explicitly targeting and adding the next version of node manually is that we will forget to update

Right now we are not testing on node 8. Also I don't think we will forget as there are a few people in the team that follow closely when there are new releases.


molant:
The problem we currently have with node 11 and Travis is with Chrome. It gets stalled from time to time in different areas. jsdom tests are passing without issues. I can't reproduce the error consistently, but I've seen it locally.

PR #1475, opened 2018-11-02
Chore: Update `typescript-eslint-parser` to `v20.1.1`

molant:
## Pull request checklist

Make sure you:

Fix: #1468

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "rimraf": "^2.6.2",
     "shelljs": "^0.8.2",
     "typescript": "^3.1.5",
-    "typescript-eslint-parser": "20.0.0",
+    "typescript-eslint-parser": "20.1.1",
     "yargs": "^12.0.2"
   },
   "engines": {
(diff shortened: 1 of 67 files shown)


antross:
> I'm going to see if I can address these in my context.report PR 

Scratch that. Looks like this will have to touch a lot more files than I originally thought now that I ran the linter locally (run `yarn lint-packages` from the root). We should probably fix it as a separate PR.

Of note is this new version seems to have broken eslint's understanding of inline types for object deconstruction such that it incorrectly thinks we are re-declaring a variable when we're actually specifying a type. That rule is useful in other contexts, though so I'd hate to disable it globally, nor do I want to pepper the code base with a bunch more eslint-disable-line statements...

@molant Thoughts?

molant:
Couple options I can think of:

1. Disable globally any other rules that break for now and track the issue I mention in the commit so we can re-enable it later when it's fixed (could take a long time).
1. Revert back the commits and add `typescript-eslint-parser` to the list of packages to ignore by green keeper for now (and still track that issue).

I think we should avoid any code changes.
In a separate note, why can't we enable `lint-packages` under the general `lint`.

PR #1480, opened 2018-11-05
New: Allow hint metadata to be imported separately

antross:
## Pull request checklist

Make sure you:

Extracts existing `HintMetadata` objects into dedicated files and
adds a new `name` field to the `docs` property.

Allows hint metadata to be imported and bundled separately
from the implementation of a hint, reducing bundle size for
documentation scenarios.

Also updates all hint names for consistency both in the new
metadata and in the title of each associated `README`.

### DIFF
--- a/packages/hint-amp-validator/src/meta.ts
+++ b/packages/hint-amp-validator/src/meta.ts
@@ -1,0 +1,20 @@
+import { Category } from 'hint/dist/src/lib/enums/category';
+import { HintScope } from 'hint/dist/src/lib/enums/hintscope';
+import { HintMetadata } from 'hint/dist/src/lib/types';
+
+const meta: HintMetadata = {
+    docs: {
+        category: Category.performance,
+        description: `Require HTML page to be AMP valid.`,
+        name: 'AMP HTML validator'
+    },
+    id: 'amp-validator',
+    schema: [{
+        additionalProperties: false,
+        properties: { 'errors-only': { type: 'boolean' } },
+        type: 'object'
+    }],
+    scope: HintScope.any
+};
+
+export default meta;
--- a/packages/hint-http-compression/src/hint.ts
+++ b/packages/hint-http-compression/src/hint.ts
@@ -21,6 +21,7 @@ import isRegularProtocol from 'hint/dist/src/lib/utils/network/is-regular-protoc
 import normalizeString from 'hint/dist/src/lib/utils/misc/normalize-string';
 
 import { CompressionCheckOptions } from './types';
+
 import meta from './meta';
 
 const decompressBrotli = promisify(brotli.decompress) as (buffer: Buffer) => Promise<Buffer>;
--- a/packages/hint-http-compression/src/hint.ts
+++ b/packages/hint-http-compression/src/hint.ts
@@ -11,18 +11,17 @@ import { promisify } from 'util';
 
 import * as brotli from 'iltorb';
 
-
-import { Category } from 'hint/dist/src/lib/enums/category';
-import { HintScope } from 'hint/dist/src/lib/enums/hintscope';
 import { HintContext } from 'hint/dist/src/lib/hint-context';
-import { FetchEnd, IAsyncHTMLElement, IHint, NetworkData, Response, HintMetadata, HttpHeaders } from 'hint/dist/src/lib/types';
+import { FetchEnd, IAsyncHTMLElement, IHint, NetworkData, Response, HttpHeaders } from 'hint/dist/src/lib/types';
 import { asyncTry } from 'hint/dist/src/lib/utils/async-wrapper';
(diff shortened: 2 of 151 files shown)


antross:
@webhintio/core If anyone has opinions on what the user-friendly names should be for any of the hints, please share and I'll update the PR. I already made a few edits to try to keep the names to a reasonable length.

packages/hint-amp-validator/src/meta.ts
```diff
@@ -0,0 +1,20 @@
+import { Category } from 'hint/dist/src/lib/enums/category';
+import { HintScope } from 'hint/dist/src/lib/enums/hintscope';
+import { HintMetadata } from 'hint/dist/src/lib/types';
+
+const meta: HintMetadata = {
+    docs: {
+        category: Category.performance,
+        description: `Require HTML page to be AMP valid.`,
```
molant: A few months back @alrra and I were talking about how most of the hints started with `Require XXX` and how we wanted to change that because it looks weird on the site 😅

I think now is a good time. I'll create a new comment with a table so it's easier to digest and propose changes.

packages/hint-sri/src/meta.ts
```diff
@@ -0,0 +1,26 @@
...
+import { HintScope } from 'hint/dist/src/lib/enums/hintscope';
+import { HintMetadata } from 'hint/dist/src/lib/types';
+
+import { algorithms } from './types';
+
+const meta: HintMetadata = {
+    docs: {
+        category: Category.security,
+        description: `Require scripts and link elements to use Subresource Integrity`,
+        name: ''
```
molant: Missing name here

antross:
Updated with my full list of proposals.

molant:
Added some proposals myself. For the empty ones I like what @antross proposes.

antross:
Once we're in agreement on the names I'll update the PR (including the headers in each `README`).

molant:
@antross I think we've left enough time for people to chime in. I think we should go ahead and update the titles. The ones I feel more strong about are the `aXe` and `webpack` ones. The others I don't mind if we use one or another. Do you want to flip a coin for those or should we fight with knives? 😝

antross:
I’m fine with all of your proposed changes on top of my original list. I’ll update the PR tomorrow so we can move this forward.

packages/hint-http-compression/src/hint.ts
```diff
@@ -11,18 +11,17 @@ import { promisify } from 'util';
...
+import { FetchEnd, IAsyncHTMLElement, IHint, NetworkData, Response, HttpHeaders } from 'hint/dist/src/lib/types';
 import { asyncTry } from 'hint/dist/src/lib/utils/async-wrapper';
 import { getFileExtension, isTextMediaType } from 'hint/dist/src/lib/utils/content-type';
 import getHeaderValueNormalized from 'hint/dist/src/lib/utils/network/normalized-header-value';
 import isHTTP from 'hint/dist/src/lib/utils/network/is-http';
 import isRegularProtocol from 'hint/dist/src/lib/utils/network/is-regular-protocol';
 import normalizeString from 'hint/dist/src/lib/utils/misc/normalize-string';
+
 import { CompressionCheckOptions } from './types';
+import meta from './meta';
```
sarvaje: Nit: You have an empty line before `import meta from './meta';` everywhere but here hehehe.

molant:
@sarvaje can you update the scripts of three sure while you are at it or do you want me to do it and will freak with the conflicts later?

sarvaje:
> freak with the conflicts later?

hahahaha

I'm working on the documentation part in the web, so I can do it.

antross:
> Right now, to generate the website we are reading the `index.ts` to look for the category of a hint.

@sarvaje Are we actually loading the raw source and doing a string search, or are we using `import`/`require` and accessing the static `meta` property on the class? If the latter, this should still work as I import the refactored meta information and continue to expose it via the old static `meta` property.

Also, where are these scripts located? I'd like to take a look at them.

molant:
> Also, where are these scripts located? I'd like to take a look at them.

In the `helpers` folder, the one related to this is [`updater.js`](https://github.com/webhintio/webhint.io/blob/master/helpers/updater.js)

sarvaje:
> Also, where are these scripts located? I'd like to take a look at them.

The scripts are in the folder `helper` as @molant says, but the file is in `update-content.js`.

> @sarvaje Are we actually loading the raw source and doing a string search, or are we using import/require and accessing the static meta property on the class? If the latter, this should still work as I import the refactored meta information and continue to expose it via the old static meta property.

We are doing the first one, we have a regex to do a string search.

The file `update-content`, right now, downloads the repository instead of use the downloaded package.
I think that using import is not the best option, because to do so, we need to install all the packages one by one, and that is going to be prone to errors.

PR #1484, opened 2018-11-08
Fix/vscode

### LINKED ISSUE #1437: VS Code extension errors with `Unable to start webhint...` even if everything is installed
Especially if `hint` is installed globally, the VS Code extension won't work complaining: `Unable to start webhint. Ensure you are using the latest version of the hint and @hint/configuration-development packages`.

Even when the global installs are removed, and everything that is needed is installed locally, it doesn't always work (e.g.: I think I manage to make it work by reinstalling everything locally and restarting VS Code).

### LINKED ISSUE #1482: Prevent `canvas` from being loaded when using the VS Code extension
The biggest problem right now with the VS Code extension is related to `canvas`. `canvas` is a binary package. VS Code uses node 8.9 but a developer could be using any other version and thus cause the extension to fail loading (see https://github.com/jsdom/jsdom/issues/2410).

After talking with @antross and @jdalton, the idea is to hijack `require`'s cache and serve a fake canvas module, enough for jsdom to not crash and continue the execution while running on the VS Code extension.

molant:
## Pull request checklist

Make sure you:

Fix the most common issues with the VS Code extension without having to modify the VS Code extension 🤯

Fix #1437 
Fix #1482

### DIFF
--- a/packages/extension-vscode/src/server.ts
+++ b/packages/extension-vscode/src/server.ts
@@ -37,18 +37,20 @@ const hasYarnLock = (directory: string): Promise<boolean> => {
 };
 
 // Adds webhint and configuration-development to the current workspace.
-const installWebhint = (): Promise<void> => {
+const installWebhint = ({ global }: { global: boolean }): Promise<void> => {
     return new Promise(async (resolve, reject) => {
         connection.sendNotification(notifications.showOutput);
 
         // Build the installation commands.
         const packages = 'hint @hint/configuration-development';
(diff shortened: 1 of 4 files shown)


molant:
It will be cool if someone could test this in their machines. It's a bit cumbersome because you will have to replace files in your global install so maybe we could do it together to make sure this is working correctly outside my machine. @antross, are you in (@sarvaje and @alrra are in the other side of the globe at the moment)?

PR #1486, opened 2018-11-09
Fix links to Apache and IIS pages

DavidDeSloovere:
Hope it's the right syntax.

### DIFF
--- a/packages/hint/docs/user-guide/server-configurations/summary.md
+++ b/packages/hint/docs/user-guide/server-configurations/summary.md
@@ -11,5 +11,5 @@ get started:
 
 <!-- Link labels: -->
 
-[apache config]: ./apache/
-[iis config]: ./iis/
+[apache config]: ./../apache/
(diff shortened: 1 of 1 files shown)


molant:
Hi @DavidDeSloovere,

Thanks for your PR!
The reason those links seem broken is because in the website the links need to be like that so unfortunately we can't merge this change 😞. If you go to https://webhint.io/docs/user-guide/server-configurations/summary/ you will see they link to the right place (`./apache/` and `./iis/`).

It will be good to improve [`update-content.js`](https://github.com/webhintio/webhint.io/blob/master/helpers/update-content.js) so we can have valid links in the markdown and the website.
Do you mind opening an issue in [webhintio/webhint.io](https://github.com/webhintio/webhint.io/issues/new)?

molant:
There's [an issue](https://github.com/webhintio/hint/issues/1518) to fix this now.

PR #1503, opened 2018-11-14
Fix: Ignore status codes for dns-prefetch and preconnect links

chrisguttandin:
## Pull request checklist

Make sure you:

This pull request ensures that status codes are ignored when checking an URL of a [dns-prefetch](https://www.w3.org/TR/resource-hints/#dfn-dns-prefetch) or [preconnect](https://www.w3.org/TR/resource-hints/#dfn-preconnect) link. The spec for [Resource Hints](https://www.w3.org/TR/resource-hints) says that those type of links should only be used for DNS lookups, TCP handshakes, and optional TLS negotiations. Therefore I think checking the resource but ignoring the status code it returns is a good approximation to check those links.

I used https://example.com/404 in the tests to simulate a 404 response. But dns-prefetch and preconnect links should normally be used without any path in the URL. Is there a way to mock a server which returns a 404 response when receiving a request without a path in the tests?

Please let me know if there is anything I should change.

### DIFF
--- a/packages/hint-no-broken-links/src/hint.ts
+++ b/packages/hint-no-broken-links/src/hint.ts
@@ -203,7 +203,17 @@ export default class NoBrokenLinksHint implements IHint {
                     return requester
                         .get(fullURL)
                         .then((value: NetworkData) => {
-                            return handleSuccess(value, fullURL, element);
+                            if (element.nodeName !== 'LINK') {
+                                return handleSuccess(value, fullURL, element);
+                            }
+
+                            const relAttribute = element.getAttribute('rel');
+
+                            if (relAttribute !== 'dns-prefetch' && relAttribute !== 'preconnect') {
+                                return handleSuccess(value, fullURL, element);
+                            }
+
+                            return undefined;
                         })
                         .catch((error: any) => {
                             return handleRejection(error, fullURL, element);
--- a/packages/hint-no-broken-links/src/hint.ts
+++ b/packages/hint-no-broken-links/src/hint.ts
@@ -86,6 +86,16 @@ export default class NoBrokenLinksHint implements IHint {
             return context.report(url, 'Broken link found (domain not found).', { element });
         };
 
+        const isDNSOnlyResourceHint = (element: IAsyncHTMLElement): boolean => {
+            if (element.nodeName !== 'LINK') {
+                return false;
+            }
+
+            const relAttribute = element.getAttribute('rel');
+
+            return (relAttribute === 'dns-prefetch' || relAttribute === 'preconnect');
+        };
+
         /**
          * The callback to handle success handler returned from the `head` method
          * We will check the response status againist the brokenStatusCodes list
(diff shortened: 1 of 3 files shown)


molant:
Thanks a lot for this PR @chrisguttandin !

> Is there a way to mock a server which returns a 404 response when receiving a request without a path in the tests?

Yes, you can do something like this:

```ts
const serverConfig = {
    '/': 'page content goes here...',
    '/404': {
        content: '',       
        status: 404
    }
};
```
There's a bit more details on how to configure the test server in [Test Rules#Serverconfig](https://webhint.io/docs/contributor-guide/how-to/test-rules/#serverconfig).

chrisguttandin:
Hi @molant, thanks for the quick reply. If I understand that correctly your example would create a local server which responds with a 404 to /404. To test the 100% correct behavior I would need to create something with a different origin like https://example.com which responds with anything but 200 to a request to / (without any path).

https://fonts.gstatic.com/ is such a URL and is actually the reason why I discovered this bug. Would it be okay to use it in the tests?

packages/hint-no-broken-links/src/hint.ts
```diff
@@ -203,7 +203,17 @@ export default class NoBrokenLinksHint implements IHint {
                     return requester
                         .get(fullURL)
                         .then((value: NetworkData) => {
-                            return handleSuccess(value, fullURL, element);
+                            if (element.nodeName !== 'LINK') {
```
molant: I think I'll move this logic to `handleSuccess` because it's there when we decide if something should be reported or not. What do you think of something similar to:

```ts
const isResourceHint = (element: IAsyncHTMLElement): boolean => {
   if (element.nodeName !== 'LINK') {
     return false;
  }

  // More magic here
}

const handleSuccess = (networkData: NetworkData, url: string, element: IAsyncHTMLElement) => {
   if(isResourceHint(element)) {
    return Promise.resolve();
  }

  const statusIndex = brokenStatusCodes.indexOf(
    networkData.response.statusCode
  );

  if (statusIndex > -1) {
    return context.report(url, element, `Broken link found (${brokenStatusCodes[statusIndex]} response).`);
  }

  fetchedURLs.push({ status: networkData.response.statusCode, url });
    return Promise.resolve();
  };
```

chrisguttandin: Yes, I think it is a good idea. Having all the code in the promise chain looks very messy.

chrisguttandin:
I updated the tests to use localhost instead. I think it's fine to use an URL with a path as well. But I at least wanted to mention that it is not 100% correct. :-)

I refactored the decision logic as you proposed it. I just used a more specific name for the function. It's called `isDNSOnlyResourceHint()`. I hope I got the capitalization correct this time.

chrisguttandin:
Hi @antross, thanks for your review. I've addressed your feedback and added the tests and updated the README.

molant:
@chrisguttandin I've been working on it for the last few days. The most recent issue [has a PR open for it](https://github.com/webhintio/hint/pull/1545). I should be able to publish everything later tonight 🤞

chrisguttandin:
@molant Sorry for being so impatient. It's not urgent on my end. I just thought it might have been forgotten. Many thanks in advance.

molant:
No worries. We should have published it a while ago but run into several unexpected issues 😞
In any case [it's published now](https://www.npmjs.com/package/@hint/hint-no-broken-links). I still have a few more packages to publish before reaching to a new version of the `configuration-` ones so you might have to install this one manually.

PR #1524, opened 2018-11-21
Fix: Add types to tests

### LINKED ISSUE #1517: [Tests] Pass correctly typed objects/methods into sinon spy/stub
## The devDependency [@types/sinon](https://github.com/DefinitelyTyped/DefinitelyTyped) was updated from `5.0.5` to `5.0.6`.

🚨 [View failing branch](https://github.com/webhintio/hint/compare/master...webhintio:greenkeeper%2Fdefault%2F%40types%2Fsinon-5.0.6).

This version is **covered** by your **current version range** and after updating it in your project **the build failed**.

@types/sinon is a devDependency of this project. It **might not break your production code or affect downstream projects**, but probably breaks your build or test tools, which may **prevent deploying or publishing**.

<details>
<summary>Status Details</summary>

- ❌ **continuous-integration/travis-ci/push:** The Travis CI build failed ([Details](https://travis-ci.org/webhintio/hint/builds/456660726?utm_source=github_status&utm_medium=notification)).
</details>

---

<details>
<summary>FAQ and help</summary>

There is a collection of [frequently asked questions](https://greenkeeper.io/faq.html). If those don’t help, you can always [ask the humans behind Greenkeeper](https://github.com/greenkeeperio/greenkeeper/issues/new).
</details>

---

Your [Greenkeeper](https://greenkeeper.io) Bot :palm_tree:

sarvaje:
Fix #1517

@antross I need your help hehehe.

As you can see, there is a lot of `as any`, e.g.:
`t.true((t.context.notifier.notify as any).calledOnce);`

In this case, I have defined `notifier.notify` as `() => any`, so calledOnce doesn't exists.
I think I have to use the type: `SinonSpy`:
```ts
interface SinonSpy<TArgs extends any[] = any[], TReturnValue = any>
        extends Pick<
                SinonSpyCallApi<TArgs, TReturnValue>,
                Exclude<keyof SinonSpyCallApi<TArgs, TReturnValue>, 'args'>
            >
```
But if I do something like:
```ts
type StubbedNotifier = {
    notify: sinon.SinonSpy<any>;
    update: {} | null;
};
```
and then I do:
```ts
const stubbedNotifier: StubbedNotifier = {
    notify() { },
    update: {}
};
```

I get the error:
`Type '() => void' is not assignable to type 'SinonSpy<any,any>'`

I think I'm missing something, but no idea what it is :s

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/connector-local/tests/tests.ts
+++ b/packages/connector-local/tests/tests.ts
@@ -9,8 +9,24 @@ import delay from 'hint/dist/src/lib/utils/misc/delay';
 import asPathString from 'hint/dist/src/lib/utils/network/as-path-string';
 import { getAsUri } from 'hint/dist/src/lib/utils/network/as-uri';
 
-const chokidar = { watch() { } };
-const isFile = { default() { } };
+type IsFile = {
+    default: (filePath: string) => boolean;
+};
+
+type Chokidar = {
+    watch: () => Stream;
+};
+
+const chokidar: Chokidar = {
+    watch() {
+        return new Stream();
+    }
+};
+const isFile: IsFile = {
+    default(filePath: string): boolean {
+        return false;
+    }
+};
 
 mock('hint/dist/src/lib/utils/fs/is-file', isFile);
 mock('chokidar', chokidar);
@@ -26,7 +42,6 @@ test.beforeEach((t) => {
         notify() { },
         on() { }
     };
-    t.context.isFile = isFile;
 });
 
 test.after(() => {
@@ -38,7 +53,7 @@ test.serial(`If target is a file, it should emit 'fetch::start::target' event`, 
 
     const sandbox = sinon.createSandbox();
 
-    sandbox.stub(t.context.isFile, 'default').returns(true);
+    sandbox.stub(isFile, 'default').returns(true);
     sandbox.spy(t.context.engine, 'emitAsync');
 
     const connector = new LocalConnector(t.context.engine as any, {});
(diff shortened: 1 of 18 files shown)


antross:
@sarvaje 

Overall this is going to be tricky to combine `Sinon` with `Context` in a fully typed way. 

The error is because the stub method doesn't match the method signature. For the first level of typing we should define the types directly on the methods themselves. Ultimately these types should match those of the method we're replacing (so they shouldn't include `sinon.SinonSpy`). 

E.g. In this specific example those would be types from the `update-notifier` package:
```ts
const stubbedNotifier = {
    notify(customMessage?: updateNotifier.NotifyOptions) { },
    update: {} as updateNotifier.UpdateInfo
};
```

However, getting the context typed to include the `stub/spy` APIs is still going to be hard as the `Sinon` APIs are dynamically added/removed on the original object (which is invisible to the type system). At the context level reverting to an explicit cast to `any` may be necessary to access `sinon` APIs (though we could explicitly keep track of created stubs in an `Array` or something for restoring). Overall I'd recommend keeping the stubs within individual tests where possible (see below for how this helps). 

In cases where the stub/spy is only used directly within a test case, I'd recommend storing the stub in a local variable and using that directly in the assertions, then we get the types with extra APIs "for free" from `sinon`.

E.g. in this contrived scenario:
```ts
    const notifyStub = sinon.stub(notifier, 'notify');
    ...
    t.true(notifyStub.calledOnce)
```

Hope that helps.

sarvaje:
Thanks @antross I will try that.

sarvaje:
@webhintio/core this is ready to review.

sarvaje:
Adding again the label `work-in-progress` I forgot some `as any`.

sarvaje:
Now, this is ready to review :)

molant:
@antross whenever you have the time take a look and merge (feel free to squash everything).

Thanks!

PR #1536, opened 2018-11-26
Chore: Add tests for extension-browser

### LINKED ISSUE #1497: [extension-browser] Add tests for major components
Tests for background-script, content-script, and devtools panel.

### LINKED ISSUE #1498: [extension-browser] Add automated end-to-end test
Much like a similar task filed for the VS Code extension, we want to have something that actually launches the browser extension "for real" and ensures basic end-to-end functionality is intact.

antross:
## Pull request checklist

Make sure you:

Fix #1497.
Fix #1498.

I've marked this as "in progress" while I validate whether the Puppeteer tests can run in CI and finish getting to full branch coverage for the background script tests.


packages/extension-browser/tests/content-script.ts
```diff
@@ -0,0 +1,221 @@
...
+import { FetchEnd } from 'hint/dist/src/lib/types';
+
+import { Config, Events, Results } from '../src/shared/types';
+
+import { readFixture } from './utils/helpers';
+
+const base = '../src/content-script';
+const paths: { [name: string]: string } = {
+    'web-async-html': `${base}/web-async-html`,
+    connector: `${base}/connector`, // eslint-disable-line
```
molant: What is `eslint` complaining about in here?

packages/extension-browser/tests/content-script.ts
```diff
@@ -0,0 +1,221 @@
...
+        sendFetch(url, html);
+    });
+
+    stubContext(url, html);
+
+    require(paths.webhint);
+
+    const results = await resultsPromise;
+
+    t.is(results.categories.length, 1, 'Restricted to one category');
```
molant: nit: Maybe check that the category is also accessibility?

packages/extension-browser/tests/content-script.ts
antross: The order of the properties not being alphabetical. However they need to be in this order as the lowest dependency must be overridden before modules which depend on it. I added a comment to clarify.

antross:
@molant @sarvaje Can you take another look at this when you get the chance?

packages/extension-browser/src/background-script.ts
```diff
@@ -216,8 +216,8 @@ const webRequestEvents = [
 ];
 
 const webRequestHandlers = webRequestEvents.map((event) => {
```
molant: Comment: Maybe we shouldn't rely that much in infering the types in this kind of situations 🙄

packages/extension-browser/tests/background-script.ts
```diff
@@ -0,0 +1,443 @@
...
+    const globals = stubGlobals();
+    const { browser } = globals;
+
+    const onMessagePromise = awaitListener(sandbox, browser.runtime.onMessage);
+
+    loadBackgroundScript(globals);
+
+    const onMessage = await onMessagePromise;
+
+    t.true(typeof onMessage === 'function', 'Background script provided a listener');
```
molant: IIRC, the `message` parameter should be the reason why the assertion is failing. In this case it should be `Background script did not provide a listener`. This applies to all your tests.

packages/extension-browser/tests/background-script.ts
```diff
@@ -0,0 +1,443 @@
...
+    sandbox.restore();
+});
+
+test('It reloads the target when enabled.', async (t) => {
+    const sandbox = createSandbox();
+    const globals = stubGlobals();
+    const { browser } = globals;
+    const tabId = 5;
+
+    const reloadSpy = sandbox.stub(browser.tabs, 'reload');
```
sarvaje: If you don't need to change the return value or anything, you can use a `spy` instead of an `stub`

packages/extension-browser/tests/background-script.ts
```diff
@@ -0,0 +1,443 @@
...
+});
+
+test('It starts/stops when browser action is clicked.', async (t) => {
+    const sandbox = createSandbox();
+    const globals = stubGlobals();
+    const { browser } = globals;
+    const tabId = 2;
+
+    const onClickedPromise = awaitListener(sandbox, browser.browserAction.onClicked);
+    const onReloadSpy = sandbox.stub(browser.tabs, 'reload');
```
sarvaje: `spy`?

packages/extension-browser/src/background-script.ts
antross: Yeah, that's a fairly complex return value. I agree having an explicit type to describe it would be a good idea. I'll update the PR.

packages/extension-browser/tests/background-script.ts
antross: I did some digging and it's surprisingly hard to find solid recommendations on best practices for this. :)

That said, what I did find suggested omitting the messages when the context was already clear which aligns with what I see most of our other tests doing elsewhere in the project (e.g. for `parser-javascript`). Where I do see us providing messages they appear to match the structure you're suggesting.

I'll update my tests to omit the message where appropriate and in cases like this one I'll switch to describing why the assertion is failing.

packages/extension-browser/tests/background-script.ts
antross: Heh, the variable name even suggests that was my intent. I'll fix it.

antross:
@molant Mind giving this one more (hopefully last) look now that I've addressed your feedback? :)

molant:
@antross ✅

PR #1578, opened 2018-12-13
Fix: Update tsconfig schema

### LINKED ISSUE #1487: Update tsconfig schema
With TypeScript 3.0 there are new configurations (like `composite`). We have to update our schema to the new version (and find a way to stay in sync).

image

sarvaje:
Fix #1487

This PR update the schema.json for typescript and add the logic to update the file automatically every day.

## Pull request checklist

Make sure you:


packages/parser-typescript-config/src/parser.ts
```diff
@@ -39,6 +49,54 @@ export default class TypeScriptConfigParser extends Parser<TypeScriptConfigEvent
...
+
+                if (this.typeAcquisitionExists(res)) {
+                    res.definitions.typeAcquisitionDefinition.properties.typeAcquisition.additionalProperties = false;
+                }
+
+                this.schema = res;
+
+                await writeFileAsync(this.schemaPath, JSON.stringify(res, null, 2));
+            }
+        } catch (e) {
```
molant: There are 2 main action that can fail:

1. `fs.stat` for the local file
1. Download the remote file

I think I'll split the `try/catch` in 2 so we can have better debugging output. They could even be separate functions so we don't have a `try/catch` in this method.

packages/parser-typescript-config/tests/tests.ts
```diff
@@ -34,7 +68,7 @@ test(`If the resource doesn't match the regex, nothing should happen`, async (t)
     sandbox.restore();
 });
 
-test('If the file contains an invalid json, it should fail', async (t) => {
+test.serial('If the file contains an invalid json, it should fail', async (t) => {
```
antross: Any particular reason these were switched to be `serial`?

sarvaje: If you stub the same object in different tests that run in parallel, you get an error because the object is already stubbed.

sarvaje: Anyway, I'm removing some unnecessary `.serial`

antross: Got it. 

Not that you need to do anything now, but for future note you can work around this by creating the stubbed object on a per-test basis. In this case that would also move the proxyquire and subsequent import call into a helper. You can see [`extension-browser/tests/background-script.ts`](https://github.com/webhintio/hint/blob/e1c8c4037ab45a7403ab33c70b61e265446bd632/packages/extension-browser/tests/background-script.ts#L16) for an example along these lines.

molant:
LGTM

PR #1584, opened 2018-12-14
Hint compat api html 

borgitas21:
## This PR implements the HTML hints.  More info webhintio/rfcs#10

To do for HTML part of this task:
- [x] Implement walk feature.
- [x] Use Location into reports
- [x] Don't check more than once the same scenario (cache)

Feature: Which properties or values are deprecated in the HTML
- [x] Check element tags.
- [x] Check element attributes.
- [x] Check global attributes.
- [x] Tests
- [x] Docs

Feature: Which properties or values are not broadly supported in the HTML.
- [x] Check element tags.
- [x] Check element attributes.
- [x] Check global attributes.
- [x] Tests
- [x] Docs


packages/hint-compat-api/src/core/api-hint.ts
```diff
@@ -110,37 +127,37 @@ export default abstract class BaseCompatApiCSS implements IHint {
...
+            needContextMessage = versions.length > 0;
         }
 
-        return this.getNotSupportedFeatureMessage(feature, groupedBrowserSupport);
+        return this.getNotSupportedFeatureMessage(feature, groupedBrowserSupport, needContextMessage);
     }
 
     private getNotSupportedBrowserMessage(feature: FeatureInfo): string {
-        return `${feature.name} was never supported on any of your browsers to support.`;
+        return `${feature.displayableName} is not supported on any of your browsers to support.`;
```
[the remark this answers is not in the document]
borgitas21: I agree, it sounds better.

packages/hint-compat-api/src/core/api-hint.ts
```diff
@@ -1,39 +1,55 @@
...
 import { StyleParse, StyleEvents } from '@hint/parser-css/dist/src/types';
-import { CompatApi, CompatCSS } from './helpers';
-import { FeatureInfo, BrowsersInfo, SupportStatementResult } from './types';
-import { SimpleSupportStatement, VersionValue, SupportBlock, SupportStatement } from './types-mdn.temp';
+import { CompatAPI, CompatCSS, CompatHTML, userBrowsers } from '../helpers';
+import { FeatureInfo, BrowsersInfo, SupportStatementResult, MDNTreeFilteredByBrowsers } from '../types';
+import { SimpleSupportStatement, VersionValue, SupportBlock, SupportStatement, CompatStatement } from '../types-mdn.temp';
+
+import { browserVersions } from '../helpers/normalize-version';
+import { HTMLParse, HTMLEvents } from '../../../parser-html/dist/src/types';
```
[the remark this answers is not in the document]
borgitas21: Do we really need to add postcss? I guess this is a dependency of `@hint/parser-css`. If so, I think it could be redundant.

packages/hint-compat-api/src/core/api-hint.ts
```diff
@@ -1,39 +1,55 @@
...
+    private compatApi: CompatAPI;
+    private compatLibrary: ICompatLibrary;
 
     abstract getFeatureVersionValueToAnalyze(browserFeatureSupported: SimpleSupportStatement): VersionValue;
     abstract isSupportedVersion(browser: BrowsersInfo, feature: FeatureInfo, currentVersion: number, version: number): boolean;
     abstract isVersionValueSupported(version: VersionValue): boolean;
     abstract isVersionValueTestable(version: VersionValue): boolean;
+    abstract getContextualMessage(needContextMessage: boolean): string;
+
+    public constructor(namespaceName: CompatNamespace, context: HintContext<StyleEvents | HTMLEvents>, isCheckingNotBroadlySupported: boolean) {
```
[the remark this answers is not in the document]
borgitas21: Nice suggestion 👍, Thanks!

packages/hint-compat-api/src/core/api-hint.ts
```diff
@@ -1,39 +1,55 @@
...
 
-    public constructor(context: HintContext<StyleEvents>, statusName: CSSFeatureStatus, isCheckingNotBroadlySupported: boolean) {
-        this.compatApi = new CompatApi('css', context, isCheckingNotBroadlySupported);
-        this.compatCSS = new CompatCSS(context, this.testFeatureIsSupported.bind(this));
-        this.statusName = statusName;
+        this.compatApi = new CompatAPI(namespaceName, mdnBrowsersCollection, isCheckingNotBroadlySupported);
+        this.compatLibrary = new libraries[namespaceName](context, this.testFeatureIsSupported.bind(this));
 
-        context.on('parse::end::css', this.onParseCSS.bind(this));
+        context.on(`parse::end::${namespaceName}` as any, this.onParse.bind(this));
```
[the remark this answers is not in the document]
borgitas21: Yes, I was not happy with this implementation. 

I just moved the event listener to each specific library, so there is no need to build eventName.

borgitas21:
@antross after following the steps you mentioned above, I was not able to reproduce the error. 
I'm opening a fixture file but nothing is hapenning, any idea?

> At this point webhint should run automatically and start showing errors in VS Code. Once you see an initial error, make any edit to the file and the ones from hint-compat-api will disappear (but other webhint errors will still remain).

packages/hint-compat-api/src/helpers/compat-base.ts
```diff
@@ -28,16 +28,15 @@ export abstract class CompatBase<T> {
...
+            return;
         }
 
         this.cachedFeatures.add(feature);
+        await this.hintContext.report(this.hintResource, message, { location });
+    }
 
-        return false;
+    public isFeatureAlreadyReported(feature: FeatureInfo): boolean {
+        return this.cachedFeatures.has(feature);
```
borgitas21: This commit should take care of this issue https://github.com/webhintio/hint/issues/1589

antross:
@borgitas21 

Awesome progress on the updates! This is looking really good. I'll take another look later today and let you know if I spot anything else.

> after following the steps you mentioned above, I was not able to reproduce the error.
> I'm opening a fixture file but nothing is hapenning, any idea?

I see I forgot one other addition to `configuration-development/index.json`. You also need to add `css` to the `parsers` list. Here's the full contents of my modified version of the file just in case: 
```json
{
    "connector": "local",
    "extends": [
        "progressive-web-apps"
    ],
    "formatters": [
        "html",
        "summary"
    ],
    "hints": {
        "apple-touch-icons": "off",
        "axe": "error",
        "babel-config/is-valid": "error",
        "compat-api/css": "error",
        "compat-api/css-next": "error",
        "compat-api/html": "error",
        "compat-api/html-next": "error",
        "disown-opener": "error",
        "highest-available-document-mode": "error",
        "manifest-exists": "off",
        "meta-charset-utf-8": "error",
        "meta-viewport": "error",
        "no-bom": "error",
        "no-protocol-relative-urls": "error",
        "typescript-config/consistent-casing": "error",
        "typescript-config/import-helpers": "error",
        "typescript-config/is-valid": "error",
        "typescript-config/no-comments": "error",
        "typescript-config/strict": "error",
        "typescript-config/target": "error",
        "webpack-config/is-installed": "error",
        "webpack-config/is-valid": "error",
        "webpack-config/module-esnext-typescript": "off",
        "webpack-config/modules/false-babel": "error",
        "webpack-config/no-devtool-in-prod": "error"
    },
    "hintsTimeout": 10000,
    "parsers": [
        "babel-config",
        "css",
        "html",
        "typescript-config",
        "webpack-config"
    ]
}
```

If that still doesn't help, in the launched debug instance of VS Code (marked with "[Extension Development Host]" in the title), go to "View" > "Output", then select "webhint" from the dropdown in the "OUTPUT" panel that opens. If you see any errors in there it may give us a clue why webhint isn't running.

packages/hint-compat-api/src/core/deprecated-hint.ts
```diff
@@ -0,0 +1,38 @@
+import { HintContext } from 'hint/dist/src/lib/hint-context';
+import { SimpleSupportStatement, VersionValue } from '../types-mdn.temp';
+
+import meta from '../meta/css';
+import { APIHint } from './api-hint';
+import { CSSFeatureStatus, CompatNamespace } from '../enums';
+import { FeatureInfo, BrowsersInfo } from '../types';
+import { Events, Event } from 'hint/dist/src/lib/types';
```
sarvaje: IIRC, we have a convention (I don't know if it is write anywhere) for the `import`.

1st. Packages that node have by default (`fs`, `path`, etc.).
2nd. "External" dependencies (`hint`, `lodash`, etc.)
3rd. "Local" dependencies (`../types`, `../enum`, etc.)

This would apply to few files.

packages/hint-compat-api/src/helpers/compat-base.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        this.cachedFeatures = new CachedCompatFeatures();
+        this.testFunction = testFunction;
+        this.hintContext = hintContext;
+        this.MDNData = MDNData;
+
+        // Clear feature cache between scans. Contexts like extension-vscode scan multiple times.
+        (this.hintContext as HintContext<Events>).on('scan::end', () => {
+            this.cachedFeatures = new CachedCompatFeatures();
+        });
```
antross: @borgitas21 I figured out why results were disappearing in the VS Code extension - the feature cache was never being cleared/reset. In some contexts we can run multiple scans with the same hint instances, so this state needed cleared to allow reporting issues again on subsequent scans. 

I went ahead and just made the fix, but feel free to refactor or move this if you have a different preference on how this should be done.

packages/hint-compat-api/src/helpers/compat-html.ts
```diff
@@ -0,0 +1,109 @@
...
+
+        hintContext.on('parse::end::html', this.onParse.bind(this));
+    }
+
+    public async searchFeatures(parser: HTMLParse): Promise<void> {
+        await this.walk(async (elementFound: ElementFound) => {
+            const { element } = elementFound;
+            const location = element.getLocation();
+            /*
+             * WORKAROUND: Element listener is being fired twice. The second
```
sarvaje: I'm just curious. Why is the listener fired twice?

antross:
Looks like the `<blink>` tag is getting reported by both our deprecated and next hints. The deprecated hint is catching browsers that supported it at one point whereas the next hint is catching browsers that never supported it. Ideally cases like this would also result in a single message (probably from the deprecated hint as the fact that it was never implemented in some browsers doesn't make it "new").

```
blink element is not supported on firefox 60, firefox 62-63, firefox_android 63, opera 55-56 browsers. (compat-api/html)
blink element is not supported on chrome, chrome_android, edge, ie, safari, safari_ios, samsunginternet_android, webview_android browsers. (compat-api/html-next)
```

packages/hint-compat-api/src/types.ts
```diff
@@ -10,20 +10,22 @@ export type BrowserSupportCollection = {
...
 };
 
-export type TestFeatureFunction = (feature: FeatureInfo, supportBlock: SupportBlock) => void;
+export type TestFeatureFunction = (feature: FeatureInfo, collection: CompatStatement | undefined) => void;
 
 export type BrowserVersions = {
     [key: string]: string[];
 };
 
+// eslint-disable: no-use-before-define, typescript/no-use-before-define
```
sarvaje: Why do you need this?

Anyways, enable the rule as soon it is not necessary.

packages/hint-compat-api/src/helpers/compat-html.ts
borgitas21: To be honest, I didn’t spend too much time investigating  the issue. I will take a look and let you know

packages/hint-compat-api/src/types.ts
borgitas21: As you can see, I’m using the same type in a child property. The linting was warning me because I was using the type before defining it. I guess the rule still makes sense in other scenario, so I just disabled it in this context.

sarvaje: That is weird, you should be able to use the same type you are defining as the type of one of its property without any warning...

packages/hint-compat-api/src/helpers/compat-html.ts
sarvaje: ok!

packages/hint-compat-api/src/types.ts
borgitas21: For some reason, now eslint is ok without disabling the rule. I don't understand why I was getting the warning 😩 . I just removed this line 👍

packages/hint-compat-api/src/core/deprecated-hint.ts
borgitas21: I was not aware of this convention. However I was trying to apply a similar approach but it seems I missed some imports.

borgitas21: Done.

packages/hint-compat-api/src/helpers/compat-html.ts
borgitas21: So after digging into the issue I found the `element::$tagName` event is fired from two different places:
-  `packages\connector-jsdom\src\connector.ts`. It doesn't include the location.
- `packages\parser-html\src\parser.ts`. It includes the location.

I'm not sure if this is expected or not, but it sounds weird to me.

borgitas21:
> Looks like the `<blink>` tag is getting reported by both our deprecated and next hints. The deprecated hint is catching browsers that supported it at one point whereas the next hint is catching browsers that never supported it. Ideally cases like this would also result in a single message (probably from the deprecated hint as the fact that it was never implemented in some browsers doesn't make it "new").
> 
> ```
> blink element is not supported on firefox 60, firefox 62-63, firefox_android 63, opera 55-56 browsers. (compat-api/html)
> blink element is not supported on chrome, chrome_android, edge, ie, safari, safari_ios, samsunginternet_android, webview_android browsers. (compat-api/html-next)
> ```

I'm not 100% sure I'm getting your request. I'm sharing the MDN information for this specific feature. 

As you can see, at some point, the `blind` tag was supported on firefox, this is why we are reporting and issue if you are using a new version of firefox and the feature was deprecated.

For the broadly-supported tests we are indicating this feature was never implemented on these browsers. I still think this report make sense. Maybe I'm missing something.

```json
{
  "html": {
    "elements": {
      "blink": {
        "__compat": {
          "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/blink",
          "support": {
            "chrome": {
              "version_added": false
            },
            "chrome_android": {
              "version_added": false
            },
            "edge": {
              "version_added": false
            },
            "edge_mobile": {
              "version_added": false
            },
            "firefox": {
              "version_added": "1",
              "version_removed": "22"
            },
            "firefox_android": {
              "version_added": "4",
              "version_removed": "22"
            },
            "ie": {
              "version_added": false
            },
            "opera": {
              "version_added": "2",
              "version_removed": "15"
            },
            "opera_android": {
              "version_added": "2",
              "version_removed": "14"
            },
            "safari": {
              "version_added": false
            },
            "safari_ios": {
              "version_added": false
            },
            "samsunginternet_android": {
              "version_added": false
            },
            "webview_android": {
              "version_added": false
            }
          },
          "status": {
            "experimental": false,
            "standard_track": false,
            "deprecated": true
          }
        }
      }
    }
  }
}

```

antross:
> > Looks like the `<blink>` tag is getting reported by both our deprecated and next hints. The deprecated hint is catching browsers that supported it at one point whereas the next hint is catching browsers that never supported it. Ideally cases like this would also result in a single message (probably from the deprecated hint as the fact that it was never implemented in some browsers doesn't make it "new").
> > ```
> > blink element is not supported on firefox 60, firefox 62-63, firefox_android 63, opera 55-56 browsers. (compat-api/html)
> > blink element is not supported on chrome, chrome_android, edge, ie, safari, safari_ios, samsunginternet_android, webview_android browsers. (compat-api/html-next)
> > ```
> 
> I'm not 100% sure if I'm getting your request. I'm sharing the MDN information for this specific feature.
> 
> As you can see, at some point, the `blind` tag was supported on firefox, this is why we are reporting and issue if you are using a new version of firefox and the feature was deprecated.
> 
> For the broadly-supported tests we are indicating this feature was never implemented on these browsers. I still think this report make sense. Maybe I'm missing something.

I'm saying we should treat the `blink` tag (and any other CSS or HTML feature following this pattern) as deprecated even in the browsers that never implemented it. Clearly nobody new is going to implement it going forward. Thus I'd only expect to see one report along the following lines: 
```
blink element is not supported on chrome, chrome_android, edge, firefox 60, firefox 62-63, firefox_android 63, ie, opera 55-56, safari, safari_ios, samsunginternet_android, webview_android browsers. (compat-api/html)
```

packages/hint-compat-api/src/helpers/compat-html.ts
sarvaje: > I'm not sure if this is expected or not, but it sounds weird to me.

That sounds like you are using both in your configuration file, the connector `jsdom` and the html parser.

IIRC, the html parser is mean to be use just with the `local` connector, @antross am I right?

molant: We should do something to avoid this:
- Make the configuration fail (not a huge fan)
- Make the parser smarter and disable itself if it receives one of the events it's supposed to send?
- Something else?

borgitas21: FYI: If we go for the second option (it seems smarter), we should take into account the event sent from the connector is not populating the location

borgitas21:
> > ```
> > blink element is not supported on chrome, chrome_android, edge, ie, safari, safari_ios, samsunginternet_android, webview_android browsers. (compat-api/html-next)
> > ```

After discussion with @CKGrafico, probably the issue comes from a bug in the css-next. Some reported messages were using the `supported` instead of `added` word so we were getting the same report for different issues. 

Furthermore, we are replacing `supported` with `deprecated`. It seems a bit more specific and clear for the end-user. What do you think?

If you still think we must merge the reports, we should implement something "new" to handle this specific scenario. Please let me know your thoughts.

packages/hint-compat-api/src/helpers/compat-html.ts
borgitas21: In any case, I guess this issue is not related with the compat-api-html implementation. Maybe we should deal with this in a different branch.

antross:
> If you still think we must merge the reports, we should implement something "new" to handle this specific scenario. Please let me know your thoughts.

Yes, I think we must "merge" the reports to provide a single report about deprecation coming from `compat-api/html`. 

To be clear, at an implementation level I don't really see this as a "merge", but rather a change in conditions to treat browsers that never implemented a *deprecated* feature as part of the "deprecated" set (`compat-api/html`, `compat-api/css`) instead of the "not broadly supported" set (`compat-api/html-next`, `compat-api/css-next`).

packages/hint-compat-api/src/helpers/compat-html.ts
antross: > Maybe we should deal with this in a different branch.

Agreed. Glad we found it, though. For this PR, you should be able to fix your local configuration and remove the workaround. I've filed #1593 to track fixing `parser-html` to prevent this.

antross: > FYI: If we go for the second option (it seems smarter), we should take into account the event sent from the connector is not populating the location

Also agreed. The `report` API will "infer" a location if passed an `element` (one of the optional parameters). In fact behind the scenes this calls `element.getLocation` first before trying to infer the location. 

Some of this is for historical reasons and it may be better at some point to extract this inference into a helper and update connectors to use it as a fallback in `element.getLocation` when no parser-provided location is known. However, this is how things work today.

borgitas21: In order to skip the `jsdom` connector I had to modify the `connectors.ts` file removing the `jsdom`option. But this is not a custom configuration, this is the default configuration in the repo.

```typescript
packages\utils-tests-helpers\src\connectors.ts

export const ids = [
    'chrome',
    'jsdom'
];
```

After this update I can confirm the workaround is not needed anymore.

@antross About your last message.

> You should be able to fix your local configuration and remove the workaround.

I'm not sure if there is a way to fix my local configuration without updating any file in the repo. Am I missing something?

packages/hint-compat-api/tests/html-next.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+
+const elementAddedAlwaysTrue: HintTest[] = [
+    {
+        name: 'Elements that have added as true should pass.',
+        serverConfig: generateHTMLConfig('div')
+    }
+];
+
+hintRunner.testHint(hintPath, elementAddedAlwaysTrue, { browserslist: ['last 2 Edge versions'], parsers: ['html']});
```
antross: Remove the explicit HTML parser from this and all other tests and the double-events issue (one from `parser-html` and one from the connector) should go away. We don't need the HTML parser here because both connectors we use in the test suite (`jsdom` and `chrome`) will generate `element::*` events on their own.

```suggestion
hintRunner.testHint(hintPath, elementAddedAlwaysTrue, { browserslist: ['last 2 Edge versions']});
```

packages/hint-compat-api/src/helpers/compat-html.ts
antross: > I'm not sure if there is a way to fix my local configuration without updating any file in the repo.

I saw the issue and sorry I missed it in my earlier reviews. We need to exclude specifying `parsers: ['html']` in the configuration for the test cases to fix this (it's not needed). I added a review suggestion on one of the tests so you can see exactly what I mean.

sarvaje: >In order to skip the jsdom connector I had to modify the connectors.ts file removing the jsdomoption. But this is not a custom configuration, this is the default configuration in the repo.

> I saw the issue and sorry I missed it in my earlier reviews. We need to exclude specifying parsers: ['html'] in the configuration for the test cases to fix this (it's not needed). I added a review suggestion on one of the tests so you can see exactly what I mean.

Maybe this is not the case, but just in case @borgitas21 doesn't know it. You can configure your tests to skip one connector or another, add parsers, etc. for example, here we are ignoring the `chrome` connector: https://github.com/webhintio/hint/blob/master/packages/hint-http-compression/tests/tests-http.ts#L23

packages/hint-compat-api/package.json
```diff
@@ -8,6 +8,8 @@
   },
   "description": "hint to validate if the HTML, CSS, and JavaScript APIs of the project are deprecated or not broadly supported",
   "devDependencies": {
+    "@hint/parser-css": "^2.0.2",
+    "@hint/parser-html": "^2.0.1",
```
sarvaje: Do we still need the `parser-html`?

antross: We shouldn't need it after the next set of updates because we won't be using the `HTMLEvents` and `HTMLParse` types anymore. Just plain old `element::*` events...

packages/hint-compat-api/src/helpers/compat-html.ts
borgitas21: Cool 😀. Thank you guys!

packages/hint-compat-api/src/helpers/compat-html.ts
```diff
@@ -0,0 +1,100 @@
...
+
+import { MDNTreeFilteredByBrowsers, TestFeatureFunction, FeatureInfo } from '../types';
+import { CompatStatement } from '../types-mdn.temp';
+import { CompatBase } from './compat-base';
+
+export class CompatHTML extends CompatBase<HTMLEvents, HTMLParse> {
+    public constructor(hintContext: HintContext<HTMLEvents>, MDNData: MDNTreeFilteredByBrowsers, testFunction: TestFeatureFunction) {
+        super(hintContext, MDNData, testFunction);
+
+        hintContext.on('parse::end::html', this.onParse.bind(this));
```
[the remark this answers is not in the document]
borgitas21: Ok, I just updated everything to remove the `html-parser` dependency. As expected, a bunch of tests are failing because the missing location issue. Should I apply any workaround to handle this situation or wait for your changes in master?

packages/hint-compat-api/package.json
borgitas21: Done.

packages/hint-compat-api/src/helpers/compat-html.ts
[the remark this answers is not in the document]
antross: I dig some digging and realized the helper `context.report` uses to resolve a location given an `element` is, in fact, public. That made this a pretty straightforward fix to swap out the current `element.getLocation` call with `hintContext.findProblemLocation(element)` (which uses `getLocation` internally).

I validated this fixed the tests (though I also found and fixed an issue with how one of the other helpers was being used) and updated the PR with the fix.

packages/hint-compat-api/tests/css-next.ts
```diff
@@ -239,6 +239,20 @@ const mixedFeaturedCompatibility: HintTest[] = [
 
 hintRunner.testHint(hintPath, mixedFeaturedCompatibility, { browserslist: ['firefox 63', 'edge 18'], parsers: ['css']});
 
+const notSupportedFeaturesShouldNotSeparatelyLog: HintTest[] = [
+    {
+        name: 'Features not supported should not separately log the feature and value.',
+        reports: [
+            { message: 'appearance prefixed with -webkit- is not supported on ie browser.', position: { column: 4, line: 1 }},
+            { message: 'appearance prefixed with -moz- is not supported on ie browser.', position: { column: 4, line: 2 }},
+            { message: 'appearance is not supported on ie browser.', position: { column: 4, line: 3 }}
```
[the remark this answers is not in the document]
antross: @borgitas21 Looks like this is still happening and the `compat-api/css` hint is logging these same reports in addition to the `compat-api/css-next` hint. As this is not a deprecated feature, only `compat-api/css-next` should be reporting it.

Here's the log I get in the "Problems" panel when opening the `hint-compat-api/tests/fixtures/css/appearance.css` in VS Code using our extension:
```
[webhint] appearance prefixed with -webkit- is not supported by ie. (compat-api/css)
[webhint] appearance prefixed with -webkit- is not supported by ie. (compat-api/css-next)
[webhint] appearance prefixed with -moz- is not supported by ie. (compat-api/css)
[webhint] appearance prefixed with -moz- is not supported by ie. (compat-api/css-next)
[webhint] appearance is not supported by ie. (compat-api/css)
[webhint] appearance is not supported by ie. (compat-api/css-next)
```

The expected result is just one report:
```
[webhint] appearance is not supported by ie. (compat-api/css-next)
```

FWIW I'm okay tracking this as a separate issue for a follow-up PR if we want to get this PR wrapped up and merged. That said, I do think we should fix this before turning the hint on-by-default.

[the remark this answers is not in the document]
antross: Likely related, missing support for the `integrity` attribute is getting flagged by both `compat-api/html` and `compat-api/html-next` (only the latter should report this).

Note that the list of browsers differs between these as well, which seems like a separate bug.

Actual:
```
[webhint] integrity attribute of the link element is not supported by edge, ie, safari, safari_ios. (compat-api/html)
[webhint] integrity attribute of the link element is not supported by edge, ie, safari, safari_ios, samsunginternet_android 4, webview_android 4.4.3-4.4.4. (compat-api/html-next)
```

Expected:
```
[webhint] integrity attribute of the link element is not supported by edge, ie, safari, safari_ios, samsunginternet_android 4, webview_android 4.4.3-4.4.4. (compat-api/html-next)
```

packages/hint-compat-api/src/helpers/compat-html.ts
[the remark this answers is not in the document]
borgitas21: 👍

packages/hint-compat-api/tests/css-next.ts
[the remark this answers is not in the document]
borgitas21: @antross I think this is getting in conflict with:

> Yes, I think we must "merge" the reports to provide a single report about deprecation coming from compat-api/html.
>
> To be clear, at an implementation level I don't really see this as a "merge", but rather a change in conditions to treat browsers that never implemented a deprecated feature as part of the "deprecated" set (compat-api/html, compat-api/css) instead of the "not broadly supported" set (compat-api/html-next, compat-api/css-next).

Previously, we moved the "never implemented features" to the deprecated hint. If I understood correctly, now we want to move it to the next hint. Anyway, I found an issue because we were still handling never implemented features by the next hint in some scenarios. However we still get two different reports. Am I missing something? Could you bring some light to the subject?

```
[webhint] integrity attribute of the link element is not supported by edge, ie, safari, safari_ios. (compat-api/html)
[webhint] integrity attribute of the link element is not supported by samsunginternet_android 4, webview_android 4. (compat-api/html-next)
```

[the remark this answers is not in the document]
borgitas21: Neat 👏 . Thanks! Please, let me know your feedback after my last change.

About the previous issue, I'm aware of it, but yes, please, let's tackle it in an specific issue since this is not part of this development and I don't want to delay more this PR.

packages/hint-compat-api/package.json
```diff
@@ -8,8 +8,7 @@
   },
   "description": "hint to validate if the HTML, CSS, and JavaScript APIs of the project are deprecated or not broadly supported",
   "devDependencies": {
-    "@hint/parser-css": "^2.0.2",
```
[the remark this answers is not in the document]
antross: FYI the current version is `2.0.3`.

borgitas21:
> Looks like the deprecated distinction is partly working, but a few non-deprecated CSS properties are still getting reported by both the css and css-next hints (e.g. cursor, resize, text-decoration-color).

@antross Good catch! I forgot to check the deprecated flag when testing the feature and not just when filtering testable features.

borgitas21:
Great. Thanks, @antross! Hopefully https://github.com/webhintio/hint/issues/1598 will be ready soon 😃

packages/hint-compat-api/src/helpers/compat-api.ts
```diff
@@ -107,15 +103,13 @@ export class CompatApi {
 
         // Sometimes the API returns an array but only the first seems relevant
         if (Array.isArray(currentBrowserFeatureSupported) && currentBrowserFeatureSupported.length > 0) {
-            if (prefix) {
-                currentBrowserFeatureSupported = currentBrowserFeatureSupported.find((info) => {
-                    return info.prefix === prefix;
+            currentBrowserFeatureSupported = currentBrowserFeatureSupported
```
sarvaje: Why are you using `filter` and `find`? I think you should be able to do everything inside `find`.

packages/hint-compat-api/src/helpers/compat-api.ts
```diff
@@ -52,34 +51,31 @@ export class CompatApi {
...
-                }
-
-                typedFeatures[childKey] = childValue;
+    private getFeaturesAndChildrenRequiredToTest(featureValue: CompatStatement & MDNTreeFilteredByBrowsers): CompatStatement {
+        const typedFeatures = {} as CompatStatement & MDNTreeFilteredByBrowsers;
 
-                return true;
-            });
+        if (typeof featureValue === 'object' && Object.keys(featureValue).length) {
+            Object.entries(featureValue as object)
```
sarvaje: Why `filter` and then `forEach`? In the worst case, you are iterating twice the same collection.
Why don't do everything in the `forEach`?

packages/hint-compat-api/src/helpers/compat-api.ts
borgitas21: In both cases, I was trying to increase the readability of the code splitting the responsibility per operator. Of course, this is against performance. Probably, in that case, the performance is key so I will remove one of the operators.

borgitas21: I closed the previous conversation since the issue seems similar.

antross:
@sarvaje Any other comments now that this PR has been updated?

sarvaje:
> @sarvaje Any other comments now that this PR has been updated?

Nop. LGTM

PR #1592, opened 2018-12-19
package.json parser

borgitas21:
## Pull request checklist

Make sure you:

Implemented parser for package.json files. This is related to hint-compat-api-css when the issue was brought up to look for autoprefixer being used by checking in the project's package.json.


packages/parser-package-json/src/parser.ts
```diff
@@ -0,0 +1,77 @@
...
+import { PackageJsonEvents } from './types';
+
+export * from './types';
+
+export default class PackageJsonParser extends Parser<PackageJsonEvents> {
+    private schema: any;
+
+    public constructor(engine: Engine<PackageJsonEvents>) {
+        super(engine, 'package-json');
+        this.schema = loadJSONFile(path.join(__dirname, 'schema.json'));
```
antross: Where did we get the schema definition from? A comment would be useful here so we know where to look when we need to update it.

packages/parser-package-json/src/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type PackageJsonInvalidJSON = ErrorEvent;
+
+/** Data type sent when the parse starts parsing */
+export type PackageJsonParseStart = Event;
+
+/** The object emitted by the `package-json` parser */
+export type PackageJsonParsed = Event & {
+    /** The package json parsed */
+    config: any;
```
antross: We haven't done this before, but want to try something like [json-schema-to-typescript](https://github.com/bcherny/json-schema-to-typescript) to automatically create some TypeScript types to reference here instead of `any`?

We've chatted about going in the other direction in the past in webhintio/rfcs#43, but I think since we have a number of externally sourced schemas, using the schema itself as the "source of truth" may be better.

packages/parser-package-json/src/parser.ts
antross: Also, does this schema allow additional custom properties at the root? I know a number of packages have taken to extending package.json (e.g. [browserslist](https://github.com/browserslist/browserslist#browserslist-)) and I don't think we want to fail validation in that case.

packages/parser-package-json/src/parser.ts
```diff
@@ -0,0 +1,77 @@
...
+        const resourceFileName = path.basename(resource);
+        const isPackageJson: boolean = resourceFileName === 'package.json';
+
+        if (!isPackageJson) {
+            return;
+        }
+
+        try {
+            const response = fetchEnd.response;
+            // When using local connector to read local files, 'content' is empty.
```
antross: Is this comment left-over from something? I don't think it's true and I don't see any code that attempts to handle empty content either.

packages/parser-package-json/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "lint:md": "node ../../scripts/lint-markdown.js",
+    "test": "npm run lint && npm run build && npm run test-only",
+    "test-only": "nyc ava",
+    "test-release": "npm run lint && npm run clean && npm run build:assets && tsc && npm run test-only",
+    "watch": "npm run build && npm-run-all --parallel -c watch:*",
+    "watch:assets": "npm run build:assets -- -w --no-initial",
+    "watch:test": "ava --watch",
+    "watch:ts": "npm run build:ts -- --watch"
+  },
+  "version": "2.0.0"
```
antross: Also need to reset the version for the new package.
```suggestion
  "version": "1.0.0"
```

packages/parser-package-json/src/parser.ts
borgitas21: AFAIK JsonSchema allow you to validate properties defined in the schema, if you add some extra properties not considered in the schema, these properties won't be validated against the schema, so the json should still be valid..

isabellachen: @antross There is a test that checks for this - line 132 in tests/tests.ts `If 'package.json' contains a non standard property, it should still emit the event parse::end::package-json`

isabellachen: I've updated the test using the browserslist example...

antross: Great, thanks for clarifying!

packages/parser-package-json/src/schema.ts
```diff
@@ -0,0 +1,308 @@
+/**
+ * This file was automatically generated by json-schema-to-typescript.
+ * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,
+ * and run json-schema-to-typescript to regenerate this file.
```
antross: Very nice! I played around with this a bit in VS Code and really liked getting the intellisense suggestions when accessing the config object.

Since this file is auto-generated, let's add a (pre-)build step to generate it from the current `schema.json` and a local `.gitignore` to exclude the generated file from source control.

isabellachen: The auto-generated schema did have an error which I had to fix by hand (against orders) -(scripts prop on generated schema for package.json needs to be fixed)[https://github.com/bcherny/json-schema-to-typescript/issues/210]  ... not sure how to proceed ...

isabellachen: Looks like editing the schema.json causes the ajv validator to  not work as expected, resulting in valid `package.json` configurations of fail the tests. I've open an issue to track the auto-generation of the TS schema from JSON schema at [Add pre-build step to auto-generate Typescript schema from JSON schema #1594](https://github.com/webhintio/hint/issues/1594)

borgitas21: @antross I guess, that was the only thing we were missing to close the PR, right?

packages/parser-package-json/README.md
```diff
@@ -0,0 +1,72 @@
...
+    "hints": {
+        ...
+    },
+    "parsers": ["package-json"],
+    ...
+}
+```
+
+This parser detects if a `package.json` file is present in the project and
+validates its configuration against the the `package.json` schema.
```
sarvaje: `the` is repeated.

packages/parser-package-json/README.md
```diff
@@ -0,0 +1,72 @@
...
+  which contains the following information:
+
+  * `resource`: the parsed resource.
+  * `error`: the error emited parsing the package.json.
+
+* `parse::error::package-json::schema`, of type `PackageJsonInvalidSchema`
+  which contains the following information:
+
+  * `resource`: the parsed resource.
+  * `errors`: all the errors that the schama validator returns.
```
sarvaje: There is more information that we are passing here in this event.

packages/parser-package-json/README.md
```diff
@@ -0,0 +1,72 @@
...
+* `parse::start::package-json`, of type `PackageJsonParseStart`
+  which contains the following information:
+
+  * `resource`: the resource we are going to parse.
+
+* `parse::end::package-json`, of type `PackageJsonParsed`
+  which contains the following information:
+
+  * `resource`: the parsed resource.
+  * `config`: an object with a valid configuration.
```
sarvaje: `getLocation` is missed.

isabellachen:
The docs have been updated.

antross:
Actually merged via 6a9ac06927b756cbaa17250a33a07adee3e44ff5 (had to fix the commit message).

PR #1599, opened 2019-01-02
Issues/browserslist config

borgitas21:
## Pull request checklist

Make sure you:


packages/hint/src/lib/types.ts
```diff
@@ -73,6 +73,15 @@ export type UserConfig = {
...
 };
 
+export type ConfigFiles = {
+    '.hintrc'? : string | string[];
+    '.hintrc.js'?: string | string[];
+    'hintrc.json'?: string | string[];
+    'package.json'?: string | string[];
+    'hintConfig'?: string | string[];
+    [k: string]: string | string[] | undefined;
+};
```
borgitas21: I would reuse the ConfigFile type so we can avoid duplicating the config names. What do you think?
```typescript
export type ConfigFiles = {
    [key in ConfigFile]?: string | string[]
};
```

packages/hint/tests/lib/config.ts
```diff
@@ -53,6 +54,93 @@ test.afterEach((t: TestContext) => {
...
+//     sandbox
+//         .stub(process, 'cwd')
+//         .returns(path.join(__dirname, './fixtures/browserslist-valid-package-json'));
+
+//     const result = config.Configuration.loadBrowsersList();
+
+//     t.is(result.length, 2);
+//     t.true(isEqual(result, jsonConfig.browserslist));
+// });
+
```
borgitas21: 👀

packages/hint/src/lib/config.ts
```diff
@@ -253,9 +251,49 @@ export class Configuration {
...
+
+        return configs;
+    }
+
+    private static browserslistConfigError (configKeys: string[]): void {
+
+        const fileNamesJoined = configKeys.length > 2 ?
+            `${configKeys.slice(0, configKeys.length - 1).join(', ')} and ${configKeys[configKeys.length - 1]}`:
+            configKeys.join(' and ');
+
```
borgitas21: If we assume this is receiving more than one configKey, I think you don't need the ternary operator.

isabellachen:
This is ready for review @molant @antross @sarvaje @alrra

packages/hint/src/lib/config.ts
```diff
@@ -253,17 +252,55 @@ export class Configuration {
...
+        if (!basename) {
+            throw new Error(`unrecognised config file '${basename}'`);
+        }
+
+        const hintConfig = Configuration.loadConfigFile(file);
+
+        if (basename === ConfigFile.PackageJson) {
+
+            const packageJson = loadJSONFile(file);
```
antross: Since `Configuration.loadConfigFile` already handles loading `package.json` I'm confused as to why this separate load is needed. Is there a particular reason this needs to be special-cased that I'm missing?

packages/hint/src/lib/config.ts
```diff
@@ -253,17 +252,55 @@ export class Configuration {
...
+
+            if (packageJson.browserslist) {
+                configs[basename] = packageJson.browserslist;
+            }
+
+            if (hintConfig && hintConfig.browserslist) {
+                configs.hintConfig = hintConfig.browserslist;
```
antross: What does this do?

packages/hint/src/lib/config.ts
isabellachen: The way I understand it, it's possible to declare browserslist in both the root of the package.json and in the hintConfig prop in the package.json. So I'm trying to catch the case if browserslist has been declared twice in package.json, once in the root and once in the hintConfig prop.

packages/hint/src/lib/config.ts
isabellachen: It gets the value in the user's hintConfig property declared in the package.json file.

packages/hint/src/lib/config.ts
```diff
@@ -253,17 +252,55 @@ export class Configuration {
         return browserslist(config.browserslist);
     }
 
+    private static groupHintConfigs(configs: ConfigFiles, file: string): ConfigFiles {
+        const basename = path.basename(file) as ConfigFile;
+
+        if (!basename) {
+            throw new Error(`unrecognised config file '${basename}'`);
```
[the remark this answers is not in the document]
isabellachen: The problem is if I remove the 'as ConfigFile' constraint, I get an error on line 282 'Element implicitly has an 'any' type because type 'BrowsersListConfigs' has no index signature.' I've removed the constraint and instead placed an index signature on the BrowsersListConfigs type.

packages/hint/src/lib/config.ts
isabellachen: Since `loadConfigFile()` gets the `hintConfig` value and `loadJsonFile` gets the contents of the entire package.json.

packages/hint/docs/user-guide/configuring-webhint/browser-context.md
```diff
@@ -1,8 +1,8 @@
 # Browser configuration
 
 `webhint` allows you to define what browsers are relevant to your
-scenario by adding the property `browserslist` to your `.hintrc`
-file, or in the `package.json` file. This property follows the same
+scenario by adding the property `browserslist` to either your project's `.hintrc`
```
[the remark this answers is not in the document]
antross: As far as populating the list goes, `browerslist` should do this automatically as long as we don't pass a configuration when one isn't specified (see the discussion at https://github.com/webhintio/rfcs/issues/59#issuecomment-437077515).

That said, this PR also adds checks to validate there aren't multiple definitions of `browserslist` across config files (e.g. one in `.hintrc` and one in `package.json`). Currently that only covers the files where webhint itself can also be configured. 

As far as the validation checks are concerned, going beyond the current list probably requires a discussion of where else to draw the line as there's also `browserslist` as a valid config file as well as the `BROWSERSLIST` environment variable, plus the `BROWSERSLIST_ENV` environment variable which can point to an arbitrary config file somewhere else.

packages/hint/tests/lib/fixtures/browserslist-multiple/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "app",
+    "version": "1.0.0",
+    "browserslist": ["chrome 45"]
+  }
+
```
alrra:
```suggestion
```

packages/hint/tests/lib/fixtures/browserslist-package-json-hintconfig/package.json
```diff
@@ -0,0 +1,12 @@
...
+    "version": "1.0.0",
+    "browserslist": ["chrome 45"],
+    "hintConfig": {
+      "connector": {
+          "name": "connectorName"
+      },
+      "browserslist": ["chrome 45"]
+    }
+  }
+
```
alrra:
```suggestion
```

packages/hint/tests/lib/fixtures/browserslist-multiple/package.json
isabellachen: @antross requested for a newline at the end of the test fixtures... https://github.com/webhintio/hint/pull/1599#discussion_r244851401 
I looked a some other fixtures... can't seem to discern what the standard is, some have newlines, some don't. Should I keep or remove?

alrra: > requested for a newline at the end of the test fixtures

Sorry the misunderstanding, my comment was about removing the extra empty line at the end.

>  can't seem to discern what the standard is, some have newlines, some don't. Should I keep or remove?

[newlines should be kept](https://github.com/webhintio/hint/blob/5ddaa75e4a2634b28c4b364ed127c38906896011/.editorconfig#L15).

sarvaje:
>@sarvaje @molant Do you have any context on the history of the `loadBrowsersList` function? It looks to be dead (per my comment copied below) meaning we're going to need to add these checks in a different place.

Nope.

antross:
After digging into this further, I also confirmed that `.browserslistrc` already works with the current version of webhint (perhaps the original bug was a misunderstanding of our documentation). Thus, the only additional functionality we might add from the discussion in webhintio/rfcs#59 would be throwing an error if a user specifies a browserslist in more than one configuration file. I don't think that's high enough priority to add on it's own. 

Rather than refactor this code to avoid the dead APIs just to introduce an error message, I suggest we close this PR and move on to a different task. 

@webhintio/core Any objections? If not I'll close this PR and webhintio/rfcs#59 tomorrow.

molant:
Maybe we should update our documentation to be clearer on that aspect and indicate the priority order and that someone could define a different configuration in `.hintrc` than `.broserslistrc`?

antross:
> Maybe we should update our documentation to be clearer on that aspect and indicate the priority order and that someone could define a different configuration in `.hintrc` than `.broserslistrc`?

Agreed. Opened as PR #1611.

antross:
Closing as there isn't a functional issue to fix, we're clarifying documentation in #1611, and we're okay not adding additional error messages at this time.

antross:
@isabellachen Sorry I didn't catch the dead code issues sooner, but thank you for working this regardless!

Issue #1605, opened 2019-01-07
[Feature] Add option to specify filename for HTML formatter output


asos-francesca:
# 🚀 Feature request

## Description

Currently the html-formatter outputs into a path of `hint-report/<url_analyzed>`. Suggestion to allow this to be configured.

Currently the output for our URL test includes characters that our server cannot handle saving, so the html-formatter is not useable for us in this case.

PR #1607, opened 2019-01-07
Fix: Savename of HTMLformatter output replaces q-marks with -query-

asos-francesca:
## Pull request checklist

Previously the HTML Formatter package output the HTML file with the filename `<url-analyzed>`. If you have query strings in your URL this will create a filename with a `?` symbol which is incompatible with some systems.

This update changes the saved filename to replace `?` with `-query-` so the URL is still readable but can be saved as the filename is no longer malformed.

### DIFF
--- a/.idea/typescript-compiler.xml
+++ b/.idea/typescript-compiler.xml
@@ -1,7 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="TypeScriptCompiler">
-    <option name="isCompilerEnabled" value="true" />
-    <option name="useConfig" value="true" />
-  </component>
-</project>
--- a/.idea/typescript-compiler.xml
+++ b/.idea/typescript-compiler.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="TypeScriptCompiler">
+    <option name="isCompilerEnabled" value="true" />
+    <option name="useConfig" value="true" />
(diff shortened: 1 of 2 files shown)


molant:
Thanks a lot for this PR @asos-francesca !

Question for you and @webhintio/core: maybe we should rethink the whole file name? The folder name can become pretty big really soon, especially when adding query parameters so maybe we should switch to something like a time stamp (`yyyy-mm-dd-hh-mm`)? This way results will not be overriden when the site is scanned again.
What do you think?

sarvaje:
we can use the timestamp or just the domain instead of the full url, or a mix of timestamp + domain

molant:
formatter configurations is something that will require us a bit more of time to implement as we don't have any infrastructure to do that but definitely worth creating a feature request (I'll create it later today, we are going to do some changes in the RFCs repo). In the meantime, what will unblock you?

1. We could remove directly all the query parameters from the filename
1. Use a timestamp (+ domain)

asos-francesca:
At the moment, just the PR above to change `?` to something legal will work. I've tested this with our URL with query params and it's successful.

sarvaje:
Maybe we can merge this PR as a temporal solution, and open an issue to change the output name + allow the user to specify the final name.

asos-francesca:
I've already raised an issue for the more involved solution of custom configurability here: https://github.com/webhintio/hint/issues/1605

But this PR would help greatly as it'll allow us to use the HTML-formatter (currently having to revert to stylish)

PR #1610, opened 2019-01-07
Chore: Deprecate unused `Configuration` methods

antross:
## Pull request checklist

Make sure you:

Both `fromFilePath` and `loadBrowsersList` are no longer used by
the online scanner, CLI, or extensions. Deprecating prior to removal
since these APIs are public.

### DIFF
--- a/packages/hint/src/lib/config.ts
+++ b/packages/hint/src/lib/config.ts
@@ -28,6 +28,7 @@ import loadJSONFile from './utils/fs/load-json-file';
 import { validateConfig } from './config/config-validator';
 import normalizeHints from './config/normalize-hints';
 import { validate as validateHint, getSeverity } from './config/config-hints';
+import * as logger from './utils/logging';
 import * as resourceLoader from './utils/resource-loader';
 
 const debug: debug.IDebugger = d(__filename);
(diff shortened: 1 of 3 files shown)


packages/hint/src/lib/config.ts
```diff
@@ -213,10 +213,13 @@ export class Configuration {
     }
 
     /**
+     * @deprecated
      * Generates the list of browsers to target using the `browserslist` property
      * of the `hint` configuration or `package.json` or uses the default one
      */
     public static loadBrowsersList(config: UserConfig) {
+        console.warn('`Configuration.loadBrowsersList` is deprecated. Use `Configuration.fromConfig` instead.');
```
molant: Do we want to use our [logger](https://github.com/webhintio/hint/blob/master/packages/hint/src/lib/utils/logging.ts) for this or is it a bit overkill?

antross: Looks like `logger` currently only has `log` and `error`. Do you want to me to add a `warn` or just use `error` here?

molant:
I think this should be a warning


packages/hint/src/lib/config.ts
antross: Ok, I updated the logger to include a `warn` and used that.

antross:
If I have to. ;) I’m busy at the moment, but will update the PR tomorrow.

PR #1664, opened 2019-01-11
Fix: Manifest parser runs twice

sarvaje:
It seems like with some chrome versions, the manifest parser runs twice, one when we are traversing the dom and the connector find the `link` element (`element::link`) and a second time when the connector loads the manifest. This second case seems to be the one that didn't happen before but now it is happening.

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/parser-manifest/src/parser.ts
+++ b/packages/parser-manifest/src/parser.ts
@@ -45,6 +45,11 @@ export default class ManifestParser extends Parser<ManifestEvents> {
 
         engine.on('element::link', this.fetchManifest.bind(this));
         engine.on('fetch::end::manifest', this.validateManifest.bind(this));
+        engine.on('scan::end', this.onScanEnd.bind(this));
+    }
+
+    private onScanEnd() {
+        this.parsed = false;
     }
 
     private async fetchManifest(elementFound: ElementFound) {
@@ -110,7 +115,7 @@ export default class ManifestParser extends Parser<ManifestEvents> {
                 element,
                 error: error || new Error(`'${hrefValue}' could not be fetched (status code: ${statusCode}).`),
                 hops: (manifestNetworkData && manifestNetworkData.response.hops) || [manifestURL],
-                resource
+                resource: manifestURL
             });
 
             return;
@@ -121,7 +126,7 @@ export default class ManifestParser extends Parser<ManifestEvents> {
         await this.engine.emitAsync(this.fetchEndEventName, {
             element,
             request: manifestNetworkData.request,
-            resource,
+            resource: manifestURL,
             response: manifestNetworkData.response
         });
     }
--- a/packages/parser-manifest/src/parser.ts
+++ b/packages/parser-manifest/src/parser.ts
@@ -26,6 +26,7 @@ const schema = require('./schema.json');
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
 export default class ManifestParser extends Parser<ManifestEvents> {
+    private parsed: boolean = false;
 
     // Event names.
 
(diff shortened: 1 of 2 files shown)


molant:
I think this is a bit more complicated. IIRC one of the events has the `resource` sends the filename of the webmanifest and the other doesn't.
I can't recall now which one is the valid one. You should check what we are sending with jsdom and make sure we send the properties. Maybe we should add the `resource` property to some of the tests?

packages/parser-manifest/src/parser.ts
```diff
@@ -125,7 +126,16 @@ export default class ManifestParser extends Parser<ManifestEvents> {
...
+    private async validateManifest(fetchEnd: FetchEnd) {
+        /*
+         * Chrome fetch the manifest automatically so we need to
+         * ensure that the validation is done just once.
+         */
+        if (this.parsed) {
+            return;
+        }
+
+        this.parsed = true;
```
antross: Make sure to reset `this.parsed` on `scan::end` or errors won't get reported from `connector-local` on re-runs (e.g. when run from `extension-vscode` or the CLI in `watch` mode).

This isn't the first time I've seen cached data cause a problem so I filed #1665 to track enhancing `utils-test-helpers` to automatically test re-runs for all hints.

sarvaje:
> I think this is a bit more complicated. IIRC one of the events has the resource sends the filename of the webmanifest and the other doesn't.
I can't recall now which one is the valid one. You should check what we are sending with jsdom and make sure we send the properties. Maybe we should add the resource property to some of the tests?

`jsdom` uses the one without the filename. IMO, both are valid, but they come from different "sources". In the tests I have done, the first one running (and with my changes the only one we are running) is the one without filename, as `jsdom`.

molant:
I would actually expect the one with the filename to be the valid one 🤔
When we report an error on an image aren't we using the image's URL as the `resource` and the `element` of the HTML?

sarvaje:
> When we report an error on an image aren't we using the image's URL as the resource and the element of the HTML?

Do you have an example?

sarvaje:
I have checked the tests, and all the reports (`jsdom` and `chrome`) receive as resource the url without the filename. (you can receive some filename in chrome if you are debugging hehehe, but if not if you use a `console.log`)

sarvaje:
Anyways, maybe we should change that, and the method `fetchManifest` should dispatch the event with the resource modify to the url with the filename.

molant:
> Do you have an example?

Go to https://webhint.io/scanner/af10713b-9e11-4f6f-a5c3-0c3393c5de83#Performance

the image optimization one uses the url of the file.

image

sarvaje:
I think we are mixing things. That is not the `manifest-is-valid` rule or any rule that use the `parser-manifest` right?

packages/parser-manifest/src/parser.ts
sarvaje: done!

molant:
After talking with @sarvaje the changes he's going to do are:

* Make sure we are sending the right `resource` for the manifest events (URL of the manifest) in all connectors
* In the CDP connectors send only one event for the `fetch::end::manifest`, this way the parser can still receive multiple events and solve the scenario when in local we have multiple manifest files

sarvaje:
@molant @antross ready for another review

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -335,6 +335,17 @@ export class Connector implements IConnector {
...
         const eventName = `fetch::end::${suffix}` as 'fetch::end::*';
 
+        /*
+         * New versions of chrome are fetching the manifest
+         * automatically, so to prevent emit the event `fetch::end::manifest`
+         * twice, we need to ignore the event here.
+         * The manifest is already captured and processed by
+         * `parser-manifest` when the element is detected in the DOM.
+         */
+        if (suffix === 'manifest') {
```
molant: I'm not fully convinced about this. This code path could be executed before the traversing in the future (because chrome changes how this is implemented or because we change our traversing implementation). Maybe we should check in both places if the manifest has already been fetched before doing anything? During the travers this should be done before we do the request to download it and thus saving some time and resources.

What do you think?

PR #1690, opened 2019-01-16
Fix hint sri schema

chrisguttandin:
## Pull request checklist

Make sure you:

I noticed a little regression in the schema definition of hint-sri which this pull request tries to fix.

The current schema definition essentially compiles to something like this:

```json
"properties": {
    "baseline": {
        "oneOf": [ [ "1", "2", "3", "sha256", "sha384", "sha512" ] ],
        "type": "string"
    },
    "originCriteria": {
        "oneOf": [ [ "1", "2", "all", "crossOrigin" ] ],
        "type": "string"
    }
}
```

This pull request removes the extra nesting of the enums and also removes the number indexes which are added by the TypeScript compiler. Afterwards the schema should look like this:

```json
"properties": {
    "baseline": {
        "oneOf": [ "sha256", "sha384", "sha512" ],
        "type": "string"
    },
    "originCriteria": {
        "oneOf": [ "all", "crossOrigin" ],
        "type": "string"
    }
}
```

Please let me know if any of this was intentional and/or I should change something.

### DIFF
--- a/packages/hint-sri/src/meta.ts
+++ b/packages/hint-sri/src/meta.ts
@@ -15,11 +15,11 @@ const meta: HintMetadata = {
         additionalProperties: false,
         properties: {
             baseline: {
-                oneOf: [Object.keys(Algorithms)],
+                oneOf: Object.keys(Algorithms),
                 type: 'string'
             },
             originCriteria: {
-                oneOf: [Object.keys(OriginCriteria)],
+                oneOf: Object.keys(OriginCriteria),
                 type: 'string'
             }
         }
--- a/packages/hint-sri/src/meta.ts
+++ b/packages/hint-sri/src/meta.ts
@@ -15,11 +15,15 @@ const meta: HintMetadata = {
         additionalProperties: false,
         properties: {
             baseline: {
-                oneOf: Object.keys(Algorithms),
+                oneOf: Object.keys(Algorithms).filter((key) => {
+                    return isNaN(parseInt(key, 10));
+                }),
                 type: 'string'
             },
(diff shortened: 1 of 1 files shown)


molant:
Definitely not intentional. Documentation doesn't expect that neither: https://webhint.io/docs/user-guide/hints/hint-sri/#can-the-hint-be-configured

Wonder why it doesn't complain during the test phase because I don't think we are passing an array.

Will investigate during the morning.

Thanks @chrisguttandin!

molant:
@chrisguttandin I'm curious, how did you find out this was a regression?
I'm looking at the [history and it's always been `oneOf: [Object.keys]` (minus the `originCriteria`)](https://github.com/webhintio/hint/blame/f3d1fab729453b8d3f53bd2404acf9ba80e35f3c/packages/hint-sri/src/hint.ts#L46).

I keep looking at the tests and trying to find out what's going on.

chrisguttandin:
@molant After the breaking changes of v3 I still wanted to check the subresource integrity of all resources because I'm paranoid. :-) Therefore I tried to change my .hintrc config file and noticed the problem.

To be honest, I should not have called this a regression. I just noticed it after the update because I changed my config file. Sorry for that.

molant:
> I still wanted to check the subresource integrity of all resources because I'm paranoid. :-)

❤

molant:
Thanks @chrisguttandin!

We are fixing an issue with the release script but we should publish this later today.

alrra:
> we should publish this later today.

Done in `@hint/hint-sri` `v3.0.1`.

PR #1695, opened 2019-01-16
Fix: Use enum instead of oneOf

chrisguttandin:
## Pull request checklist

Make sure you:

I'm sorry for the confusion, but unfortunately my other pull request (#1690) did not fix the issue completely.

This pull request replaces the `oneOf` keyword with the `enum` keyword in the definition of the JSON schema. And this time I actually tested the changes by running the compiled changes with the `hint` command.

I think `oneOf` is meant for allowing different types, whereas `enum` is a special case where the type is always the same but can have different values.

https://json-schema.org/understanding-json-schema/reference/combining.html#oneof
https://json-schema.org/understanding-json-schema/reference/object.html#properties

### DIFF
--- a/packages/hint-sri/src/meta.ts
+++ b/packages/hint-sri/src/meta.ts
@@ -15,13 +15,13 @@ const meta: HintMetadata = {
         additionalProperties: false,
         properties: {
             baseline: {
-                oneOf: Object.keys(Algorithms).filter((key) => {
+                enum: Object.keys(Algorithms).filter((key) => {
                     return isNaN(parseInt(key, 10));
                 }),
                 type: 'string'
             },
             originCriteria: {
-                oneOf: Object.keys(OriginCriteria).filter((key) => {
+                enum: Object.keys(OriginCriteria).filter((key) => {
                     return isNaN(parseInt(key, 10));
                 }),
                 type: 'string'


packages/hint-sri/src/meta.ts
```diff
@@ -15,13 +15,13 @@ const meta: HintMetadata = {
         additionalProperties: false,
         properties: {
             baseline: {
-                oneOf: Object.keys(Algorithms).filter((key) => {
+                enum: Object.keys(Algorithms).filter((key) => {
```
molant: Another rule that [does something similar is `performance-budget`](https://github.com/webhintio/hint/blob/64d0ed2357f34aac346ed20c689033ca266e510a/packages/hint-performance-budget/src/meta.ts#L18).
The code in this case is:

```ts
connectionType: {
    oneOf: [{ enum: Connections.ids }],
        type: 'string'
    },
```

RIght now I don't know which one is the "most correct" one 🤔
If you have time, do you mind doing a bit of digging? If not I'll research in the morning. Tomorrow should be a calmer day.

Thanks!

chrisguttandin: As far as I understand the JSON schema rules the `performance-budget` schema is redundant. The following definition should be equivalent:

```javascript
{
    // ...
    connectionType: {
        enum: Connections.ids,
        type: 'string'
    },
    // ...
}
```

I think `oneOf` says: "one of the things that follow in the array". In the case of the `connectionType` property the array has only one member and therefore I think it means "one of one".

I tested this assumption and can confirm that the `performance-budget` hint can still be configured when the changes described above are made to its metadata definition.

molant:
[v3.0.2 is published now](https://www.npmjs.com/package/@hint/hint-sri). Thanks again!

chrisguttandin:
Should I open a new pull request to condense the schema of the `performance-budget` hint?

molant:
Sure!

Issue #1580, opened 2018-12-13
[parser-css] Use postcss-safe-parser for fault tolerance


antross:
Currently malformed CSS will cause the entire file to be ignored. By switching to use the fault-tolerant parser for PostCSS such files can still be analyzed, behaving more like a browser would.

https://github.com/postcss/postcss-safe-parser

PR #1696, opened 2019-01-17
Use postcss-safe-parser

borgitas21:
## Pull request checklist

Make sure you:

Use postcss-safe-parser for fault tolerance #1580


packages/parser-css/tests/integration-tests.ts
```diff
@@ -0,0 +1,74 @@
...
+    const code = '.foo { color: #fff } a {';
+
+/*     await t.context.engine.emitAsync('fetch::end::css', {
+        resource: 'styles.css',
+        response: {
+            body: { content: code },
+            mediaType: 'text/css'
+        }
+    }); */
+
```
borgitas21: @antross I'm not sure how to do a real integration test without mocking the `fetch::end::css`. Could you guide me a bit?

antross: Mocking `fetch::end::css` is fine. That's consistent with what we do for similar tests in `parser-html` and `parser-javascript`. My main interest is having some coverage that actually runs `postcss` with both well-formed and malformed input which this would still do.

borgitas21: 👍

packages/hint-compat-api/package.json
```diff
@@ -24,6 +24,7 @@
     "npm-run-all": "^4.1.5",
     "nyc": "^13.1.0",
     "postcss": "^7.0.13",
+    "postcss-safe-parser": "^4.0.1",
```
molant: Why do we need `postcss` and `postcss-safe-parser` in the `compat-api` hint? Isn't all the work done in the parser?

molant:
@borgitas21 can you please rebase with the latest in master and fix the conflicts?

packages/hint-compat-api/package.json
borgitas21: Yes, you are right, but since we had to add postcss to this hint, I thought we need to manage it in the same way. Maybe postcss is only useful for typing purposes in the compat-api hint. Is that correct?

borgitas21:
> @borgitas21 can you please rebase with the latest in master and fix the conflicts?
@molant Done!

antross:
@borgitas21 We've been working on migrating our tests to avoid `.serial` so I wanted to make some more significant changes. I also previously missed that there was a "basic-tests" file already along the lines of the "integration" tests I was looking for (just needing a malformed case). Additionally I wanted to ensure we didn't pollute our "public" types in `types.ts` with test-only types.

TL;DR I took the core of this PR and opened #1757 with the set of tests trimmed back a bit and refactored to run in parallel. Thanks for getting this started!

PR #1708, opened 2019-01-17
Migration to Dependabot

molant:
## Pull request checklist

Make sure you:

* Remove all references to greenkeeper
* Make Dependabot commit messages valid
* Add Dependabot's `config.yml`


.dependabot/config.yml
```diff
@@ -0,0 +1,29 @@
...
+      - match:
+          dependency_type: "development"
+          update_type: "all"
+      - match:
+          dependency_type: "production"
+          update_type: "all"
+    ignored_updates:
+      - match:
+          dependency_name: "angular"
+          version_requirement: "^1.4.9"
```
molant: @greysteil is this the right format for our case? The documentation only says:

> The range format is specific to the package manager

greysteil: Yes - that will work perfectly. With the `^` operator in JS Dependabot will interpret this as "ignore angular versions < 2.0.0".

greysteil: I'll also update the docs so it's clearer what we mean with that comment. Basically it's that Dependabot will interpret requirements in the same way as the package manager would, so with Ruby you can write `~> 1.0` and with JS you can write `1.x.x` and Dependabot will correctly interpret what you mean.

molant: @greysteil and if I want to ignore all updates for a package should I just remove the `version_requirement` field? For angular and jquery we use it to validate our vulnerability detection so we don't want to ever have to upgrade the packages.

Thanks!

greysteil: At the moment you have to specify something in the `version_requirement` field, but using a `*` would mean ignore everything.

(Would be nice to not be able to specify it and have that inferred, and since that's non-breaking I'll have a look at making that change now, too.)

molant: > Update: Dependabot now infers that you want all versions ignored if you don't specify a `version_requirement`.

Awesome 🎉

PR #1794, opened 2019-01-30
Build: New test everything script

### LINKED ISSUE #1714: [Feature] Run all tests on scheduled build in Azure Pipelines
# 🚀 Feature request

## Description

Right now we run an scheduled build at 11.00pm PST but it takes into account the latest commits to decide what to run so potentially not all test will be run.

**What do you want to be added?**

I'd like the scheduled build to run all the test suite. We will probably have to add a new task in `package.json` like `test-all` and refactor a bit the azure pipelines templates.

**Are you willing to work on this?**

Yup

This will guarantee that we don't have any problems that we potentially skipped because of the number of commits and commit messages.

molant:
## Pull request checklist

Make sure you:

New script to be executed on the Scheduled builds of Azure Pipelines
that test all the packages. It uses `tsconfig.references` to know
which one need to be tested and how to resolve the dependencies that
need to be built and tested first.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Fix #1714


packages/connector-jsdom/package.json
```diff
@@ -61,7 +61,7 @@
   },
   "repository": "webhintio/hint",
   "scripts": {
-    "build": "npm run clean && npm-run-all build:*",
+    "build": "npm-run-all build:*",
```
sarvaje: IMO, we still needs a build with the `npm run clean` in case we are just testing a package.

sarvaje: Maybe we can have `build` as it was, and then a `build-all` without the clean.

antross: I'm fine with not automatically running `clean` on every build and leaving it for explicit use. I think running `clean` for a full build from the root is still a good idea, which this PR still does, so I'm good with the changes.

molant: OK, can you 👍👎 this PR whenever you have the time and I'll fix the conflicts with `yarn.lock` and get it merged right after.

PR #1826, opened 2019-02-04
Fix: Update 'hint-sri' to work with 'connector-local'

### LINKED ISSUE #1632: Update `hint-sri` to work with `connector-local` and validate external resources [2]
Currently this hint relies on external resources being fetched, but `connector-local` intentionally doesn't fetch external resources as it works with static files.

I propose we update this hint to fetch *external* resources (e.g. from a CDN) itself when running locally. Local files would be difficult to validate as they're potentially changing frequently... perhaps we could do something special for items coming from `package.json` dependencies. This should likely also include some caching to avoid an abundance requests to external resources from repeated runs during development.

sarvaje:
Fix #1632

## Pull request checklist

Make sure you:


molant:
@sarvaje looks like there are some issues with CI. Can you check if they are because of your changes?

antross:
@sarvaje Let's add `hint-sri` to `configuration-development` as part of this change too.

packages/hint-sri/src/hint.ts
```diff
@@ -296,17 +362,126 @@ Actual:   ${hashes.join(', ')}`;
...
+         * for local files.
+         */
+        if (resource.startsWith('file://')) {
+            debug(`Ignoring local resource: ${resource}`);
+
+            return;
+        }
+
+        /*
+         * Requester is not included in webpack bundle for `extension-browser`.
```
antross:
```suggestion
         * `requestAsync` is not included in webpack bundle for `extension-browser`.
```

Maybe also worth commenting why we're NOT using `Requester` here.

packages/hint-sri/tests/fixtures/local-no-pass/test.html
```diff
@@ -0,0 +1 @@
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/core.js" integrity="sha256-YCbKJH6PpUlk130udu/JepdKVpXjdEyzje+z1pE=" crossorigin="anonymous"></script>
```
antross: Nit: include newlines at the end of this and the other test files.

sarvaje:
@antross done!

molant:
@sarvaje can we have #1828 in this PR as well?

packages/hint-sri/src/hint.ts
```diff
@@ -296,17 +362,130 @@ Actual:   ${hashes.join(', ')}`;
...
+        /*
+         * 'this.isScriptOrLink' has already checked
+         * that the src attribute exists, so it is safe to use !.
+         */
+        evt.resource = new URL(evt.element.getAttribute('src')!, evt.resource).href;
+        await this.validateElement(evt, origin);
+    }
+
+    /** Validation entry point for events element::link */
+    private async validateStyleElement(evt: ElementFound) {
```
molant: Can't this be combined with `validateScriptElement`?
Both seem to do the same thing and the only thing that changes is  the attribute to get (`src` or `href`).

packages/hint-sri/src/hint.ts
```diff
@@ -296,17 +362,130 @@ Actual:   ${hashes.join(', ')}`;
         debug(`Validating integrity of: ${evt.resource}`);
 
         await everySeries(validations, async (validation: Function) => {
-            return await validation(evt);
+            return await validation(evt, origin, cacheErrors);
         });
     }
 
+    private async validateElement(evt: ElementFound, origin: string) {
+        if (this.isInCache(evt)) {
```
molant: I think we should try to follow the code flow we had earlier with `validateResource` where we have all the checks in `validations` and split `validateElement` into multiple separate checks (and have a single `validate` method).

E.g.:
```ts
private async isScanEnded(evt, origin) {
            /*
             * The item is cached. For the VSCode extension and the
             * local connector with option 'watch' activated we
             * should report what we have in the cache after the
             * first 'scan::end'.
             */
            if (this.scanEnded) {
                const promises = this.getCache(evt).map((error) => {
                    this.context.report(error.resource, error.message, error.options);
                });

                 await Promise.all(promises);

                 return true;  
            }

             return false;
}

private async isLocalResource(evt, origin) {
        const { resource, element } = evt;
        if (resource.startsWith('file://')) {
            debug(`Ignoring local resource: ${resource}`);

             return Promise.resolve(true);
        }
        return Promise.resolve(false);
}

// And add this checks to `validations` and use the same entry point for all the event handlers
... 
```

sarvaje:
> @sarvaje can we have #1828 in this PR as well?

I created a new PR because that is in another hint, but I can move it here if you prefer that.

molant:
Nervermind, I thought it was the same hint.

packages/hint-sri/src/hint.ts
sarvaje: The think is that if we move all of this to the validation flow, how and when do you do the request to get the content of the element?

It would be weird to have a step in the validation that just download something.

packages/hint-sri/src/hint.ts
```diff
@@ -340,139 +340,158 @@ Actual:   ${integrities.join(', ')}`;
...
+            const promises = this.getCache(evt).map((error) => {
+                return this.context.report(error.resource, error.message, error.options);
+            });
 
-                await Promise.all(promises);
-            }
+            await Promise.all(promises);
 
-            return;
+            // Return false to stop the validations.
```
molant: No need for this comment

packages/hint-sri/src/hint.ts
```diff
@@ -340,139 +340,158 @@ Actual:   ${integrities.join(', ')}`;
...
          *
-         * Note: We are not using `Requester` becuase it depends on `iltorb` and it can
+         * Note: We are not using `Requester` because it depends on `iltorb` and it can
          * cause problems with the vscode-extension because `iltorb` dependens on the
          * node version for which it was compiled.
+         *
+         * We can probably use Requester once https://github.com/webhintio/hint/issues/1604 is done,
+         * and vscode use the node version that support it.
+         *
+         * When the crossorigin="use-credentials" but the response contains
```
molant: > When using crossorigin="use-credentials" and the response contains
> the header `Access-Control-Allow-Origin` with value `*` Chrome blocks the access

packages/hint-sri/src/hint.ts
```diff
@@ -340,139 +340,158 @@ Actual:   ${integrities.join(', ')}`;
...
+         * the header `Access-Control-Allow-Origin` with value `*` Chrome block the access
+         * to the resource by CORS policy, so we will reach this point
+         * through the traverse of the dom and response.body.content will be ''. In this case,
+         * we have to prevent download any resource.
```
molant: > we have to prevent the download of the resource.

packages/hint-sri/src/hint.ts
```diff
@@ -340,139 +340,158 @@ Actual:   ${integrities.join(', ')}`;
...
+            return false;
+        }
+
         if (!requestAsync) {
-            return;
+            return true;
+        }
+
+        /*
+         * If the content already exists, we don't need to download it.
```
molant: Nit: You can collapse and use a single line comment `//` or `/* */` in the same line.

packages/hint-sri/src/hint.ts
```diff
@@ -268,45 +284,227 @@ export default class SRIHint implements IHint {
...
+
+        return Promise.resolve(true);
+    }
+
+    private isNotLocalResource(evt: FetchEnd) {
+        const { resource } = evt;
+
+        /*
+         * If the resource is a local file, ignore the analysis.
+         * The sri usually is added on the building process before publish,
```
molant: Nit: Move this comment to the beginning of the method:

```js
/**
  * If the resource is...
  */
private isNotLocalResource()
```

sarvaje:
@molant done!

packages/hint-sri/src/hint.ts
```diff
@@ -268,45 +284,224 @@ export default class SRIHint implements IHint {
...
+        }
+
+        return Promise.resolve(true);
+    }
+
+    private async isScanEnded(evt: FetchEnd) {
+        const isInCache = this.isInCache(evt);
+
+        if (this.scanEnded && isInCache) {
+            /*
```
molant: You forgot to put this comment outside

packages/hint-sri/src/hint.ts
```diff
@@ -268,45 +284,224 @@ export default class SRIHint implements IHint {
...
+            return false;
+        }
+
+        return !isInCache;
+    }
+
+    private async downloadContent(evt: FetchEnd) {
+        const { resource, response, element } = evt;
+
+        /*
```
molant: And this one too.

sarvaje:
@molant done again hahaha

sarvaje:
@antross done!

PR #1904, opened 2019-02-13
New: Add browser extension telemetry

### LINKED ISSUE #1572: [extension-browser] Add telemetry
We should collect telemetry to help us improve the browser extension.

This should likely cover:
* How many scans were done.
* What configurations were used.
* What errors were reported.
* What browser was used.

As part of researching this we should also look at what other extensions do here, how we need to ask for the user's permission, and how we can respect settings users may have already set with their browser.

antross:
## Pull request checklist

Make sure you:

Fix #1572


antross:
Pre-built bundle for testing (though to actually verify any of the generated logs you'll need to sign into our analytics account):
* [extension-browser-pr-1904.zip](https://github.com/webhintio/hint/files/2862515/extension-browser-pr-1904.zip)

molant:
@antross looks like there are conflicts and this time it isn't my fault...

packages/extension-browser/src/devtools/panel/panel.ts
```diff
@@ -55,20 +71,22 @@ const onMessage = (message: Events) => {
...
         // Display the "Results" page after any pending delays.
-        if (Date.now() < delayResultsUntil) {
-            resultsTimeout = setTimeout(showResults, delayResultsUntil - Date.now());
+        if (now() < delayResultsUntil) {
+            resultsTimeout = setTimeout(showResults, delayResultsUntil - now());
         } else {
             showResults();
         }
+
+        trackFinish(getScanDuration());
```
molant: Random thought: Wouldn't it be great to use decorators for this things? 😔

packages/extension-browser/src/manifest.json
```diff
@@ -7,6 +7,7 @@
     "icons": {
         "180": "icon.png"
     },
+    "content_security_policy": "script-src 'self' https://az416426.vo.msecnd.net/; object-src 'self'",
```
molant: Does `applicationinsights-js` download an external resource? Shouldn't we try to have it locally and bundle it? What happens if the domain is not reachable?

antross: Yes, the bundled bit is just a bootstrapper that downloads the full analytics script. The bootstrapper includes stubs for supported APIs and queues calls until the full script has loaded (so worst case scenario is we don't log analytics for a given session - nothing will break). This is the recommended approach per https://github.com/Microsoft/ApplicationInsights-JS#initializing-application-insights-js-sdk-script.

If we want to bundle the *entire* script, it unfortunately doesn't support being loaded as a module (only the bootstrapper does) and [the issue to fix that](https://github.com/Microsoft/ApplicationInsights-JS/issues/213) has been open for ~2.5 years 😞

antross:
> @antross can you fix the `package.json` and create a new bundle?

Will do. I'm ready for dependabot to only run on the weekends...

molant:
> I'm ready for dependabot to only run on the weekends

Yeah... Or outside working hours. I've been retriggering some dependabot builds that were failing because of missing binaries too.

Issue #1532, opened 2018-11-26
Offer a Node API


donaldpipowitch:
> 1. Propose a new feature/change:
> 
>    Write a short description of your proposal with (if applicable)
>    some examples of the expected behaviour.

As far as I can tell the `hint` package only exports an `Engine` class which doesn't offer enough functionality to use `hint` with a regular Node API, because a lot of the common functionality (e.g. loading configs) is hidden in the CLI:

https://github.com/webhintio/hint/blob/ad3fa3f556e4f55af14e4d2b371fcf0c681ef8ed/packages/hint/src/bin/hint.ts#L98

It'd be nice, if there was an easy to use Node API with a similar "interface"/config as the CLI.

---

One of my use cases: I'm still not entirely sure how the integrate `hint` to check my whole website. I don't just want to call the CLI and give one or two static URLs. I want to create some custom crawl logic for example and want to call `hint` when I need it.

molant:
You can `import` any file we are using internally for your own use. That's what we do in [`extension-vscode`](https://github.com/webhintio/hint/blob/970f3860bfe4e87a22838e0e0afe113e86ca0d66/packages/extension-vscode/src/server.ts#L19) to load some specific resources and change the detected configuration (we force a parser and change the connector IIRC).

The crawler request has been asked a couple times (last week by @mgifford [in this tweet](https://twitter.com/mgifford/status/1065257183837896706)). We could look into creating something that will be supported officially depending on the requirements (e.g.: pass a .txt with links, or a sitemap, or something similar).

In any case, @antross and I have talked in the past about this and you are right it's not easy to use it unless you know the code base well. What do you think of the following tasks for us:

* [ ] Create new `exports` that simplify the paths of the most common used files for better consumption. Maybe something like `import hintAPI from 'hint/api'` that exposes `Engine`, load configurations, etc. (or something similar)
* [ ] Create a guide on how to use `hint`'s API with an example (maybe sitemap support for a crawler?)

donaldpipowitch:
Thank you for the feedback!

> Create new `exports` that simplify the paths of the most common used files for better consumption. Maybe something like `import hintAPI from 'hint/api'` that exposes `Engine`, load configurations, etc. (or something similar)

[...]

PR #1910, opened 2019-02-14
Cdp and jsdom using a copy of the DOM

sarvaje:
## Pull request checklist

Make sure you:


molant:
@sarvaje looks like there are conflicts, can you please fix those?

packages/connector-chrome/package.json
```diff
@@ -14,17 +14,19 @@
   },
   "description": "hint connector for Google Chrome",
   "devDependencies": {
+    "@hint/parser-html": "^2.0.2",
```
molant: I'm not sure we should depend on the parser even if it's a `devDependency`. If the parser is the one that is going to do all the traversing then the related tests should be moved there and deleted from any connector.

packages/connector-jsdom/src/connector.ts
```diff
@@ -293,9 +269,7 @@ export default class JSDOMConnector implements IConnector {
 
                     debug(`${this.finalHref} loaded, traversing`);
                     try {
-                        await this.server.emitAsync('traverse::start', event);
-                        await this.traverseAndNotify(window.document.children[0] as HTMLElement);
-                        await this.server.emitAsync('traverse::end', event);
+                        await this.parseHTML();
```
molant: Why can't we just do a `await this.server.emitAsync('can-traverse::html')`? I don't think we are doing anything similar with other parsers, right?

packages/connector-jsdom/src/connector.ts
```diff
@@ -167,6 +128,21 @@ export default class JSDOMConnector implements IConnector {
...
 
+    private parseHTML(): Promise<void> {
+        return new Promise(async (resolve) => {
+            const html = await this.html;
+
+            await this.server.on('end-traverse::html', () => {
+                resolve();
+            });
+
+            await this.server.emitAsync('can-traverse::html', {
```
molant: We had a conversation about this but I can't remember now why can't we use `fetch::end::html`. Can you please explain it here?

packages/hint/docs/contributor-guide/getting-started/events.md
```diff
@@ -224,6 +225,31 @@ type CanEvaluateScript {
 }
 ```
 
+## `can-traverse::html`
+
+Event is emitted **when** the `connector` is ready to traverse
+the html.
```
molant: What does it mean to be ready?

packages/parser-html/README.md
```diff
@@ -34,14 +34,19 @@ And then activate it via the [`.hintrc`][hintrc] configuration file:
...
-This `parser` emits the event `parse::html` of type `HTMLParse`
+This `parser` emits the event `parse::end::html` of type `HTMLParse`
 which has the following information:
 
 * `window`: an [`IAsyncWindow`][asynchtml] object containing the
   parsed document.
 * `html`: a string containing the raw HTML source code.
 * `resource`: the parsed resource.
 
+And the event `parse::start::html` of type `Event` witch has the
```
antross:
```suggestion
And the event `parse::start::html` of type `Event` which has the
```

packages/connector-jsdom/src/connector.ts
sarvaje: fetch::end::html is trigger at the begining, when you are fetching the site, but we don't want to traverse that html, we need to wait (`waitFor`) until the scripts have run and then, traverse the html. That is why we can't use the event `fetch::end::html`

antross: Also, I think we should have further discussion about whether we want `fetch::end::html` to still trigger parsing, just not traversal.

Context: I'd like to use `parse::end::html` in the future to analyze HTML beyond the top-level page (e.g. nested frames or fragments fetched via XHR), but with the expectation that hints using this are directly interrogating the `IAsyncWindow` provided by `parse::end::html` instead of listening for traversal/element events.

packages/connector-jsdom/src/connector.ts
sarvaje: You are right, I always forget that the `emitAsync` is going to wait until the listeners ends.

packages/connector-jsdom/src/connector.ts
molant: `hint-meta-charset-utf-8` requires the initial HTML to validate the tag is in the right place. Right now it's using cheerio and the initial HTML. This could be a good opportunity to rethink how/when we are doing all the HTML parsing.

sarvaje: What about having bot events `fetch::end::html` will just parse the html and will return `parse::end::html`.
Then we can have `can-traverse::html` which will use a new jsdom object to traverse.

antross: I'm thinking `parse::end::html` should always represent the initial HTML (allowing for static analysis of the original source) and the traversal events should always represent the "initialized" DOM (dynamic analysis). Hints should use one or the other depending on their needs (if a hint uses both it will need to ensure it avoids duplicate reports).

To me that means when we're taking a snapshot of the DOM, we should NOT generate `parse::end::html`, but we should generate it in response to `fetch::end::html` events.

sarvaje: Let me clarify this, because I think we are thinking the same hehehe.
`fetch::end::html` will end emitting a `parse::end::html` and will not traverse the html.
`can-traverse::html` will traverse the html that it get by param and will NOT emit `parse::end::html`.

Is this what are you saying?

antross: Yep, I think we're on the same page as I didn't see your comment until after I posted mine. 😉

Even so, I'll paraphrase just to lock it in. 
1. `parser-html` will continue respond to the existing `fetch::end::html` event and generate a `parse::end::html` event that includes an `IAsyncWindow` generated via `jsdom`, but will NOT traverse the HTML in this context.
2. `parser-html` will additionally respond to the new `can-traverse::html` event to traverse the HTML using `jsdom`, but will NOT generate a `parse::end::html` in this context.

sarvaje: Perfect! I will work on that.

molant: Small clarification because it got me thinking for a minute:

2. `parser-html` will additionally respond to the new `can-traverse::html` event to traverse the HTML using `jsdom` sending all the `traverse` events (like `traverse::end`) but will NOT generate a `parse::end::html` in this context.

packages/hint/docs/contributor-guide/getting-started/events.md
sarvaje: I have change `is ready` for `wants to`, is that better?

packages/hint-amp-validator/tests/tests.ts
```diff
@@ -38,4 +38,4 @@ const defaultTests: HintTest[] = [
     }
 ];
 
-hintRunner.testHint(hintPath, defaultTests);
+hintRunner.testHint(hintPath, defaultTests, { parsers: ['html'] });
```
molant: I think we should modify the hintRunner to add the HTML parser by default to avoid having to do this.

sarvaje: I have been talking with Tony, and I'm going to move the traverse stuff to a utils, instead of being part of the parser.

The parser will just parse. So this will not be a dependency anymore.

molant: FTFY 😛
> The parser will just ~~parse~~ traverse.

sarvaje:
@molant @antross, this is ready to review

packages/hint-no-protocol-relative-urls/tests/tests.ts
```diff
@@ -76,4 +76,4 @@ const tests: HintTest[] = [
     }
 ];
 
-hintRunner.testHint(getHintPath(__filename), tests);
+hintRunner.testHint(getHintPath(__filename), tests, { parsers: ['html'] });
```
molant: Why is this the only hint that needs this?

packages/connector-jsdom/src/connector.ts
```diff
@@ -36,13 +36,16 @@ import { getContentTypeData, getType } from 'hint/dist/src/lib/utils/content-typ
...
     IConnector,
-    ElementFound, Event, FetchEnd, FetchError, TraverseDown, TraverseUp,
+    Event, FetchEnd, FetchError,
     NetworkData
 } from 'hint/dist/src/lib/types';
-import { JSDOMAsyncHTMLElement, JSDOMAsyncHTMLDocument } from 'hint/dist/src/lib/types/jsdom-async-html';
+import { JSDOMAsyncHTMLElement, JSDOMAsyncHTMLDocument, JSDOMAsyncWindow } from 'hint/dist/src/lib/types/jsdom-async-html';
 import { Engine } from 'hint/dist/src/lib/engine';
 import isHTMLDocument from 'hint/dist/src/lib/utils/network/is-html-document';
+import createAsyncWindow from 'hint/dist/src/lib/utils/dom/create-async-window';
```
molant: We really need to do #1532 so this isn't that ugly 😔

packages/hint-no-protocol-relative-urls/tests/tests.ts
sarvaje: Let me ckeck, but maybe I just missed that one

molant:
Superseeded by #1943

PR #1924, opened 2019-02-16
New: Checks if any SVG element is created using createElement()

### LINKED ISSUE #1625: [Hint] Can't use `createElement` to create SVG elements [5]
Not everyone knows that namespaces are still a thing at the DOM level, especially if you're new to the web platform. As such, I've often seen developers initially surprised when starting to work with SVG in HTML and [calls like `createElement('svg')` and `createElement('circle')` don't do what they expect](https://stackoverflow.com/questions/19281122/svg-circle-not-being-drawn-with-javascript).

I propose creating a hint using the AST from webhint's `parse::end::javascript` event to look for `createElement` calls where the first parameter is a known SVG element name (that doesn't overlap with an HTML element name) and inform users they need to use `createElementNS('http://www.w3.org/2000/svg', elementName)` instead.

*Note: this is another hint that needs a new category as it's more of a general web-platform pitfall than an interoperability issue - all browsers behave the same way here.*

utsavized:
--------------------------------------------------------------

Fix #1625

## Pull request checklist

Make sure you:


packages/hint-create-element-svg/package.json
```diff
@@ -0,0 +1,71 @@
...
+  "devDependencies": {
+    "@hint/parser-javascript": "^2.0.3",
+    "@hint/utils-tests-helpers": "^2.0.3",
+    "@types/debug": "0.0.31",
+    "@types/node": "10.12.18",
+    "ava": "^0.25.0",
+    "cpx": "^1.5.0",
+    "eslint": "^5.12.0",
+    "eslint-plugin-markdown": "^1.0.0",
+    "eslint-plugin-typescript": "0.14.0",
```
sarvaje:
```suggestion
    "@typescript-eslint/eslint-plugin": "^1.3.0",
```

packages/hint-create-element-svg/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "npm-link-check": "^3.0.0",
+    "npm-run-all": "^4.1.5",
+    "nyc": "^13.1.0",
+    "rimraf": "^2.6.3",
+    "hint": "^4.3.1",
+    "typescript": "^3.2.2",
+    "typescript-eslint-parser": "21.0.2"
```
sarvaje:
```suggestion
    "@typescript-eslint/parser": "1.3.0"
```

packages/hint-create-element-svg/src/svgElements.ts
```diff
@@ -0,0 +1,73 @@
+const svgElements: string[] = [
```
sarvaje: Question: What is faster, ask use 'includes' in an array, or `has` in a Set?, if set is faster, I think we should change this to a Set.

utsavized: Is it Friday yet? Duh! Yes, set is for sure the way to go here for the lookups.

packages/hint-create-element-svg/README.md
```diff
@@ -0,0 +1,80 @@
+# create-element-svg (`@hint/hint-create-element-svg`)
```
antross: This title should align better with our other hints (probably another place `create-hint` could use an update).
```suggestion
# No `createElement` with SVG (`create-element-svg`)
```

packages/hint-create-element-svg/src/meta.ts
```diff
@@ -0,0 +1,15 @@
+import { Category } from 'hint/dist/src/lib/enums/category';
+import { HintScope } from 'hint/dist/src/lib/enums/hint-scope';
+import { HintMetadata } from 'hint/dist/src/lib/types';
+
+const meta: HintMetadata = {
+    docs: {
+        category: Category.pitfalls,
+        description: `Infrom users that they need to use createElementNS to create SVG elements instead of createElement`
```
antross: We support a "friendly name" in the metadata now, usually matching the `README` header (another place `create-hint` could use an update). Also fix `Infrom` to be `Inform` in the description.
```suggestion
        description: `Inform users that they need to use createElementNS to create SVG elements instead of createElement`,
        name: 'No `createElement` with SVG'
```

packages/hint-create-element-svg/tests/tests.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const htmlElementCreate = 'document.getElementById("container").appendChild(document.createElement("a"));';
+const invalidCircleCreate = 'document.getElementById("container").appendChild(document.createElement("circle"));';
+const invalidSvgCreate = 'document.getElementById("container").appendChild(document.createElement("svg"));';
+const validSvgCreate = 'document.getElementById("container").appendChild(document.createElementNS("http://www.w3.org/2000/svg", "svg"));';
+
+const tests: HintTest[] = [
+    {
+        name: 'Should not use createElement to create SVG element',
+        reports: [{ message: 'Avoid using createElement to create SVG elements; use createElementNS instead' }],
```
antross: Let's validate we're getting the report generated at the right position in the file too.

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,71 @@
...
+
+        const validateScript = async (scriptData: ScriptParse) => {
+
+            debug(`Validating hint create-element-svg`);
+
+            const sourceCode = scriptData.sourceCode;
+            const results = linter.verify(sourceCode, { rules: { 'svg-create': 'error' } });
+
+            for (const result of results) {
+                await context.report(scriptData.resource, result.message);
```
antross: Looks like we're not passing along location information here. Not sure if there's a good place to get it off of `result` or not (maybe @molant knows)...
```suggestion
                await context.report(scriptData.resource, result.message, { location });
```

packages/hint-create-element-svg/package.json
antross: @sarvaje @molant Do we still need to fix this in the `create-hint` package?

molant: No, I just need to publish a new version and I've been a bit busy this week 😔

packages/hint-create-element-svg/src/hint.ts
molant: I think that ESLint provides that information already. What I don't know is what happens if the script is inline with the HTML:

```html
<html>
<head>
  <script>
document.createElement('svg');
  </script>
</head>
</html>
```

Hopefully ESlint reports line 4 but we should cofirm this is the case (although maybe this just happens when using [`eslint-plugin-html`](https://www.npmjs.com/package/eslint-plugin-html).

utsavized: This was one thing I wasn't sure about ... result object has the location information, and I think report accepts location object, so I was thinking of supplying that here, but wasn't sure if eslint automagically does that stuff when it reports back?

Also, what's the best way to test this hint out against a real page (not test case) to see how it is reported back to a real user?

molant: Enable this in packages/hint/.hintrc and disable the rest of the rules. You can also install the vs code extension and it should detect all the rules and packages so you can create some js that uses it (you will still need a .hintrc file in the root)

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -62,7 +62,7 @@ export default class CreateElementSvgHint implements IHint {
             const results = linter.verify(sourceCode, { rules: { 'svg-create': 'error' } });
 
             for (const result of results) {
-                await context.report(scriptData.resource, result.message);
+                await context.report(scriptData.resource, result.message, { location: { column: result.column, line: result.line }});
```
antross: Awesome! Don't forget to update the tests to validate the location too.

utsavized:
@antross @molant @sarvaje using eslint's result location did not work, as it sets line number to the beginning of the script tag and not the full source code. To solve that, I subscribed to the `element::script` event, and tried to use the script element's location to find the real problem location. However, I had some issues with `element.getLocation()` being `null` (I think because it is a promise and isn't resolved until the after the `report` call is made). So, used the `findProblemLocation()` function from `location-helper` to get position of the script element, then offset eslint's result location to be relative to the full source code. Not sure if this is the most optimal solution, but I am getting accurate line/col numbers now.

hint-code

molant:
I believe the issue with `getLocation` should be solved once #1943 gets merged (and if not we should make sure it is).

sarvaje:
> I believe the issue with `getLocation` should be solved once #1943 gets merged (and if not we should make sure it is).

Yep, that PR should solve the problem.

molant:
> that PR should solve the problem.

Then maybe we should try to get that one merged asap and update once it's done.

antross:
Looking at the screenshot suggests the position is still off by a little bit as `document` is getting underlined instead of `createElement`...

utsavized:
@antross Fixed.

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,96 @@
...
+         * a 'createElement' property. If yes, check if that call expression's
+         * first argument is an SVG element. In that case, report an error as
+         * the correct identifier property for creating SVG elements is
+         * 'createElementNS'
+         */
+        linter.defineRule('svg-create', {
+            create(eslintContext) {
+                return {
+                    CallExpression(node) {
+                        const exp = node as ESTree.CallExpression;
```
antross: @molant and @sarvaje I'd be curious to hear your thoughts as well, but if we rely a bit more on duck-typing we can help TypeScript figure out the types automatically instead of requiring casts (it even removes the need to import `ESTree`). 

Also I'm a fan of using conditional returns at the beginning of functions to clearly call out separate checks and reduce nesting:
```javascript
CallExpression(node) {
    if (!('callee' in node && 'property' in node.callee && 'name' in node.callee.property)) {
        return;
    }

    if (node.callee.property.name !== 'createElement') {
        return;
    }

    const arg = node.arguments[0];

    if (arg && 'value' in arg && typeof arg.value === 'string' && svgElements.has(arg.value.toLowerCase())) {
        eslintContext.report({
            messageId: 'avoidElement',
            node: node.callee.property
        });
    }
}
```

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,96 @@
...
+                                eslintContext.report({
+                                    messageId: 'avoidElement',
+                                    node: property
+                                });
+                            }
+                        }
+                    }
+                };
+            },
+            meta: { messages: { avoidElement: 'Avoid using createElement to create SVG elements; use createElementNS instead' } }
```
antross: Maybe make this a bit stronger since not only should you avoid it - it won't work.
```suggestion
            meta: { messages: { avoidElement: 'SVG elements cannot be created with createElement; use createElementNS instead' } }
```

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,96 @@
...
+            for (const result of results) {
+                let position: ProblemLocation | null = scriptElement.getLocation() || null;
+                let line = 0;
+
+                /*
+                 * Eslint returns location data starting at <script>
+                 * This offsets the line number so that it is relative
+                 * to the full source code and not just the <script> tag
+                 */
+                if (scriptElement) {
```
antross: What happens if we encounter a script element referencing an external JavaScript file? We don't want an offset in that case... Probably a good idea to also check that the resource is internal (inline).
```suggestion
                if (scriptElement && scriptData.resource === 'Internal javascript') {
```

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        const validateScript = async (scriptData: ScriptParse) => {
+
+            debug(`Validating hint create-element-svg`);
+
+            const sourceCode = scriptData.sourceCode;
+            const results = linter.verify(sourceCode, { rules: { 'svg-create': 'error' } });
+
+            for (const result of results) {
+                let position: ProblemLocation | null = scriptElement.getLocation() || null;
```
antross: Redundant type definition.
```suggestion
                let position = scriptElement.getLocation() || null;
```

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,96 @@
...
+                }
+                await context.report(scriptData.resource, result.message, { location: { column: result.column, line: line + result.line }});
+            }
+        };
+
+        const enterScript = (elementFound: ElementFound) => {
+            scriptElement = elementFound.element;
+        };
+
+        context.on('element::script', enterScript);
```
antross: This is more of a nit and @molant may disagree, but I'd favor registering anonymous functions directly for both `element::script` and `parse::end::javascript` since we only register once and never deregister. This also allows TypeScript to correctly infer parameter types so you have to import even fewer things. For example:
```suggestion
        context.on('element::script', (elementFound) => {
            scriptElement = elementFound.element;
        });
```

packages/hint-create-element-svg/README.md
```diff
@@ -0,0 +1,80 @@
+# No `createElement` with SVG (`create-element-svg`)
+
+This hint informs users that they need to use [`createElementNS`][createElementNS] to create SVG elements instead of [`createElement`][createElement]
+
+## Why is this important?
+
+XML documents may contain elements and attributes that are defined for and used by multiple softwares.
```
antross: Discussion of `XML` may be confusing here as most developers will get this warning when building `HTML` documents. Here's some proposed wording that focuses more on `HTML` instead.
```md
SVG-in-HTML is a fantastic addition to the web platform, but since SVG
is an XML-based language there is some nuance to how it can be used.
When parsing HTML, SVG elements are automatically created correctly so
long as they are inside an `<svg>...</svg>` block. However, SVG elements
cannot be *dynamically* created using `createElement` in the same way
as HTML elements.

For example, calling `document.createElement('circle')` actually returns
an *HTML* element named `circle` instead of an SVG element. This is for
compatibility with existing web content which may have created custom
(although invalid) HTML elements using these same names.

In order to dynamically create SVG elements, you must explicitly tell the
browser you want SVG by using the [SVG namespace][svg namespace] with
`createElementNS`. For example, to create an SVG `circle` call
`document.createElementNS('http://www.w3.org/2000/svg', 'circle')`.
```

packages/hint-create-element-svg/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+The XML namespace helps standardize these naming constructs and provides other useful information about them.
+For example, in case of SVG elements, using `createElement` to create a `circle` element will be
+misinterpreted by the browser as an invalid HTML element with the `<circle>` tag. However, creating the
+`circle` object using `createElementNS` along with the correct [SVG namespace][svg namespace] will help the
+browser correctly interpret it as an SVG element.
+
+## What does the hint check?
+
+This hint scans the abstract syntax tree to check if any SVG objects are created using `createElement`
```
antross:
```suggestion
This hint scans JavaScript source code to check if `createElement` is called with any known SVG element names.
```

packages/hint-create-element-svg/README.md
```diff
@@ -0,0 +1,80 @@
+# No `createElement` with SVG (`create-element-svg`)
+
+This hint informs users that they need to use [`createElementNS`][createElementNS] to create SVG elements instead of [`createElement`][createElement]
```
antross:
```suggestion
This hint informs users that they need to use [`createElementNS`][createElementNS] to create SVG elements instead of [`createElement`][createElement].
```

packages/hint-create-element-svg/src/hint.ts
antross: @molant I was thinking we should export a constant with the `''Internal javascript'` value from `parser-javascript` for just this kind of check. What do you think?

molant: The less magic strings the better :)
Maybe as part of a different PR?

packages/hint-create-element-svg/src/hint.ts
molant: Not going to fight you on this one 😛

packages/hint-create-element-svg/package.json
```diff
@@ -0,0 +1,71 @@
...
+    ],
+    "timeout": "1m"
+  },
+  "description": "Infrom users that they need to use createElementNS to create SVG elements instead of createElement",
+  "devDependencies": {
+    "@hint/parser-javascript": "^2.0.3",
+    "@hint/utils-tests-helpers": "^2.0.3",
+    "@types/debug": "0.0.31",
+    "@types/node": "10.12.18",
+    "ava": "^0.25.0",
```
molant: This version is wrong and should be 1.x something. This is probably related to the issue you were having running `profile.js`. There will probably be some other outdated dependencies. Can you please update them?
I'll try to do a release today with the updated `create-hint` package if #1937, #1944 and #1950 get merged soon.
@antross can you please take a look at those when you have time?

packages/hint-create-element-svg/src/hint.ts
antross: Agreed on both points.

I was thinking a bit more about this and perhaps we should include an `IAsyncHTMLElement` as part of the `ScriptParse` event when the script was inline? That way we also have the context we need to get the location without having to tie things together across two events (eliminating issues with cached state, etc.)

molant: Sounds good to me, although hopefully it will be an `IHTMLElement` (do we need the interface now that everyone should be using the same parser?). Want to open a PR for that?

antross: Opened as #1953.

utsavized:
@antross Does https://github.com/webhintio/hint/pull/1953 mean that I don't have the subscribe to `element::script` anymore, and have access to the element directly from ScriptParse object?

packages/hint-create-element-svg/package.json
utsavized: @molant This fixed the ava issue I was having -- thanks! :)

utsavized:
@antross @molant @sarvaje PR Feedback addressed. Thanks!

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,95 @@
...
+
+                let position = element.getLocation() || null;
+                let line = 0;
+
+                /*
+                 * Eslint returns location data starting at <script>
+                 * This offsets the line number so that it is relative
+                 * to the full source code and not just the <script> tag
+                 */
+                if (scriptData.resource === 'Internal javascript') {
```
antross: We shouldn't need to check the string anymore as `scriptData.element` is only set when the resource was internal. I think we just need to check that both `element` exists and `element.getLocation()` returns a result, then we can use that returned result to adjust the position accordingly (we shouldn't need to call both `getLocation` and `findProblemLocation`).

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,95 @@
...
+            debug(`Validating hint create-element-svg`);
+
+            const sourceCode = scriptData.sourceCode;
+            const results = linter.verify(sourceCode, { rules: { 'svg-create': 'error' } });
+
+            for (const result of results) {
+                const element = scriptData.element;
+
+                if (element === null) {
+                    return;
```
antross: Not sure we should be returning here as `element` won't be set when the script was external.

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -70,26 +68,28 @@ export default class CreateElementSvgHint implements IHint {
...
                 /*
                  * Eslint returns location data starting at <script>
                  * This offsets the line number so that it is relative
                  * to the full source code and not just the <script> tag
                  */
-                if (scriptElement && scriptData.resource === 'Internal javascript') {
-                    position = await findProblemLocation(scriptElement, { column: 0, line: 0 }, result.source === null ? '' : result.source);
+                if (scriptData.resource === 'Internal javascript') {
+                    position = await findProblemLocation(element, { column: 0, line: 0 }, result.source === null ? '' : result.source);
                     line = position !== null ? position.line : 0;
```
antross: Does this work if we have `<script>document.createElement("circle")</script>` all on one line? We may need to adjust `column` in addition to `line`. This probably also means we need to account for the size of the opening script tag, e.g. in `<script type="text/javascript">document.createElement("circle")</script>`.

antross: Looks like `parse5` includes [location information for the end of the start tag](https://github.com/inikulin/parse5/blob/master/packages/parse5/docs/source-code-location/start-tag-location.md) of an element. We should be able to extend our internal `HTMLElement` type to expose this once #1943 is merged.

packages/hint-create-element-svg/README.md
```diff
@@ -0,0 +1,87 @@
...
+(although invalid) HTML elements using these same names.
+
+In order to dynamically create SVG elements, you must explicitly tell the
+browser you want SVG by using the [SVG namespace][svg namespace] with
+`createElementNS`. For example, to create an SVG `circle` call
+`document.createElementNS('http://www.w3.org/2000/svg', 'circle')`.
+
+## What does the hint check?
+
+This hint scans JavaScript source code to check if `createElement` is called with any known SVG element names.
```
antross: This line probably needs to be broken in two to pass the `yarn lint` check.
```suggestion
This hint scans JavaScript source code to check if `createElement` is called
with any known SVG element names.
```

packages/hint-create-element-svg/src/hint.ts
antross: We still need to delete this `if` block. We can pass `scriptData.element` directly to `context.report`, but definitely still need to log even if it's `null` or we won't report for external scripts.

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,81 @@
...
+                                node: node.callee.property
+                            });
+                        }
+                    }
+                };
+            },
+            meta: { messages: { avoidElement: 'SVG elements cannot be created with createElement; use createElementNS instead' } }
+        });
+
+        const validateScript = async (scriptData: ScriptParse) => {
```
antross: Just a nit, but we've started using this pattern in other hints for convenience. Then you can delete `const sourceCode` and `const element` below.
```suggestion
        const validateScript = async ({ element, sourceCode }: ScriptParse) => {
```

packages/hint-create-element-svg/src/hint.ts
```diff
@@ -0,0 +1,81 @@
...
+
+            for (const result of results) {
+                const element = scriptData.element;
+
+                if (element === null) {
+                    return;
+                }
+
+                // ESLint location is 1-based
+                const loc = { column: result.column - 1, line: result.line - 1};
```
antross: Nit: renaming this `location` would allow the `context.report` call to use `{ element, location }` instead of `{ element, location: loc }`.

packages/hint-create-element-svg/tests/tests.ts
```diff
@@ -0,0 +1,49 @@
...
+        name: 'Should not use createElement to create Circle SVG element',
+        reports: [{
+            message: 'SVG elements cannot be created with createElement; use createElementNS instead',
+            position: { match: 'createElement("circle")' }
+        }],
+        serverConfig: generateHTMLPageWithDivTag('', generateScriptTag(invalidCircleCreate))
+    }
+];
+
+hintRunner.testHint(hintPath, tests, { parsers: ['javascript']});
```
antross: Probably also good to add one test validating this works as expected with external script files. You can check out the tests in `hint-css-prefix-order` for an example of how to set up a `serverConfig` with external files.

packages/hint-create-element-svg/src/hint.ts
antross: Also I think `eslint` is going to complain when you run `yarn lint` unless this is declared across multiple lines.
```suggestion
                const location = {
                    column: result.column - 1,
                    line: result.line - 1
                };
```

packages/hint-create-element-svg/tests/tests.ts
```diff
@@ -0,0 +1,69 @@
...
+const hintPath = getHintPath(__filename);
+
+const generateHTMLPageWithDivTag = (head: string, body: string) => {
+    return generateHTMLPage(head, `<div id="container">${body}</div>`);
+};
+const generateScriptTag = (script: string) => {
+    return `<script type="text/javascript">${script}</script>`;
+};
+const generatePageWithExternalScript = () => {
+    const externalJS = readFile(`${__dirname}/js/testjs.js`);
```
antross: Your tests are failing because the test runner is trying to run your external JS file as well. Move this to a `fixtures` folder to ensure it gets ignored.
```suggestion
    const externalJS = readFile(`${__dirname}/fixtures/testjs.js`);
```

Issue #1643, opened 2019-01-09
Release improvements


molant:
The release process is already semi-automatic (thanks @alrra 🙏) but there's always room for improvements. This Epic tracks the work (and discussions) to make it less painful.

molant:
# 🚀 Release process automation and improvements

## Background

[hint][hint] is a monorepo. This means it contains several packages under the same source control so it is easier to update them. In this case all those packages are under the `packages` folder: main CLI, `hint`s, `connector`s, `formatter`s, `configuration`s, etc. Everything maintained by the team lives there.

Releasing a monorepo following `semver` can be a challenge. How to bump versions? How to deal with circular dependencies?

This document aims to detail what webhint's current release process does, what are the pain points, and a proposal on how to aleviate them.

## Bump versioning using commit tags

Commit messages are used to decide how to bump the package's version:

* `Docs:` no version change, related to `.md` changes
* `Build:` no version change, related to build and release scripts
* `Chore:` refactoring, tests, etc. anything not user-facing will be a `patch` version change (x.x.X)
* `Upgrade:` dependency updates with no public API changes are a `patch` version change (x.x.X)
* `Fix:` bug fixes are a `patch` version change (x.x.X)
* `New:` added functionality that does not break previous behavior is a `minor` version change (x.X.x)
* `Breaking:` any change to the public API or output (like error messages) is a `major` version change (X.x.x)

## Automatic bump versioning of local dependencies

The release process updates local dependencies directly using `Upgrade` in the commit message. The reason is that [no code changes are needed to upgrade thus a `patch`](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api). If any code modification is required it should be done prior running the release script.

**⁉️ NOTE**: Publishing a new version shouldn't be done if the only change since it was last published is an `Upgrade` of dependency(ies).

Let's imagine we have `packageB`, v1.5.0, that depends on `packageA`, v1.0.5. The following are some scenarios of what should be done:

### No public changes

* `packageA` adds support for something new → `New: Support for this new thing`.
* `packageB` needs `packageA`'s new feature to fix something that was broken → `Fix: This nasty bug`.

The release script will:
* release `packageA` with a `minor` bump: version 1.1.0.
* add a new commit to `packageB`: `Upgrade: Bump packageA to 1.1.0`. Because the commit history has `Fix` and `Upgrade`, `packageB` will be bumped to 1.5.1 (`Fix` == `Upgrade`).

[...]

Issue #210, opened 2017-05-16
Tool to automatically open issues when TODO is found


molant:
There are quite a few `TODO` comments in the code and no issues associated. We should have a script that is run in travis (`master`) that:
* looks for all the `TODO` comments in the JS code (we probably just split by line)
* checks if the comment has a `#ID` somewhere or a url to the GitHub issue
  * if yes it doesn't do anything
  * if not
    * it creates a new issue in GitHub with the title as the `TODO` comment, and the body with the name of the file and the line number
    * adds the `#id` to the end of the comment and commits the changes

I don't think it will be super complicated and can probably be a separate project.\
There are some projects that do similar things, check [todo search in npm](https://www.npmjs.com/search?q=todo%20github&page=1&ranking=optimal). My 5 minute search didn't trigger anything like what I want but we should look a bit more into it.

molant:
I've created [this label](https://github.com/sonarwhal/sonar/labels/TODO-bot) for the issues created automatically.

PR #1943, opened 2019-02-20
Traverse

sarvaje:
## Pull request checklist

Make sure you:

The bundle size using this solution increase less than 500kb.

I still have to test it and see that everything is working as expected.

Also I need to polish the code, right now I just wanted to have something working to see the bundle size.

@molant @antross any feedback would be great.


packages/connector-jsdom/src/connector.ts
```diff
@@ -293,9 +254,11 @@ export default class JSDOMConnector implements IConnector {
...
                     debug(`${this.finalHref} loaded, traversing`);
                     try {
-                        await this.server.emitAsync('traverse::start', event);
-                        await this.traverseAndNotify(window.document.children[0] as HTMLElement);
-                        await this.server.emitAsync('traverse::end', event);
+                        const html = await this.html;
+
+                        const asyncWindow = createAsyncWindow(html);
+
+                        await traverse((asyncWindow as JSDOMAsyncWindow).dom!, this.server, this.finalHref);
```
molant: Do we need the `JSDOMAsyncWindow` and such with the new approach? the cdp connector doesn't seem to need this.

packages/hint/src/lib/utils/dom/new-traverse.ts
```diff
@@ -0,0 +1,38 @@
+import { NewAsyncHTMLDocument, NewAsyncHTMLElement } from '../../types/new-async-html';
```
molant: change the filename to `traverse` and similar with `new-async-html`?

packages/hint/src/lib/utils/dom/create-new-async-document.ts
```diff
@@ -0,0 +1,13 @@
+import { IAsyncHTMLDocument } from '../../types/async-html';
```
molant: We used `async` because that's how CDP works but with this probably we don't need to make the API async. It's probably a bit more changes but maybe we should make it `sync` again and update the names?

packages/hint/src/lib/types/new-async-html.ts
```diff
@@ -0,0 +1,97 @@
...
+            line: location.startLine
+        };
+    }
+
+    public isSame(element: NewAsyncHTMLElement): boolean {
+        return this._element === element._element;
+    }
+
+    public outerHTML(): Promise<string> {
+        return Promise.resolve(parse5.serialize(this._element, { treeAdapter: htmlparser2Adapter }));
```
molant: Going back to the subject of making it `sync` again, this should be removed (same in the rest of the file).

packages/connector-jsdom/src/connector.ts
sarvaje: Working on it.

packages/hint/src/lib/utils/dom/new-traverse.ts
sarvaje: working on it. I'm calling them `shadow-traverse`, `ShadowAsyncHTMLDocument`, etc.

packages/hint/src/lib/utils/dom/create-new-async-document.ts
sarvaje: sounds good.

packages/hint/src/lib/types/new-async-html.ts
sarvaje: what do you mean for 'remove'? do you mean the Promise.resolve thing or the whole outerHTML?

packages/hint/src/lib/utils/dom/new-traverse.ts
molant: Not sure about the `shadow` part as it could be confused with `shadow-dom`. I might just leave it as `traverse`, `HTMLDocument` and such. @antross what do you think?

packages/hint/src/lib/types/new-async-html.ts
molant: Remove the `Promise.resolve` so it is sync. Sorry about that.

sarvaje: Ah ok, yes, there is some `Promise.resolve` to remove :)

packages/hint/src/lib/utils/dom/new-traverse.ts
antross: I agree that `shadow` is too likely to be confused with `shadow-dom`. Let's stick with `traverse` for now unless someone has another idea.

packages/hint/src/lib/utils/dom/traverse.ts
```diff
@@ -0,0 +1,36 @@
...
+        element: new JSDOMAsyncHTMLElement(element, dom),
+        resource
+    });
+
+    const traverseEvent = {
+        element: new JSDOMAsyncHTMLElement(element, dom),
+        resource
+    } as TraverseDown | TraverseUp;
+
+    await engine.emitAsync(`traverse::down`, traverseEvent);
```
antross: Random thought, but should we just get rid of `traverse::down` and `traverse::up`? Seems like a lot of extra events that nothing listens to (with the sole exception of `analyze.ts` which sets a status message and would probably be just fine without them).

I realize this would be breaking, so I can file and track it in a separate issue if everyone agrees (although I'm betting this PR will be breaking already so maybe that's not an issue...)

molant: I'd rather not remove them TBH. Us not using them doesn't mean we will not later or that they aren't being used by someone else.

These events could be useful to know if an element is inside another one and similar checks.

packages/connector-jsdom/src/connector.ts
```diff
@@ -421,7 +385,7 @@ export default class JSDOMConnector implements IConnector {
     }
 
     /* istanbul ignore next */
-    public querySelectorAll(selector: string): Promise<JSDOMAsyncHTMLElement[]> {
+    public querySelectorAll(selector: string): IHTMLElement[] {
```
molant: I think I mentioned it ealier but now that we use the same implementation everywhere we don't need an Interface, we can use directly a class or a type. Same goes for `IHTMLDocument` (looks like we also have an `HTMLDocument`)

packages/hint-doctype/src/hint.ts
```diff
@@ -29,8 +29,8 @@ export default class implements IHint {
             line: 0
         };
 
-        const report = async (resource: string, message: string, location = defaultProblemLocation): Promise<void> => {
-            await context.report(resource, message, { location });
+        const report = (resource: string, message: string, location = defaultProblemLocation) => {
```
molant: Side effect I wasn't expecting from this change 🎉

This is a pretty big breaking change so we probably want to get #1643 done first before publishing to make sure everything is right.

sarvaje:
@antross @molant  I think this is ready to review

packages/parser-html/src/parser.ts
```diff
@@ -26,82 +26,30 @@ try {
     // `canvas` is not installed, nothing to do
 }
 
-import { JSDOM } from 'jsdom';
-import { JSDOMAsyncHTMLElement, JSDOMAsyncWindow } from 'hint/dist/src/lib/types/jsdom-async-html';
-import { Event, FetchEnd, Parser, TraverseDown, TraverseUp } from 'hint/dist/src/lib/types';
+import { Parser, FetchEnd } from 'hint/dist/src/lib/types';
 import { Engine } from 'hint';
+import createAsyncWindow from 'hint/dist/src/lib/utils/dom/create-async-window';
```
antross: Should this now be `createHTMLDocument` instead? I was under the impression `parser-html` wouldn't be referencing `jsdom` with these changes...

sarvaje: No, in this case. The `parser-html` needs the jsdom so we can have an `evalute`.

molant:
Wasn't evaluate going to be done at the connector level?


sarvaje:
AFAIK, no, the parser will have the evalute.

molant:
Aghhh email answers don't put the comment in the right converation.

I thought we agreed that connectors should do the evaluate and `parser-html` the traversing and element location. Browsers have better API support, the jsdom connecter already generates a context for this so we will be duplicating and the goal of this PR was to not have jsdom in the browser extension neither 😕

Maybe we should all sync about this?

sarvaje:
@molant maybe there is a little confusion here.

The connectors will use their own evalute, but `parse-html` will allow to run scripts using jsdom, so the `connector-local` will be able to run scripts.

The traverse was remove from the `parser-html`.

Now the traverse is just an `util` inside `hint`.

packages/connector-jsdom/src/resource-loader.ts
```diff
@@ -29,9 +46,10 @@ export default class CustomResourceLoader extends ResourceLoader {
             return await promise;
         }
 
+        const elements = options.element ? this._HTMLDocument.querySelectorAll(this.createQuery(options.element)) : null;
```
antross: Can we add a comment explaining why we're doing this?

packages/extension-browser/src/content-script/connector.ts
```diff
@@ -69,31 +70,14 @@ export default class WebExtensionConnector implements IConnector {
...
-            await this.traverseAndNotify(node.children[i], doc);
-        }
-
-        await this._engine.emitAsync(`traverse::up`, { element, resource });
-    }
-
     private setFetchElement(event: FetchEnd) {
         const url = event.request.url;
-        const elements = Array.from(document.querySelectorAll('[href],[src]')).filter((element: any) => {
+        const elements = Array.from(this._document.querySelectorAll('[href],[src]')).filter((element: any) => {
```
antross: You also needed to resolve the original element for `FetchEnd` in other connectors too, right? Should this be a shared helper? If so, we may also need to include `poster`, `data`, and `srcset` for completeness.

sarvaje:
@molant @antross test passing!! (at least on Linux hehehe)

packages/hint/src/lib/types/html.ts
```diff
@@ -0,0 +1,120 @@
...
+         * https://github.com/inikulin/parse5/issues/118
+         *
+         * The problem with this workaround will modify the
+         * parentElement and parentNode of the element, so we
+         * need to restore it before return the outerHTML.
+         */
+        const fragment = htmlparser2Adapter.createDocumentFragment();
+        const parent = this._element.parent;
+
+        htmlparser2Adapter.appendChild(fragment, this._element);
```
antross: Does this potentially alter order relative to siblings as well when we restore it?

packages/hint/src/lib/types/html.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        return htmlNode;
+    }
+
+    public pageHTML(): string {
+        return this._pageHTML;
+    }
+
+    public querySelectorAll(selector: string): HTMLElement[] {
+        // const matches = Array.from(this._document.querySelectorAll(selector));
```
antross:
```suggestion
```

packages/parser-html/README.md
```diff
@@ -34,14 +34,19 @@ And then activate it via the [`.hintrc`][hintrc] configuration file:
...
   parsed document.
 * `html`: a string containing the raw HTML source code.
 * `resource`: the parsed resource.
 
+And the event `parse::start::html` of type `Event` which has the
+following information:
+
+* `resource`: the resource that is going to be parsed.
+
 This `parser` also automatically traverses and emits events for
```
antross: We should remove this section since `parser-html` doesn't traverse any more.

packages/extension-browser/src/content-script/connector.ts
sarvaje: Are you thinking in a new `util` in `hint`?

packages/hint/src/lib/types/html.ts
sarvaje: Nope.

sarvaje: I will take another look, but IIRC, this only alter the parent element.

packages/extension-browser/src/content-script/connector.ts
sarvaje: I'm taking a look to this and they are not compatible (at least not now). In jsdom, we have the element that is fetching the content, but not in the browser extension.

Also, I found another problem here, in case of a relative url, event.request.url will contain the full url (domain + relative url), but the element will contain the relative url, so this code won't work.

I'm thinking in something like `url.endsWith(element.getAttribute('src')) || ....`.

Maybe there is another way, but I don't know this code too much.

What do you think @antross?

sarvaje: > If so, we may also need to include `poster`, `data`, and `srcset` for completeness.

@antross what elements use the attribute 'data'

molant: [From `https-only`](https://github.com/webhintio/hint/blob/3dc05691a36fd8d46f32634d964796c68440432c/packages/hint-https-only/src/hint.ts#L124):
```js
/*
 * Possible URL sources:
 *
 * * <img>, <audio>, <video>, <source> and <track> can have a `src` attribute.
 *   E.g.: <img src="http://example.com/image.jpg">
 *
 * * <img> and <source> can have a `srcset` attribute.
 *   https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source
 *   E.g.: <img src="image-src.png" srcset="image-1x.png 1x, image-2x.png 2x">
 *
 * * <video>  can have a `poster` attribute.
 *   E.g.: <video width="480" controls poster="https://archive.org/download/WebmVp8Vorbis/webmvp8.gif" />
 *
 * * <object> has a `data` attribute.
 *   E.g.: <object data="movie.swf" type="application/x-shockwave-flash"></object>
 *   https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object
 */
```

There's also `srcset` that uses another format that we should probably take into consideration.

sarvaje: I'm taking into consideration this format `image-1x.png 1x, image-2x.png 2x`. There is any other format?

packages/hint/src/lib/types/html.ts
sarvaje: At least for now, the only change the parent property: (prev is always empty, because we are using a new fragment everytime.
```js
const appendChild = (exports.appendChild = function(parentNode, newNode) {
    const prev = parentNode.children[parentNode.children.length - 1];

    if (prev) {
        prev.next = newNode;
        newNode.prev = prev;
    }

    parentNode.children.push(newNode);
    newNode.parent = parentNode;
});
```

Anyways, I think I'm going to add code to restore `parent`, `prev` and `next`, just in case this code change in the future.

packages/hint/src/lib/types/html.ts
```diff
@@ -74,13 +76,15 @@ export class HTMLElement {
...
-        const parent = this._element.parent;
+        const { parent, next, prev } = this._element;
 
         htmlparser2Adapter.appendChild(fragment, this._element);
 
         const result = parse5.serialize(fragment, { treeAdapter: htmlparser2Adapter });
 
         this._element.parent = parent;
+        this._element.next = next;
+        this._element.prev = prev;
```
antross: For completeness you'll also want to set `next.prev = this._element` and `prev.next = this._element` (after checking to ensure they're not `null` of course).

sarvaje: done!

packages/hint/src/lib/types/jsdom-async-html.ts
```diff
@@ -134,6 +134,10 @@ export class JSDOMAsyncWindow implements IAsyncWindow {
         return this._document;
     }
 
+    public get dom(): JSDOM | undefined {
```
antross: Shouldn't this whole file be able to be deleted since we're only using `JSDOM` indirectly for evaluation purposes now? In the JSDOM connector direct access is available and we don't need an intermediate abstraction layer.

packages/hint/src/lib/utils/dom/create-async-window.ts
```diff
@@ -0,0 +1,9 @@
+import { IAsyncWindow } from '../../types/async-html';
+import { JSDOMAsyncWindow } from '../../types/jsdom-async-html';
+import createJsdom from './create-jsdom';
+
+export default (html: string, allowScripts: boolean = false): IAsyncWindow => {
```
antross: Given my other comments about removing the `IAsync*` interfaces, can we have this return something more scoped that only provides access to evaluate scripts?

sarvaje:
@antross done!

packages/hint/src/lib/types/html.ts
```diff
@@ -0,0 +1,131 @@
...
+import {
+    INamedNodeMap,
+    ProblemLocation
+} from '../types';
+
+type Attrib = {
+    [key: string]: string;
+};
+
+export type ParsedHTMLElement = Element & {
```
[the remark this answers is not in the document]
sarvaje: > Does a ParsedHTMLElement actually include support for everything Element defines (since this seems to refer to the built-in Element type)?

Nop. It support just some `element` properties.

packages/hint/src/lib/types/html.ts
```diff
@@ -0,0 +1,131 @@
...
+        this.ownerDocument = ownerDocument;
+    }
+
+    public get attributes(): INamedNodeMap {
+        const x = this._element.attribs;
+
+        return Object.entries(x).map((y) => {
+            return {
+                name: y[0],
+                value: y[1] as string
```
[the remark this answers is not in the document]
sarvaje: It looks like that casting is outdated. We don't need it anymore.

antross:
Also, I took this for a spin with `extension-browser` which still worked as advertised. Test bundle here if anyone else wants to try:
* [extension-browser-pr-1943.zip](https://github.com/webhintio/hint/files/2929276/extension-browser-pr-1943.zip)

sarvaje:
@antross does the extension needs a zero based location?

If I can reproduce your error, but If I use the local connector for the same file, I get this result:
image

As you can see, the locations are good.

P.S.: In this capture, the issue with `axe` is fixed hehehe.

sarvaje:
Yeah!

> thanks @sarvaje for being patient with me on this one! 

hehehe, no problem!!

packages/hint-sri/src/hint.ts
```diff
@@ -120,14 +120,14 @@ export default class SRIHint implements IHint {
...
-    private async isEligibleForIntegrityValidation(evt: FetchEnd, urls: URLs): Promise<boolean> {
+    private isEligibleForIntegrityValidation(evt: FetchEnd, urls: URLs): Promise<boolean> {
         debug('Is eligible for integrity validation?');
 
         const { element, resource } = evt;
         const resourceOrigin: string = new URL(resource).origin;
 
         if (urls.origin === resourceOrigin) {
-            return true;
+            return Promise.resolve(true);
```
molant: We should remove all the `Promise.resolve` that are no longer necessary. I've opened https://github.com/webhintio/hint/issues/2035 to track this so no need to do anything in this PR.

packages/hint-validate-set-cookie-header/src/hint.ts
```diff
@@ -296,20 +295,22 @@ export default class ValidateSetCookieHeaderHint implements IHint {
...
                     parsedSetCookie.resource = resource;
                 } catch (err) {
-                    await context.report(resource, err.message, { element });
+                    context.report(resource, err.message, { element });
 
                     return;
                 }
 
-                for (const defaultValidator of defaultValidators) {
+                defaultValidators.every((defaultValidator) => {
```
molant: I think we should do a `forEach` in here, otherwise we will just report the first batch of `messages` and stop processing right after which is not what we were doing before.

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -3,8 +3,9 @@ import { URL } from 'url';
 import * as path from 'path';
 
 import * as async from 'async';
-import { default as ora, Ora }  from 'ora';
+import { default as ora, Ora } from 'ora';
 const boxen = require('boxen'); // `require` used because `boxen` exports a function
+
```
molant: Nit: remove this extra empty line that you added.

packages/hint/docs/contributor-guide/how-to/connector.md
```diff
@@ -74,49 +71,10 @@ export interface IConnector {
     /** Evaluates the given JavaScript `code` asynchronously in the target. */
     evaluate(code: string): Promise<any>;
     /** Finds all the nodes that match the given query. */
-    querySelectorAll(query: string): Promise<IAsyncHTMLElement[]>
+    querySelectorAll(query: string): ITMLElement[]
```
molant:
```suggestion
    querySelectorAll(query: string): HTMLElement[]
```

packages/hint/docs/contributor-guide/how-to/hint.md
```diff
@@ -93,15 +93,15 @@ with the website and report errors.
 To report an error, the hint has to do the following:
 
 ```ts
-await context.report(resource, message, { element: element });
+context.report(resource, message, { element: element });
 ```
 
-* `context.report()` is an asynchronous method, you should always `await`.
+* `context.report()` is a synchronous method, you don't need to `await`.
```
molant: Nit: We can probably remove this line, I think we needed before to make it clear but being sync is the expected thing now.

packages/hint/package.json
```diff
@@ -16,19 +16,23 @@
...
     "content-type": "^1.0.4",
+    "css-select": "^2.0.2",
     "debug": "^4.1.1",
     "eventemitter2": "^5.0.1",
     "file-type": "^10.9.0",
     "file-url": "^2.0.2",
     "globby": "^9.1.0",
     "is-ci": "^2.0.0",
     "is-svg": "^4.0.0",
+    "jsdom": "^13.2.0",
```
molant: Do we need `jsdom` in here? I don't see it being used in `hint`.

packages/hint/src/lib/utils/dom/get-element-by-url.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        return total;
+    }, [] as string[]);
+
+    return urls;
+};
+
+export default (dom: HTMLDocument, url: string): HTMLElement | null => {
+    const elements = dom.querySelectorAll('[href],[src],[poster],[srcset]').filter((element: any) => {
+        const elementUrl = element.getAttribute('href') || element.getAttribute('src') || element.getAttribute('poster');
```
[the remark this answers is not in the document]
sarvaje: I don't think this is going to be called more than once with the same url (if so, is because the user has something duplicated is the code). But we can cache the result of the querySelectorAll.

packages/connector-jsdom/src/resource-loader.ts
```diff
@@ -4,22 +4,39 @@ import { debug as d } from 'hint/dist/src/lib/utils/debug';
...
 
-    public async fetch(url: string, options: { element: HTMLElement }): Promise<Buffer | null> {
+    private createQuery(element: Element): string {
+        let query: string = element.tagName;
+
+        for (let i = 0; i < element.attributes.length; i++) {
+            const attribute = element.attributes.item(i);
+
+            if (attribute) {
+                query += `[${attribute.name}="${attribute.value}"]`;
```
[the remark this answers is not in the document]
sarvaje: I think we can use `get-element-by-url.ts` to do this.

packages/connector-jsdom/src/resource-loader.ts
```diff
@@ -29,9 +46,15 @@ export default class CustomResourceLoader extends ResourceLoader {
             return await promise;
         }
 
+        /*
+         * We need a `HTMLElement` instead of the element returned by jsdom.
+         * To do so, we create a query from the element returned by jsdom and
+         * look for it in our `HTMLDocument`.
+         */
+        const elements = options.element ? this._HTMLDocument.querySelectorAll(this.createQuery(options.element)) : null;
```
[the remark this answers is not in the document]
sarvaje: yes, done :)

packages/hint-highest-available-document-mode/src/hint.ts
```diff
@@ -90,10 +90,10 @@ export default class HighestAvailableDocumentModeHint implements IHint {
 
         };
 
-        const checkMetaElement = async (resource: string) => {
+        const checkMetaElement = (resource: string) => {
 
-            const pageDOM: IAsyncHTMLDocument = context.pageDOM as IAsyncHTMLDocument;
-            const XUACompatibleMetaElements: IAsyncHTMLElement[] = getXUACompatibleMetaElements(await pageDOM.querySelectorAll('meta'));
+            const pageDOM: HTMLDocument = context.pageDOM as HTMLDocument;
+            const XUACompatibleMetaElements: HTMLElement[] = getXUACompatibleMetaElements(pageDOM.querySelectorAll('meta'));
```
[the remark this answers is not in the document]
sarvaje: we are `normalizing` the attribute value, that why are not using a better selector, but maybe it can be improved :)

packages/hint-no-broken-links/src/hint.ts
```diff
@@ -86,7 +88,7 @@ export default class NoBrokenLinksHint implements IHint {
          * and report if it exist there. We will also add it to the fetchedURLs
          * so that duplicate requests will not be made if 2 links have the same href value
          */
-        const handleSuccess = (networkData: NetworkData, url: string, element: IAsyncHTMLElement) => {
+        const handleSuccess = (networkData: NetworkData, url: string, element: HTMLElement) => {
```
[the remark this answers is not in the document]
sarvaje: Good catch, I missed that.

packages/hint/src/lib/utils/dom/get-element-by-url.ts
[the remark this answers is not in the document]
antross: Probably a premature optimization at this point. I'd wait.

packages/hint-validate-set-cookie-header/src/hint.ts
sarvaje: If that the right behavior? I mean, in the old code, we stopped after the first validation failed, that's why I'm using an `every`

packages/hint/src/lib/cli/analyze.ts
sarvaje: I can't remove that line, because after a  `const` we need an empty line :P.
I move the const below in the code, so all the imports are together :)

packages/hint/package.json
antross: Used by `packages/hint/src/lib/utils/dom/create-jsdom.ts`, but it appears after various refactors this is only consumed by `connector-local`. Probably best to move the code there and remove the dependency from `hint`.

sarvaje: we still have `create-jsdom.ts` in utils/dom. But it is only used in `connector-local` so I think I'm going to remove `create-jsdom.ts` from utils.

packages/hint/src/lib/utils/dom/get-element-by-url.ts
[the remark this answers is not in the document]
molant: We might end up creating the same `new Url` quite a few times depending on how many links the page has.
Can I have a `//TODO` here and hopefully one day we will do #210 ?

packages/connector-jsdom/src/resource-loader.ts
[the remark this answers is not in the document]
molant: SGTM

packages/hint-highest-available-document-mode/src/hint.ts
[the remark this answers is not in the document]
molant: I think now we can use selectors that don't take into consideration the casing and before we couldn't.

packages/hint-validate-set-cookie-header/src/hint.ts
molant: You are right. I missed the `return` right after the `await`.

packages/hint/src/lib/cli/analyze.ts
molant: I need to update how we import `boxen` and remove the types. I've created https://github.com/webhintio/hint/issues/2036 to track it.

packages/hint/src/lib/utils/dom/get-element-by-url.ts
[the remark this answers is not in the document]
sarvaje: > We might end up creating the same new Url quite a few times depending on how many links the page has.

We shouldn't have the same url repeated many often, the only scenario I can think about is when we have the same image in few places.

I will add the //TODO

sarvaje:
@molant I think everything should be done, except the commits stuff

Issue #1956, opened 2019-02-22
[Bug] CSS hints report wrong position for CSS inline in HTML [3.5]


antross:
# 🐛 Bug report

## Description

Save and open the following HTML document in VS Code with the webhint extension:
```html
<!DOCTYPE html>
<html>
    <head>
        <style>            
            .example {
                appearance: none;
                -webkit-appearance: none;
                -moz-appearance: none;
            }
        </style>
    </head>
</html>
```

Expected: Error reported at `appearance: none`.

Actual: Error reported at `<html>`

Affects both `hint-compat-api` and `hint-css-prefix-order`. Should be able to be fixed using additional event data from PR #1953.

PR #2051, opened 2019-03-11
Fix: Tie more elements from DOM snapshot back to original source

antross:
## Pull request checklist

Make sure you:

Ref #1956 

Also fix some cases where use of the snapshot vs. original document were
inconsistent between connectors (except connector-local which needs to
be different).

Notable changes:
* All connectors now use `parser-html` to obtain an original source document
  * Should we consider making `parser-html` fully built-in since HTML handling is so special?
* Snapshots are created with a reference to the original source document for locations
  * This means we no longer use locations from the snapshot itself (I turned off locations for snapshots to ensure this)
  * A new `findOriginalElement` helper was added to take an element from the snapshot and build a selector (plus filter function) to identify the same element in the original source (if one exists).
* If an element was dynamically created, we no longer try to produce a location for it.
  * This avoids producing invalid locations that don't point anywhere in actual source.
  * In future updates some contexts (e.g. the browser extension) may link to the actual element reference instead in these cases.

Marking as "Draft" while I wait for test results which may require some updates.

### DIFF
--- a/packages/connector-jsdom/src/connector.ts
+++ b/packages/connector-jsdom/src/connector.ts
@@ -45,6 +45,7 @@ import { Engine } from 'hint/dist/src/lib/engine';
 import isHTMLDocument from 'hint/dist/src/lib/utils/network/is-html-document';
 import createHTMLDocument from 'hint/dist/src/lib/utils/dom/create-html-document';
 import traverse from 'hint/dist/src/lib/utils/dom/traverse';
+
 import { Requester } from '@hint/utils-connector-tools/dist/src/requester';
 
 import CustomResourceLoader from './resource-loader';
@@ -67,6 +68,7 @@ export default class JSDOMConnector implements IConnector {
     private _targetNetworkData!: NetworkData;
     private _window!: Window;
     private _document!: HTMLDocument;
+    private _originalDocument: HTMLDocument | undefined;
     private _timeout: number;
     private _resourceLoader?: ResourceLoader;
     private _subprocesses: Set<ChildProcess>;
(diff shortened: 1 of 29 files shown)


packages/connector-jsdom/src/connector.ts
```diff
@@ -45,6 +45,8 @@ import { Engine } from 'hint/dist/src/lib/engine';
 import isHTMLDocument from 'hint/dist/src/lib/utils/network/is-html-document';
 import createHTMLDocument from 'hint/dist/src/lib/utils/dom/create-html-document';
 import traverse from 'hint/dist/src/lib/utils/dom/traverse';
+
+import { HTMLEvents } from '@hint/parser-html';
```
molant: Because we are using only the type maybe use [`import types`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#import-types) in loine 89?

antross: Yep, just using the type definitions, but of course an original document will only be set up if HTML parsing is enabled (as we'd only trigger the relevant events in that case).

I'll make the update to inline the type import.

sarvaje: What are the benefits of using `import types`?

molant: I think we should start using those when we only need something for types to make it clear that it's a dev dependency and not required at run time. I was confused with @antross comment:

> All connectors now use `parser-html` to obtain an original source document

when I was reading the code and just seeing the types being used. So basically for clarity 😊

molant:
@antross do you want to do more changes or is this ready to be merged?

antross:
@molant I made one last tweak to fix the issue I discussed earlier where an element with an ID or URL could end up falling back if no match is found and matching (incorrectly) on content. Now this should be good to merge!

packages/connector-jsdom/src/connector.ts
sarvaje: I'm asking, because for me it was weird to see that, I think it makes the code less readable.

molant: But the sections import is cleaner because you only have the imports you need to run the code and not something that is going to be use one time. If the types are going to be used more than once I think we should stick to what we are doing. If it is just once then we should evaluate using this more.

PR #2052, opened 2019-03-11
Breaking: Add API

sarvaje:
## Pull request checklist

Make sure you:

Add an API to `hint` so it will be easy use different parts of the code without knowing the file structure.


sarvaje:
@molant @antross I just want to double check with you that we all are in the same page about what to do for this task.

If you can take a look to this before I continue working on this, it would be great

antross:
That certainly gets everything to the top level. One question I have for all of us now that we've seen this: do we want all of these APIs explicitly exposed as part of our public API surface or if there are certain items we should keep as internal details?

My thinking is the official public API should only include what you need to instantiate and run webhint with the desired configuration (and process the results). This itself should be a new wrapper allowing a single API call to pass a JSON in the same format as `.hintrc` without touching `Engine` or the resource loader directly. 

Notably that would potentially exclude `Engine` itself from being exposed along with a bunch of types and helpers you'd only need when writing parsers, hints, formatters, etc. (although we could still have a separate single-file import for those scenarios). That said, perhaps it's okay to expose those other types so long as we clearly document which one is the wrapper you should use to run webhint. 

Curious to hear what @molant thinks...

packages/hint/src/lib/utils/fs/read-file.ts
```diff
@@ -3,6 +3,6 @@ import { readFileSync } from 'fs';
 import stripBom = require('strip-bom'); // `require` used because `strip-bom` exports a function
 
 /** Convenience wrapper for synchronously reading file contents. */
-export default (filePath: string): string => {
+export const readFile = (filePath: string): string => {
```
molant: Why did you remove the default? Is it to make it easier to refactor?

molant:
I think there are a few scenarios we would like to unblock with this work:

- Separate the CLI from the core. I think we should have a `hint-core` with `Engine` and all the other stuff
- There's been request to analyze multiple URLs, use `sitemap.xml`, etc. This API should make it simple to unblock this scenarios
- Reduce the `import` paths when we create new `hint`s, especially around the utils

As a user what I'd expect is to do something like (names completely arbitrary, we should use something different):

```js
const { webhint } from 'hint-core';
const analyzer = webhint.initialize(jsonconfiguration);
const results = await analyzer.analyze('https://example.com');
```

`webhint` could expose the resources that are going to be used in the analysis. I would expect most people to use only the `formatter`s  (if anything at all). Also `initialize` should fail if there's any problem loading the resources, configuration is incorrect, etc. The usual.

Some other things `hint-core` should expose are the `utils`. We should do an inventory of all the things that are currently being used by several `resouces` and document it in here. We could potentially create another `utils` package so we don't have to do a new release of `hint-core` each time we add something new (although that's another package to manage but in the absolute it might not be that bad).

I think this is a good step but maybe before jumping into the code we should do a bit of research first and write everything down in #1532, starting with what are the most imported things accross all the packages, how do we things the API should look like, etc. Also we should take into account the [feedback from @donaldpipowitch](https://github.com/webhintio/hint/issues/1532#issuecomment-441956399) and ping @renestalder for their feedback as they are the ones that have show initial interest.

sarvaje:
@molant so the API is a new package? What about types? should `hint-core` expose the types too?

molant:
> What about types? should `hint-core` expose the types too?

Most likely, depends on how we shape the API after this research.

sarvaje:
Ahhhhh, I now understand what we want to do with this... hehehehe.

I was understanding that we want to expose everything we have in `hint`. But what we want to do is basically allow people to execute webhint from node xD

packages/hint/src/lib/utils/fs/read-file.ts
sarvaje: This was for 2 reason:
1. It is easy to export.
1. To use the same method name everywhere, I mean, if we are exposing the method: `readFile`, it would be great that we are using also `readFile` everywhere, and not `readFile` in one place `readFileSync` in another, etc. I don't know if I'm explaining myself hehehe.

sarvaje: > if we are exposing the method: readFile, it would be great that we are using also readFile everywhere, and not readFile in one place readFileSync in another, etc. I don't know if I'm explaining myself hehehe.

This is just an example.

PR #2066, opened 2019-03-15
New: Release script improvements

molant:
Draft PR for the release script.

Most of the logic is there except the part of publishing to npm and github, need to figure out how to prompt for the otp multiple times in a clean way (although hopefully we will be using a token). The rest should be working.

There are a few things I'm sure can be improved like the `skipReasons` in `main.ts`. There are quite a few that are repeated but I'm not sure on how to do it better (maybe `skipIfError` is always on when using `skipReasons`?).
I repeat a lot a pattern with observables that I'll probably refactor as well. In any case, I think this is a good moment to start reviewing it.

Another thig is that right now I'm commiting the changes package by package but it's really slow. I'm thinking about just doing one commit with a generic message `Update: Dependencies and package version`. That will save a few minutes. Thoughts?

Things pending:

* [x] npm publishing
* [x] publish changes on github (`git push`)
* [x] delete github token when using user/pass
* [x] test more the `private` and unpublished packages
* [x] test on a mac
* [x] remove dependencies from old script and old script
* [x] test `eslint-plugin-import`
* [x] add help menu
* [x] improve the go back state, maybe prompt if we want to revert or just revert if changes are discarded. The reason is that there might be an issue when publishing to npm or whe puhing to GitHub and we don't want to go back to the start state in those cases

You can give it a try locally, nothing is going to get commited into GitHub or npm. If you don't validate the changes it should revert back to the origin and you can always run `git reset --hard SHA` and you can find the SHA in the `release.log` file


molant:
BTW, the way code is organized is as follows:

`release/main.ts` --> The entry point. I'll probably change the the filename later
`release/tasks/*.ts` --> The tasks executed by `main.ts`
`release/lib/*.ts` --> Helpers and such

Probably best start with main and then go task by task to review this.

release/test.ts
```diff
@@ -0,0 +1,54 @@
...
+            console.log(pkgPaths);
+        }
+    },
+    {
+        title: 'Step 3',
+        task: (ctx) => {
+            return Promise.resolve();
+        }
+    }, {
+        title: 'Step 3',
```
sarvaje: `Step 4`

release/test.ts
```diff
@@ -0,0 +1,54 @@
...
+        task: (ctx) => {
+            process.exit(0);
+        }
+    }]);
+
+tasks.run();
```
sarvaje: New line at the end of the file

release/test.ts
molant: This file needs to be deleted 🤦‍♂️

antross:
> I'm thinking about just doing one commit with a generic message Update: Dependencies and package version. That will save a few minutes. Thoughts?

I think that's preferrable, including the `update-3rd-party` dependencies as well.

molant:
> I think that's preferrable, including the `update-3rd-party` dependencies as well.

Sounds good. I look into it although I might need it for rollback purposes. I'll keep you posted.

release/lib/git-helpers.ts
```diff
@@ -0,0 +1,302 @@
...
+    const pkgText = await fs.readFile(pkg.path, 'utf-8');
+    const versionIndex = pkgText.indexOf('"version":');
+    const line = pkgText.substr(0, versionIndex).split('\n').length;
+    const command = `git blame -L ${line},${line} ${pkg.path}`;
+
+    try {
+        const { stdout } = await execa(command);
+        /**
+         * The format is `sha (Author date) lineContent`.E.g.:
+         * `c10092ade (Tony Ross 2018-08-22 17:39:49 -0500 77)   }`
```
antross: I feel famous! 😎

release/tasks/calculate-new-versions.ts
```diff
@@ -0,0 +1,76 @@
...
+import * as semver from 'semver';
+
+import { debug } from '../lib/utils';
+import updateDependencies from '../lib/update-dependencies';
+
+/**
+ *
+ * @param tag The tag of the commit
+ */
+const commitType = (tag: Tag): Bump => {
```
antross: Maybe rename to something like `getBumpTypeForTag` to be a bit clearer?

release/tasks/calculate-new-versions.ts
```diff
@@ -0,0 +1,76 @@
...
+        case 'Breaking': return Bump.major;
+        default: return Bump.none;
+    }
+};
+
+
+const calculatePackageNewVersion = (pkg: Package, bump: Bump): string => {
+
+    if (!pkg.publishedVersion) {
+        return pkg.content.newVersion;
```
antross: Should this be covered in the types somewhere? `pkg.content` is currently `any` (which means `pkg.content.newVersion` is `any` as well).

release/tasks/calculate-new-versions.ts
```diff
@@ -0,0 +1,76 @@
...
+            return;
+        }
+
+        pkg.content.version = calculatePackageNewVersion(pkg, bumpType);
+    });
+
+    // Step 2
+    updateDependencies(ctx);
+
+    // Step 3
```
antross: Probably also worth repeating what each step does here (even though it's in the comment for the whole function). It took me a second to get my bearings on the first read.

release/@types/custom.ts
```diff
@@ -0,0 +1,68 @@
+import { Arguments } from 'yargs';
+
+export type GitHubAuth = {
+    user?: string;
+    pass?: string;
+    otp?: string;
+    token?: string;
```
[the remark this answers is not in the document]
antross: Probably better to split this into two separate sub-types as well and union them together in `GitHubAuth`.

release/tasks/calculate-new-versions.ts
molant: This is most likely a typo. I think it should be `pkg.content.version`

release/lib/update-3rd-party.ts
```diff
@@ -0,0 +1,60 @@
...
+        } catch (e) {
+            observer.error(e);
+        }
+    }
+
+    observer.complete();
+};
+
+export const updateThirdPartyResources = () => {
+    return new Observable((observer) => {
```
antross: What does this do?

release/tasks/save-changes.ts
```diff
@@ -0,0 +1,15 @@
+import { debug, updatePkgJson } from '../lib/utils';
+import { Context } from '../@types/custom';
+
+/** Saves in disk the changes to `package.json` of all packages */
+export const saveChanges = (ctx: Context) => {
```
antross: Maybe rename to `savePackageChanges` to be a bit more specific?

release/tasks/validate-changes.ts
```diff
@@ -0,0 +1,46 @@
...
+        maxLength = Math.max(maxLength, name.length);
+    }
+
+    console.log(`Packages to update:`);
+    packages.forEach((pkg) => {
+        const { name: packageName, oldVersion } = pkg;
+        const newVersion = pkg.content.version;
+
+        if (newVersion !== oldVersion) {
+            console.log(`${packageName + new Array(maxLength + 1 - packageName.length).join(' ')} ${oldVersion} -> ${newVersion}`);
```
antross:
```suggestion
            console.log(`${packageName.padEnd(maxLength + 1, ' ')} ${oldVersion} -> ${newVersion}`);
```

release/tasks/run-tests.ts
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Fully tests a package
+ */
+const testPackage = async (pkg: Package) => {
+    // Package could be a configuration and nothing needs to be tested
+    if (!pkg.content.scripts || !pkg.content.scripts.test) {
+        return;
+    }
+
+    // npm run lint && npm run build && npm run test-only
```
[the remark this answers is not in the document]
antross: Why not run `yarn test` which we already configure in each package to run all three steps?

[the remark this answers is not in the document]
molant: That was my intention with line 13 but forgot to update the rest of the code 😅

molant:
> test `eslint-plugin-import` 

It doesn't seem to do what we need. I've looked at some of the resolvers for this package but none do what we need. Creating a resolver doesn't seem overcomplicated so I'm thinking about creating one that will find the first `package.json` in the tree for each file and then use the `dependencies` name to check if it is installed or not.

molant:
@antross, @sarvaje I'm going to run a full run with `--dry-run` to see how it goes and without publishing to GitHub but you should start looking into this when you have the time 🙏

Example of the help menu:
image

Example of the task menu:
image

antross:
Per our discussion, don't forget to re-add the checks to verify we only release from `master` on the official `webhintio/hint` repo.

antross:
@molant The `yarn release` command still points to the old, non-existent release script in `package.json`.

release/lib/git-helpers.ts
```diff
@@ -0,0 +1,310 @@
...
+
+const USER_AGENT = 'Nellie The Narwhal';
+
+const authors: Map<string, Promise<Author | null>> = new Map();
+const commitAuthors: Map<string, Promise<Author | null>> = new Map();
+/**
+ * Returns the GitHub user information associated with a give commit.
+ * @param commitSHA The sha of the commit to get the information from.
+ */
+const getCommitAuthorInfo = (commitSHA: string): Promise<Author | null> => {
```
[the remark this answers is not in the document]
molant: I originally had it with `async/await` but because the info for the commits is run in parallel I was be querying the same information multiple times and reach the GitHub limit for requests and everything crashed. Now that everything is stored as promises we only request the bare minimum.

release/tasks/run-tests.ts
```diff
@@ -0,0 +1,105 @@
...
+            debug(message);
+
+            try {
+                /**
+                 * Build all the depedencies manually because there are some packages that need
+                 * the assets so `tsc -b` is not reliable.
+                 */
+                await buildPackages(pkg.references, packages);
+            } catch (e) {
+                observer.error(`Build failed for package ${pkg.name} dependencies`);
```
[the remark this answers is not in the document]
molant: I think the `observer.error`  stops the execution but now I'm not sure :/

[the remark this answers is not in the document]
molant: Yup, [`error` means completion](http://reactivex.io/documentation/contract.html) from what I read:

> An Observable may make zero or more OnNext notifications, each representing a single emitted item, and it may then follow those emission notifications by either an **OnCompleted or an OnError notification, but not both.**

molant:
Feedback should be addressed and conflicts resolved!

molant:
Rebased with master and fixed the last round of issues. Now if something breaks in the middle we will continue executing tasks and ask for clean up at the end. Also capturing exception with `tasks.run().catch()` and telling the user to look the logs to get more information.

Also tested with mac and everything seems to be working here as well. I can't repro the node version limitation in this platform so maybe check the OS before validating the version? @antross what do you think?

release/lib/utils.ts
```diff
@@ -0,0 +1,181 @@
...
+import { Observable, Subscriber } from 'rxjs';
+
+export const logFilename = `release.log`;
+export const REPOSITORY_SLUG = 'webhintio/hint';
+export const REPOSITORY_URL = `https://github.com/${REPOSITORY_SLUG}`;
+
+/**
+ * The output stream to `release.log`
+ */
+// TODO: We should wait for the 'open' event to be sure everything is fine
```
antross: Was this something you wanted to do before merging?

release/tasks/get-packages.ts
```diff
@@ -0,0 +1,116 @@
...
+    return async (ctx: Context) => {
+        debug(`Loading packages`);
+        ctx.packages = new Map();
+
+        const ignored = ignoredPackages.map((pkg) => {
+            return path.join(process.cwd(), 'packages', pkg);
+        });
+
+        const pkgPaths = await globby([
+            'packages/**/*/package.json',
```
antross: Do we support nested packages generally, or would `'packages/*/package.json'` suffice here?

release/lib/utils.ts
molant: not really, it complicates the code quite a bit and we take long enough to have the stream ready before the first use... I'll just leave the comment here so at least we know I was aware in case something happens.

release/tasks/get-packages.ts
molant: I don't know what the future holds and it was easy enough to have it this way and just negate the `fixtures`

molant:
@antross I've published the refactor we talked about but I'm going to change `request` with `octokit/rest` for the GitHub interaction. The reason is that we are still being throttled from time to time and I'd like to avoid that so this is still not ready to be merged.

molant:
I've migrated to octokit and not getting any more issues requesting info. Code should be simpler now too. @sarvaje want to take another look?  You were the one suggesting to migrate to it initially 😊

release/lib/git-helpers.ts
```diff
@@ -1,78 +1,85 @@
...
+
+const octokitOptions = {
+    log: {
+        debug,
+        error: debug,
+        info: debug,
+        warn: debug
+    },
+    throttle: {
+        onAbuseLimit: (retryAfter: number, options: { method: string; url: string; request: { retryCount: number } }) => {
```
sarvaje: We should have this in the `online-service` :)

release/lib/git-helpers.ts
```diff
@@ -203,6 +210,34 @@ const extractDataFromCommit = (sha: string): Promise<Commit> => {
...
+    const options = Object.assign({}, { auth: `token ${token}` }, octokitOptions);
+
+    const kit = new Client(options);
+
+    return kit;
+};
+
+const createOctokitFromUserPass = (auth: GitHubAuth) => {
+    if ('token' in auth) {
+        return new Client();
```
sarvaje: shouldn't this be `return createOctokitFromToken(auth.token)` ?

molant: That code should never be hit, it was there so TypeScript doesn't complain about possible missing properties and ESLint about not having consistent returns. I'll add it for completeness with a comment.

PR #2069, opened 2019-03-15
New: Add parameter to output formatter output to a file

### LINKED ISSUE #2056: [Feature] Parameter to output formatter output to a file [1.5]
# 🚀 Feature request

## Description

Even though we have a few formatters that output to a directory or a file (excel, html) it will be good to have a parameter or similar so any formatter can output to one. Something like `hint --output results.json`.

Not sure if the CLI shoul try to do something generic, if it should always be done at the formatter level or what. I see pros and cons for each.

sarvaje:
Fix #2056

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/formatter-json/src/formatter.ts
+++ b/packages/formatter-json/src/formatter.ts
@@ -35,7 +35,7 @@ const debug = d(__filename);
 
 export default class JSONFormatter implements IFormatter {
     /** Format the problems grouped by `resource` name and sorted by line and column number */
-    public async format(messages: Problem[], target: string | undefined, options: FormatterOptions = {}) {
+    public async format(messages: Problem[], options: FormatterOptions = {}) {
 
         debug('Formatting results');
 
--- a/packages/formatter-json/src/formatter.ts
+++ b/packages/formatter-json/src/formatter.ts
@@ -10,18 +10,19 @@
  */
 
 import {
-    forEach,
     groupBy,
+    reduce,
     sortBy
 } from 'lodash';
 
 import { debug as d } from 'hint/dist/src/lib/utils/debug';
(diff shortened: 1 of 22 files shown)


packages/hint/src/lib/cli/analyze.ts
```diff
@@ -361,6 +361,7 @@ export default async (actions: CLIOptions): Promise<boolean> => {
         const formatterOptions: FormatterOptions = {
             config: userConfig || undefined,
             date,
+            output: actions.output,
```
antross: How about resolving this to an absolute path against the current working directory here so we don't have to do it in each formatter?

sarvaje: done!

packages/formatter-json/src/formatter.ts
```diff
@@ -34,17 +35,30 @@ const debug = d(__filename);
 
 export default class JSONFormatter implements IFormatter {
     /** Format the problems grouped by `resource` name and sorted by line and column number */
-    public format(messages: Problem[]) {
+    public async format(messages: Problem[], target: string | undefined, options: FormatterOptions = {}) {
```
molant: @antross @sarvaje what do you think about combining `target` and `options`? `target` is not used by all the `formatters` but `options` yes for the filename output so right now is a bit weird

sarvaje: I like the idea

sarvaje: @molant done.

molant:
@sarvaje can you please fix the conflicts?

sarvaje:
> @sarvaje can you please fix the conflicts?

@molant done!

Issue #1499, opened 2018-10-29
[extension-browser] Reduce webhint bundle size [4]


antross:
Part of #48.

Current bundle size after PR #2091 is ~6.3MB raw, ~3.1MB minified, and ~0.6MB minified and compressed.

See exactly how each dependency contributes to this size by downloading the breakdown [stats.zip](https://github.com/webhintio/hint/files/2994119/stats.zip), then view `report.html` or use `stats.json` with [webpack visualizer](https://chrisbateman.github.io/webpack-visualizer/).

Opportunities:
* [x] Minify bundle when building for release
* [x] Include only necessary portions of `highlight.js`
* [x] Eliminate `cheerio` dependency
* [x] Include only necessary portions of `lodash`
* [x] Include only necessary portions of `mdn-browser-compat-data`
* [x] Use `acorn` directly for `parser-javascript` (instead of `eslint`)
* [ ] Eliminate `caniuse` dependency
* [ ] Remove `async` package when bundled
* [ ] Remove `mime-db` when bundled
* [ ] Scope `snyk-snapshot.json` to libraries we can detect
* [ ] Mock `debug`

antross:
Here's a tool that can help visualize what's taking up space in a webpack bundle:
https://chrisbateman.github.io/webpack-visualizer/

An initial look puts the top 4 contributors at:
* lodash (~700K)
* axe-core (~540K)
* caniuse (~480K)
* cheerio (~360K)

Both lodash and caniuse can probably be cut down to just the pieces we use, cheerio may be able to be eliminated entirely, and axe-core is probably going to remain as-is.

molant:
@antross is this done? I see it closed but no related commit nor how much the size has improved.

antross:
Good catch. I did not mean to close this.

molant:
I believe we are bundling moment as well (at least we do for snyk and maybe others). That's also another source of potential savings.

molant:
Another package we could mock is `debug`. We are never going to see the output from the extension.

molant:
Related to #1824

molant:
@jdalton can you please point us to the tools we should be using to reduce lodash's size?

jdalton:
Sure!

There's [babel-plugin-lodash](https://www.npmjs.com/package/babel-plugin-lodash) to transform to cherry-picking and [lodash-webpack-plugin](https://www.npmjs.com/package/lodash-webpack-plugin) to reduce extraneous functionality.

[...]

Issue #2067, opened 2019-03-15
[Bug] `fetch::end::html` and `traverse::*` use different logic to detect HTML


antross:
Found because this broke some tests while trying to update `hint-meta-charset-utf-8` to remove `cheerio`. 

The logic also appears to be inconsistent between connectors, at least between `connector-jsdom` and `utils-debugging-protocol-common`. This should be fixed to use common helpers for both scenarios and across connectors.

Triggering `fetch::end::html` in `connector-jsdom`:
```ts
// Event is also emitted when status code in response is not 200.
await this.server.emitAsync(`fetch::end::${getType(mediaType!)}` as 'fetch::end::*', fetchEnd);
```

Triggering `fetch::end::html` in `utils-debugging-protocol-common`:
```ts
        let suffix = getType(response.mediaType);
        const defaults = ['unknown', 'xml'];

        if (isTarget && defaults.includes(suffix)) {
            suffix = 'html';
        }

        const eventName = `fetch::end::${suffix}` as 'fetch::end::*';
```

Both rely on the helper `getType`, but `utils-debugging-protocol-common` performs post-processing that isn't done anywhere else and treats more content as HTML. In `getType` itself, both `text/html` and `application/xhtml+xml` are accepted as HTML:
```ts
        case 'text/html':
        case 'application/xhtml+xml':
            return 'html';
```

Triggering a traverse in `connector-jsdom` and `utils-debugging-protocol-common` both use a different helper, `isHTMLDocument`, which treats anything from the filesystem as HTML regardless of type, but only accepts `text/html` for remote documents
```ts
/** Convenience function to check if a resource is a HTMLDocument. */
export default (targetURL: string, responseHeaders: HttpHeaders): boolean => {

    // If it's a local file, presume it's a HTML document.

    if (isLocalFile(targetURL)) {
        return true;
    }

    // Otherwise, check.

    const contentTypeHeaderValue: string | undefined = responseHeaders['content-type'];
    let mediaType: string;

    try {
        mediaType = parseContentTypeHeader(contentTypeHeaderValue || '').type;
    } catch (e) {
        return false;
    }

    return mediaType === 'text/html';
};
```

antross:
Per discussion with @molant we should look at what mime-types are most common in the HTTP Archive to help determine what the correct fix is here.

We also need to look at how XML is handled by each of the connectors as we'd likely still want to traverse the DOM in that case so long as it gets parsed correctly.

antross:
[...]

PR #2070, opened 2019-03-15
Remove cheerio dependency

antross:
## Pull request checklist

Make sure you:

Ref #1499 

Part of the work to reduce bundle size for the browser extension.
Required fixing an issue with recognized types for JavaScript. 

~Also enabled a new test in `hint-meta-charset-utf-8` for reporting
unnecessary XML declarations (previously a TODO).~
(abandoned due to mime-type detection issues)

### DIFF
--- a/packages/hint-meta-charset-utf-8/package.json
+++ b/packages/hint-meta-charset-utf-8/package.json
@@ -6,13 +6,10 @@
     ],
     "timeout": "1m"
   },
-  "dependencies": {
-    "cheerio": "^1.0.0-rc.2"
-  },
   "description": "hint for best practices related to the meta charset",
   "devDependencies": {
+    "@hint/parser-html": "^2.0.2",
     "@hint/utils-tests-helpers": "^3.0.0",
-    "@types/cheerio": "^0.22.11",
     "ava": "^1.3.1",
     "cpx": "^1.5.0",
     "eslint": "^5.15.1",
--- a/packages/hint-meta-charset-utf-8/src/hint.ts
+++ b/packages/hint-meta-charset-utf-8/src/hint.ts
@@ -9,11 +9,11 @@
  * ------------------------------------------------------------------------------
  */
 
-import * as cheerio from 'cheerio';
-
-import { HTMLDocument, HTMLElement, IHint, FetchEnd, TraverseEnd } from 'hint/dist/src/lib/types';
+import { IHint } from 'hint/dist/src/lib/types';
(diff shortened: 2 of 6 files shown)


packages/hint-meta-charset-utf-8/package.json
```diff
@@ -6,13 +6,10 @@
     ],
     "timeout": "1m"
   },
-  "dependencies": {
-    "cheerio": "^1.0.0-rc.2"
-  },
   "description": "hint for best practices related to the meta charset",
   "devDependencies": {
+    "@hint/parser-html": "^2.0.2",
```
molant: Should we just add this dependency to `utils-tests-helpers` and add the `html` parser directly to the configurations?

molant:
>   ✖ [jsdom] The XML charset declaration is used' No results found, should be 1

👆 this test is failing

antross:
> > ✖ [jsdom] The XML charset declaration is used' No results found, should be 1
> 
> 👆 this test is failing

It's because the test doesn't specify a mime-type and it gets inferred as `text/xml`. This ties back to the mime-type inconsistencies we were talking about between connectors earlier today. 

I'm going to fix this by explicitly specifying `text/html` as the mime-type in the test.

molant:
> I'm going to fix this by explicitly specifying `text/html` as the mime-type in the test.

How much sense does it make taking into consideration that the test is `The XML charset declaration is used`? Maybe we should just skip it for now?

packages/hint-meta-charset-utf-8/package.json
antross: In my other PR we added `html` directly to any configuration passed to `utils-tests-helpers`.

That said, I'll still need `parser-html` as a `devDependency` here in order to get the correct types when listening for the `parse::end::html` event. I just don't have to explicitly specify it in the tests.

molant: Got it.

antross:
> How much sense does it make taking into consideration that the test is `The XML charset declaration is used`? Maybe we should just skip it for now?

I think it makes sense since the problem is when someone specifies the XML declaration in HTML content. If it were an actual XML document the declaration would be perfectly fine.

antross:
> I think it makes sense since the problem is when someone specifies the XML declaration in HTML content. If it were an actual XML document the declaration would be perfectly fine.

Until I discovered an XML declaration causes us to treat content as XML internally regardless of mime-type (this differs from browser behavior). I'll add a note to #2067 and drop this for now.

PR #2091, opened 2019-03-21
Reduce browser extension bundle size

antross:
## Pull request checklist

Make sure you:

Ref #1499 

Includes a few more changes aimed at reducing the size of the browser extension bundle:
* [x] Cherry-pick just the parts of `lodash` we need
* [x] Drop `eslint` in favor of using `acorn` directly in `parser-javascript`
* [x] Filter `mdn-browser-compat-data` to include only data we need
* [x] Turn on minification when building browser extension for release

Previously the size of the `bundle` folder after a build was ~7.5MB.
With these changes it is now ~3MB when built for release (~600KB when zipped).


molant:
@antross you have some linting issues.

packages/connector-chrome/tests/events.ts
```diff
@@ -9,7 +9,8 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { URL } from 'url';
 
-import { groupBy, every } from 'lodash';
+import groupBy = require('lodash/groupBy');
```
molant: Did you take a look at [`lodash-es`](https://www.npmjs.com/package/lodash-es)?

packages/parser-javascript/package.json
```diff
@@ -7,12 +7,13 @@
...
-    "eslint": "^5.15.1",
-    "espree": "^5.0.1"
+    "acorn": "^6.1.1",
+    "acorn-walk": "^6.1.1"
   },
   "description": "webhint parser needed to analyze JavaScript files",
   "devDependencies": {
-    "@types/eslint": "^4.16.6",
+    "@types/acorn": "^4.0.5",
+    "@types/estree": "^0.0.39",
```
molant: I think you are missing `estree` dependency. You have the types and you are using it later in the code but I don't see it in the `package.json`. We are probably loading the `eslint` packed one.

packages/connector-chrome/tests/events.ts
antross: Yes, but it won't work for us until we change our TypeScript build to emit ecmascript modules instead of commonjs (which requires all our dependencies to support ecmascript modules as well). 

Thus the cherry-picked approach is our only option at the moment.

packages/parser-javascript/package.json
antross: Nope, just need the types as `acorn` emits an `estree` compatible AST.

molant: That's going to be a problem once I start working with the package resolver, need to think how to solve it.

antross: Also worth noting that there isn't an `estree` package as far as I'm aware - just `@types/estree`.

packages/extension-browser/scripts/pack-mdn-data.js
```diff
@@ -1,13 +1,245 @@
...
+        delete data.__compat;
+
+        return;
+    }
+
+    // Description is not needed for analysis.
+    delete compat.description;
+
+    // Remove unnecessary data per-browser.
+    for (const browserName in support) {
```
sarvaje: Why don't you use `Object.entries(support)` or `Object.keys(support)` so you don't need to use a `for .. in` to iterate?

packages/extension-browser/scripts/pack-mdn-data.js
```diff
@@ -1,13 +1,245 @@
...
+};
+
+/**
+ * Remove all features and data which are unnecessary for analysis.
+ * E.g. feature is universally supported or universally unsupported.
+ *
+ * @param {bcd.Identifier} data
+ */
+const removeFeatures = (data) => {
+    for (const key in data) {
```
sarvaje: Same as before

packages/extension-browser/scripts/pack-mdn-data.js
```diff
@@ -1,13 +1,245 @@
...
+    }
+};
+
+/**
+ * Strip browser details down to name only.
+ *
+ * @param {bcd.Browsers} browsers
+ */
+const removeBrowserDetails = (browsers) => {
+    for (const browserName in browsers) {
```
sarvaje: Same as before

packages/extension-browser/scripts/pack-mdn-data.js
antross: No particular reason, just habit 😊 I'll switch over and update the PR.

antross: Done!

PR #2093, opened 2019-03-22
Chore: Ignore `chrome` on CI

### LINKED ISSUE #2031: [Bug] Compat-API tests fail fairly often on Windows and Azure Pipelines [0.2]
I believe it's because there are so many tests with different configurations. This causes:

- Fully new instances of Chrome to be created
- Lots of new tests servers

From what @sarvaje told me his machine runs like a 🐌 and is on 🔥 when running these tests (and he has a good one...)

The solutions I can think of are:

- Split the tests in multiple files so ava doesn't try to execute all simultaneously
- Add the option to have a single server when testing on CI (I though I opened an issue for this but can't find it right now)

@webhintio/core thoughts?

molant:
## Pull request checklist

Make sure you:

Fix #2031

### DIFF
--- a/packages/hint-compat-api/tests/css-next.ts
+++ b/packages/hint-compat-api/tests/css-next.ts
@@ -4,6 +4,8 @@ import readFile from 'hint/dist/src/lib/utils/fs/read-file';
 import { HintTest } from '@hint/utils-tests-helpers/dist/src/hint-test-type';
 import * as hintRunner from '@hint/utils-tests-helpers/dist/src/hint-runner';
 
+import { ignoredConnectors } from './_ignored-connectors';
+
 const hintPath = getHintPath(__filename, true);
 
 const generateCSSConfig = (fileName: string) => {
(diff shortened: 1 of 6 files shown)


packages/hint-compat-api/tests/css-next.ts
```diff
@@ -4,6 +4,8 @@ import readFile from 'hint/dist/src/lib/utils/fs/read-file';
 import { HintTest } from '@hint/utils-tests-helpers/dist/src/hint-test-type';
 import * as hintRunner from '@hint/utils-tests-helpers/dist/src/hint-runner';
 
+import { ignoredConnectors } from './fixtures/ignored-connectors';
```
sarvaje: Is `fixtures` the best place to have this script?

molant: I tried with utils and a a complained about not having tests in that folder. I'll take a look see if there's another convention to use.

antross: From https://github.com/avajs/ava/blob/f4b2d196e813cf3549e19703e5a79ebc988fb8ca/docs/05-command-line.md

> Directories are recursed, with all *.js files being treated as test files. Directories named `fixtures`, `helpers` and `node_modules` are always ignored. So are files starting with `_` which allows you to place helpers in the same directory as your test files.

molant:
> I think your approach is the most practical solution to our immediate problem of flaky tests.

I actually started separating tests by files but it wasn't really practical 😞

packages/hint-compat-api/tests/css-next.ts
molant: I've renamed the file to `_ignored-connectors.ts` and moved it with the tests.

PR #2138, opened 2019-03-29
Chore: Add common VS Code settings

### LINKED ISSUE #1887: [Feature] VS Code launch config file [0.1]
# 🚀 Feature request

## Description

It will be awesome if we could add the VSCode launch config file to the project. There are a few configuration the team uses to run scripts, tests for a particular hint, etc. 

This will allow getting started on a new clone or fork pretty easily. Also it will normalize how we test accross the team and make sure we are use the best possible configuration. The main caveat is that we need to make sure not to update it when doing a PR unless we are adding something new.

molant:
## Pull request checklist

Make sure you:

Track the `.vscode` folder so the team can share the settings:

* Recommended extensions
* PreLaunchTask to create missing `node_modules/.cache/ava/`
* Debugging tasks
* Enable `ESLint` with TypeScript in the workspace as well as
  other formatting options like empty ending line, auto-trim,
  etc.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Fix #1887

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@
 .DS_Store
 .eslintcache
 .nyc_output
-.vscode
 cdp.lock
 coverage
 dist
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,11 @@
 .DS_Store
 .eslintcache
 .nyc_output
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
 cdp.lock
 coverage
 dist
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+    "recommendations": [
+        "dbaeumer.vscode-eslint",
+        "github.vscode-pull-request-github",
+        "ms-vsliveshare.vsliveshare",
+        "webhint.vscode-webhint"
+    ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -41,7 +41,7 @@
             "name": "webhint",
             "program": "${workspaceFolder}/packages/hint/dist/src/bin/hint.js",
             "args": [
-                "https://example.com"
+                "${selectedText}"
             ],
             "cwd": "${workspaceFolder}/packages/hint/",
             "sourceMaps": true,
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,57 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            // Use IntelliSense to learn about possible attributes.
+            // Hover to view descriptions of existing attributes.
+            // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+            "type": "node",
+            "request": "launch",
+            "name": "Run AVA test",
+            "preLaunchTask": "npm: cache",
+            "program": "${workspaceFolder}/node_modules/ava/profile.js",
+            "args": [
+                "${fileDirname}/../dist/tests/${fileBasenameNoExtension}.js"
+            ],
+            "sourceMaps": true,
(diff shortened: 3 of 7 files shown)


molant:
Please let me know if there are any tasks in your `launch.json` that we should add to this.

.vscode/tasks.json
```diff
@@ -0,0 +1,19 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
```
sarvaje: This is nice!

Maybe we can see if there is any task we can have to compile the modified files before run the tests.

.gitignore
```diff
@@ -7,7 +7,6 @@
 .DS_Store
 .eslintcache
 .nyc_output
-.vscode
```
sarvaje: I think .vscode should still be part .gitignore

molant:
> For example, imagine that someone doesn't like the option "editor.renderWhitespace": "all",

I'm ok removing that one. The user can still have a global configuration for vs code in general. The other options are validated via eslint so this saves time and possible linting issues.

I'm not sure having a task to create them is the best option, if we add something new we will have to regenerate those files. Also the current tasks are generic enough that don't have any hardcoded values (current AVA test, curren js test). The only one is the url for `webhint` but it's not like we use that task that often.

.gitignore
antross: How about [this recommendation](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) from GitHub?
```
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
```

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+    "recommendations": [
+        "dbaeumer.vscode-eslint",
+        "github.vscode-pull-request-github",
+        "ms-vsliveshare.vsliveshare",
+        "webhint.vscode-webhint"
```
antross: ❤

.gitignore
molant: Are there any other files that go there? Can `settings` be composed?

.vscode/launch.json
```diff
@@ -0,0 +1,57 @@
...
+            ],
+            "console": "integratedTerminal",
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "webhint",
+            "program": "${workspaceFolder}/packages/hint/dist/src/bin/hint.js",
+            "args": [
+                "https://example.com"
```
antross: A bit of a hack, but would allow this to be specified without having to edit the `launch.json`. We could even keep a common targets file handy to select URLs from for testing.
```suggestion
                "${selectedText}"
```

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "editor.renderWhitespace": "all",
```
antross: I agree this one is more of a personal preference (though I also use `all`)
```suggestion
```

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
...
+    "eslint.alwaysShowStatus": true,
+    "eslint.validate": [
+        "javascript",
+        "javascriptreact",
+        "typescript",
+        "typescriptreact"
+    ],
+    "files.insertFinalNewline": true,
```
antross: ❤

.gitignore
antross: Per https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control some seem to suggest extensions may occasionally put items in this folder so it's probably best to only allow the four files we need.

As for composing `settings` I didn't find anything suggesting this is currently possible.

.gitignore
```diff
@@ -7,7 +7,6 @@
 .DS_Store
 .eslintcache
 .nyc_output
```
molant:
```suggestion
.nyc_output
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
```

molant:
@antross I've applied the suggestions. Let me know if there's any more feedback!

PR #2140, opened 2019-03-29
Chore: Migrate browser extension to React

### LINKED ISSUE #2104: Use react in the browser extension [6]
Migrate to using (pre)react

antross:
## Pull request checklist

Make sure you:

Fix #2104

Test bundle (built for release):
* [extension-browser-react.zip](https://github.com/webhintio/hint/files/3030694/extension-browser-react.zip) (Updated 04/01/2018)

Feel free to take a look, but this is probably worth walking through together.


molant:
@antross looks like there's a problem on Linux with `Analyze.css`. I've checked parts of the output and that file seems to generate a `.d.ts` different than the rest:

```
Wrote /home/vsts/work/1/s/packages/extension-browser/src/devtools/panel/views/Page.css.d.ts
Wrote /home/vsts/work/1/s/packages/extension-browser/src/devtools/panel/views/pages/analyze.css.d.ts
Wrote /home/vsts/work/1/s/packages/extension-browser/src/devtools/panel/views/pages/Config.css.d.ts
```

It's the only one that is all lower case.

Is there any reason why these files are capitalized? In the rest of the project all is lower cased.

packages/extension-browser/src/devtools/panel/utils/inspire.ts
```diff
@@ -22,4 +25,19 @@ const inspire = (): string => {
...
+            setMessage(inspire());
+        }, 7500);
+
+        return () => {
+            clearInterval(interval);
+        };
+    });
+
+    return message;
+};
```
molant: Doesn't eslint complain about not ending empty line? I thought we had a rule for that 🤔

packages/extension-browser/src/devtools/panel/views/Header.tsx
```diff
@@ -0,0 +1,35 @@
...
+type Props = {
+    actionDisabled?: boolean;
+    actionName: string;
+};
+
+const Header = ({ actionDisabled, actionName }: Props) => {
+    return (
+        <header className={styles.root}>
+            <div className={styles.actions}>
+                <Button type="submit" className={styles.button} primary={true} disabled={actionDisabled}>
```
molant: So I guess we have to use `Button` because that's how it's defined in `packages/extension-browser/src/devtools/panel/views/Button.tsx`? I'm not a fan of relying in capitalization for differences. Is this a common practice in react?

packages/extension-browser/src/devtools/panel/views/pages/config/TargetBrowsers.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+/** Check if a user's custom `browserslist` query is valid, notifying them if it is not. */
+const validate = (value?: string): string => {
+    try {
+        browserslist(value);
+
+        return '';
+    } catch (e) {
+        // Report errors, stripping messages about "old" browserslist since the user won't have control over that.
+        return e.message.replace(' Maybe you are using old Browserslist or made typo in query.', '');
```
molant: Maybe we should replace this error saying something like:

> Maybe you made a typo in your query or are using a syntax not supported by browserslist vXXX

Having an empty error message is not very user friendly.

packages/extension-browser/src/devtools/panel/views/Header.tsx
antross: We can call it whatever we want so long as it starts with a capital letter (which is what React requires). I'm fine renaming this to something like `AppButton` if we want to have a clearer visual difference compared to an HTML `button`.

packages/extension-browser/src/devtools/panel/views/pages/config/TargetBrowsers.tsx
antross: This doesn't create an empty error message, it just strips the suffix.

molant: Great. Can you put a comment with an examination le of the message that will be modified? With the current one I was thinking the whole message was going to get removed.

packages/extension-browser/src/devtools/panel/utils/inspire.ts
antross: Looks like there already is an empty line at the end of this file...

packages/extension-browser/src/devtools/panel/views/Header.tsx
antross: Also as far as the filename is concerned we can absolutely use lowercase or kebab-case if we prefer to align with the rest of the webhint codebase. Most React projects follow the convention of PascalCase filenames, but it is not required.

antross: I went ahead and changed everything to kebab-case to align better with the rest of the `hint` project.

antross:
FYI this PR currently also has my fix for #2148 so self-hosting the extension doesn't break on most sites. I'll drop that after rebasing once it's merged.

antross:
Some remaining tasks:
* [ ] Pass disabled categories all the way to the content script
* [ ] Move inspire messages to localized strings
* [ ] Clear validation state when selecting defaults
* [ ] Trim target browsers input value...

antross:
FYI this should be ready for another round of review. I've left separate commits for recent updates to make it easier to see the changes, but I'll squash it all into one commit before we merge.

packages/extension-browser/scripts/bundle-size.js
```diff
@@ -0,0 +1,31 @@
...
+    fs.stat(filename, (err, stats) => {
+        if (err) {
+            throw new Error(`Reading bundle failed: ${err}`);
+        }
+
+        const sizeKB = Math.round(stats.size / 1000);
+        const limitKB = Math.round(packageJSON.bundleSize / 1000);
+
+        if (sizeKB > limitKB) {
+            throw new Error(`Bundle size of ${sizeKB}KB exceeds limit of ${limitKB}KB defined in package.json for ${filename}.`);
```
molant: What about `console.error` and `process.exitCode = 1`?
Otherwise we will get the callstack and such and I think this script is only executed via the CLI.

packages/extension-browser/src/devtools/views/app.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    const onResults = (results: ResultsData, duration: number) => {
+        setPage(Page.Results);
+        setResults(results);
+        trackFinish(duration);
+    };
+
+    const onStart = (config: ConfigData) => {
+        setPage(Page.Analyze);
+        sendMessage({ enable: { config } }); // Tell the background script to start scanning.
```
molant: Maybe we should document the messages that can be sent somewhere?

packages/extension-browser/src/devtools/views/pages/config/resources.tsx
```diff
@@ -0,0 +1,116 @@
...
+/** Create a regular expression to exclude URLs not part of the current origin. */
+const buildIgnoreThirdParty = (): Promise<string> => {
+    return new Promise((resolve) => {
+        browser.devtools.inspectedWindow.eval('location.origin', (origin: string) => {
+            resolve(`^(?!${escapeRegExp(origin)})`);
+        });
+    });
+};
+
+/** Check if a user's custom `browserslist` query is valid, notifying them if it is not. */
```
molant: This comment talks about browserslist but we are verifying the regex for ignored resources, right?

packages/extension-browser/package.json
```diff
@@ -6,6 +6,7 @@
     ],
     "timeout": "1m"
   },
+  "bundleSize": 700000,
```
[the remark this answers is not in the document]
antross: I picked it as it was the next higher round number above our current size (~651000 bytes). It's in the package.json so we can easily change it if we deliberately want to increase the limit, but hopefully should help us catch accidentally bundling something large in that we didn't mean to.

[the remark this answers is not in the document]
molant: I'm fine with any number, it was just a comment 😊
Hopefully we can reduce it a bit more soon-ish

packages/extension-browser/scripts/bundle-size.js
antross: Personally I prefer unhandled exceptions as they prevent subsequent script from running. I've been using them other scripts and it doesn't appear to spew a bunch of extra data to the console. 

Also per https://nodejs.org/api/process.html#process_process_exit_code:
> If it is necessary to terminate the Node.js process due to an error condition, throwing an uncaught error and allowing the process to terminate accordingly is safer than calling process.exit().

That's not recommending against `process.exitCode`, but it is calling out throwing an unhandled error as an acceptable way to exit.

packages/extension-browser/src/devtools/views/app.tsx
antross: All the possible messages are defined in `src/shared/types.ts`, but I'm happy to add additional comments there explaining what each is used for. 

That said, I've been debating refactoring this a bit to have a single utils file with methods to call for each action that sends messages (and only have that file actually call `sendMessage`).

packages/extension-browser/scripts/bundle-size.js
molant: We should probably normalize what we do across all the scripts. I believe I've been using `exitCode` in more than one place.

packages/extension-browser/src/devtools/views/pages/config/resources.tsx
antross: Yep. 🤦‍♂️

packages/extension-browser/src/devtools/views/app.tsx
molant: > That said, I've been debating refactoring this a bit to have a single utils file with methods to call for each action that sends messages (and only have that file actually call `sendMessage`).

Open an issue for the future and call it a day?

packages/extension-browser/package.json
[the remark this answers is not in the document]
antross: Gotcha. Couldn't help myself sharing more context. 😉

packages/extension-browser/src/devtools/views/pages/config/config-section.tsx
```diff
@@ -0,0 +1,24 @@
...
+    title: string;
+}
+
+/**
+ * Wrap related configuration items with a common container and styles.
+ */
+const ConfigSection = ({ children, title }: Props) => {
+    return (
+        <section className={styles.root}>
+            <h1 className={styles.header}>
```
sarvaje: We are going to have more than one `h1`, are we ok with that?

packages/extension-browser/src/@types/react-syntax-highlighter.ts
```diff
@@ -0,0 +1,19 @@
+declare module 'react-syntax-highlighter/dist/esm/light' {
```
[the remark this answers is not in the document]
antross: Yes, good catch!

packages/extension-browser/src/devtools/views/pages/config/config-section.tsx
antross: Yep, multiple `h1`s were introduced in HTML5 so long as there's only one `h1` per "sectioning" element. 

Nesting is allowed too so a nested `section` could have it's own `h1` and that's also fine. Browsers will style `h1`'s in nested `section`s as `h2`, `h3`, etc. according to the nesting level.

sarvaje: ok :)

packages/extension-browser/tests/devtools.ts
```diff
@@ -0,0 +1,137 @@
...
+        };
+    });
+};
+
+test('It builds a configuration, starts a scan, and displays results', async (t) => {
+
+    // Create a temporary server to load the panel content into a browser.
+    const server = await Server.create({
+        configuration: {
+            '/': await readFileAsync(path.resolve(__dirname, '../bundle/devtools/panel.html')),
```
molant: Nit: Execute both `readFileAsync` simulatenously with `Promise.resolve`:

```ts
const [root, js] = await Promise.all([
    readFileAsync(path.resolve(__dirname, '../bundle/devtools/panel.html'),
    readFileAsync(path.resolve(__dirname, '../bundle/devtools/panel.js')
]);
```

packages/extension-browser/tests/end-to-end.ts
```diff
@@ -10,6 +10,22 @@ import { readFixture } from './helpers/fixtures';
 
 const pathToExtension = `${__dirname}/../bundle`;
 
+const wait = (delay = 500): Promise<void> => {
```
molant: Nit: We already have a `delay` in `utils/misc`.

Also you never call this with any parameter so might as well not have `delay` as one 😊

packages/extension-browser/tests/end-to-end.ts
```diff
@@ -42,6 +58,43 @@ const findBackgroundScriptPage = async (browser: Browser): Promise<Page> => {
...
+    const devtoolsTarget = targets.filter((t) => {
+        return t.type() === 'other' && t.url().startsWith('chrome-devtools://');
+    })[0];
+
+    const devtoolsPage = await getPageFromTarget(devtoolsTarget);
+
+    await wait();
+
+    // Based on https://github.com/GoogleChrome/puppeteer/issues/3699#issuecomment-450526587
+    await devtoolsPage.keyboard.down('Control');
```
molant: Maybe add a comment for future generations explaining that this navigates to the left the selected panel on the devtools and that if we were to load multiple dev tools extensions it won't work? I know is more or less explained in the comment but I think it's better to have it in here just in case.

packages/extension-browser/tests/end-to-end.ts
antross: I'll drop the param for now. Probably will wait on cutting over to the util until after David's PR is merged.

molant: Works for me!

PR #2166, opened 2019-04-02
Docs: RFC for connector chromium

molant:
This is the RFC to create a new `connector-chromium`. It is not intended to get merged but rather to provide feedback. Once everyone is fine with the workding I'll create a new Epic for it and link issues related to it.

Happy reading!

### DIFF
--- a/rfcs/chromium-connector.md
+++ b/rfcs/chromium-connector.md
@@ -43,7 +43,7 @@ With this in mind, the options would be:
 
 | Entry Point | Values  |
 | ----------- | ------- |
-| ENV         | `WEBHINT_USER` <br> `WEBHINT_PASSWORD` <br> `WEBHINT_USER_SELECTOR` <br> `WEBHINT_PASSWORD_SELECTOR` <br> `WEBHINT_SUBMIT` |
+| ENV         | `WEBHINT_AUTH_USER` <br> `WEBHINT_AUTH_PASSWORD` <br> `WEBHINT_AUTH_USER_SELECTOR` <br> `WEBHINT_AUTH_PASSWORD_SELECTOR` <br> `WEBHINT_AUTH_SUBMIT` |
 | CLI         | `user`, `-u` <br> `password`, `-p` <br> `user-selector`, `-us` <br> `password-selector`, `-ps` <br> `submit`, `-s`
 | Configuration | `user`, `password` and `submit` properties <br> in `options` with `selector` and `value`. <br> See below for example |
 
@@ -54,17 +54,19 @@ With this in mind, the options would be:
   "connector": {
     "name": "chromium",
     "options": {
-       "user": {
-         "selector": "",
-         "value": ""
-       },
-       "password": {
-         "selector": "",
-         "value": ""
-       },
-       "submit": {
-         "selector": ""
-       }
+      "auth": {
+        "user": {
+          "selector": "",
+          "value": ""
+        },
+        "password": {
+          "selector": "",
+          "value": ""
+        },
+        "submit": {
+          "selector": ""
+        }
+      }
     }
   }
   // ...
@@ -78,15 +80,17 @@ An example of a combination of these methods is:
   "connector": {
     "name": "chromium",
     "options": {
-       "user": {
-         "selector": "#user"
-       },
-       "password": {
-         "selector": "#pass"
-       },
-       "submit": {
-         "selector": "#login"
-       }
+      "auth": {
+        "user": {
+          "selector": "#user"
+        },
+        "password": {
+          "selector": "#pass"
+        },
+        "submit": {
+          "selector": "#login"
+        }
+      }
     }
   }
   // ...
--- a/rfcs/chromium-connector.md
+++ b/rfcs/chromium-connector.md
@@ -1,0 +1,213 @@
+# 🚀 Feature request: `connector-chromium` + custom actions
+
+## TL;DR;
+
+Even though we have a `chrome connector` it should probably be renamed to `chromium` as it works with any `chromium` browser that exposes the `chrome DevTools Protocol`. Even `chrome-launcher` supports the environment variable `CHROME_PATH` to indicate the executable of the browser to run.
+
+This alone is not a good reason, so this enoughWhile renaming for the sake of renaming is probably not a good idea, I think we should take this opportunity to modernize the code and create a new connector that will use [`Puppeteer`](https://github.com/GoogleChrome/puppeteer) which was not available when we first started the project.
+
+Using `Puppeteer` will unlock requests like [custom actions](#421) more easily and with ergonomics that developers already know.
+
+## Details
+
+This section contains a list of all the changes and/or new features `connector-chromium`, the CLI, and the Node.js API should implement with some details on how to achieve them.
+
+### User actions
+
+This has been a long standing issue (#421, #1946). The main scenario to unlock is user authentication via the CLI, without forgetting users that need more control of the page before analyzing it. An example of the later will be to click in different elements before triggering the analysis.
+
+The Node.js API should just expose a general way to perform custom actions and the CLI should take advantage of it to allow user authentication.
+
+#### User authentication in the CLI
+
+The authentication steps should be handled by the CLI using the actions ergonomics provided by the Node.js API.
+
+A typical login experience has a user and password `input`s and a submit `button` (everything might or not be inside a `form`). The user fills the inputs, pressed the button, and then the website navigates to another page or updates the URL with a hash. Puppeteer considers a navigation any change in the url, even if it is just a hash so this should cover the most common scenarios.
+
+The information required to do this programatically is:
+
+* **user value** and **`input` selector**
+* **password value** and **`input` selector**
+* **button selector** so it can be `click`ed
+
+In total 5 pieces of information.
+
+The most common ways to provide variables to a CLI are (from lower to higher priority):
+
+1. Environment variables
+1. Configuration file (`.hintrc`)
+1. CLI parameters
+1. A mix of the above: the user provides the selectors via `.hintrc` but uses environment variables for the credentials
+
+With this in mind, the options would be:
+
+| Entry Point | Values  |
+| ----------- | ------- |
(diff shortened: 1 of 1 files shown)


rfcs/chromium-connector.md
```diff
@@ -0,0 +1,213 @@
...
+And then have the environment variable `WEBHINT_PASSWORD` set.
+
+**Limitations**
+
+The new API allows a user to analyze multiple websites simulatenously. This feature will be exposed also in the CLI. With this approach there are two problems:
+
+1. There is no way to provide different credentials to each URL
+1. There is no way to perform the authentication for only certain URLs
+
+These are edge cases and people in needs of this should use the Node.js API or re-architect their scripts to call `hint` multiple times with the right parameters in each case.
```
antross: Let's call out this should throw an error if someone tries to use the API with both authentication and multiple targets (unless there's some other handling we think is appropriate).

rfcs/chromium-connector.md
```diff
@@ -0,0 +1,213 @@
...
+1. Execute `action`
+1. Run the analysis on the new navigation
+
+If `actions` does not exists in `options`, then it will run things as usual.
+
+### `waitFor`
+
+Right now connectors use the `waitFor` option to decide for how long (in ms) they should wait until starting the traversal and such. `Puppeteer` has [several options](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-pagegotourl-options): `timeout`, `load`, `domcontentloaded`, `networkidle0`, `networkidle2`. The same values should be accepted (although maybe not the array option).
+
+**NOTE:** Because not all connectors will support all the values I'm not sure how the best implementation for this would be. Maybe each `connector` has its own configuration type? Something like:
```
antross: I agree this should be part of connector-specific configuration type.

rfcs/chromium-connector.md
```diff
@@ -0,0 +1,213 @@
...
+| Response notification | [`page.on('response')`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-event-response) |
+| Close browser         | [`browser.close()`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-browserclose) |
+| Navigate              | [`browser.newPage()`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-browsernewpage) <br> [`page.goto(url[, options])`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-pagegotourl-options) <br> [`page.waitForNavigation([options])`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-pagewaitfornavigationoptions) |
+| Set page content | [`page.setContent(html[, options])`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-pagesetcontenthtml-options) |
+| Input | [`page.type(selector, text[, options])`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-pagetypeselector-text-options) <br> [`page.click(selector[, options])`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-pageclickselector-options) |
+
+
+## Other comments/questions
+
+**How to run multiple pages in the same browser from different Puppeteer instances?**
```
antross: Any examples of when we would want/need to do this?

rfcs/chromium-connector.md
```diff
@@ -0,0 +1,213 @@
...
+
+The idea is to use a similar approach where the `PID` is stored in a file and alternate between [`puppeteer.launch([options])`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-puppeteerlaunchoptions) and [`puppeteer.connect(options)`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-puppeteerconnectoptions) when the browser is already running (using `browserWSEndpoint`).
+
+**How to run a different Chromium based browser?**
+
+[`puppeteer.launch([options])`](https://pptr.dev/#?product=Puppeteer&version=v1.14.0&show=api-puppeteerlaunchoptions) has an option to pass `executablePath`. An option should be exposed in `connector-chromium` configuration to specify it. This value could be the `path` or maybe a well-known `string` value (`chromium-canary`, `chrome`, etc.) and someone (CLI, `connetor`?) could resolve the path similar to what `chrome-launcher` does today to decide what browser to run.
```
antross: Assuming we allow for per-connector configuration types, should `connector-chromium` allow passing forward any options supported by `puppeteer`?

rfcs/chromium-connector.md
```diff
@@ -0,0 +1,213 @@
...
+Another thing to take into account is if the user starts a session with one `chromium` browser and decides to launch another session with another while it is still running. The `cdp.pid` (or whatever new file will be used) should probably contain information about the executable path to use or maybe this is an scenario that is not supported.
+
+**Package updates**
+
+* `chrome-connector` and `utils-debugging-protocol-common` should be deprecated and point users to download the new one.
+* `configuration-*` packages should be updated to use `connector-chromium`
+
+**Viewport**
+
+Puppeteer's viewport default to 800x600 viewport. `null` disables the default viewport. This should be exposed in the options somehow and I think we should default to `null` in our case.
```
antross: While I agree I'd personally more commonly want `null`, perhaps we should stick closely to `puppeteer`'s defaults, making it easier to just forward an entire group of options along without further processing by us. Thoughts?

rfcs/chromium-connector.md
```diff
@@ -0,0 +1,213 @@
...
+1. Environment variables
+1. Configuration file (`.hintrc`)
+1. CLI parameters
+1. A mix of the above: the user provides the selectors via `.hintrc` but uses environment variables for the credentials
+
+With this in mind, the options would be:
+
+| Entry Point | Values  |
+| ----------- | ------- |
+| ENV         | `WEBHINT_USER` <br> `WEBHINT_PASSWORD` <br> `WEBHINT_USER_SELECTOR` <br> `WEBHINT_PASSWORD_SELECTOR` <br> `WEBHINT_SUBMIT` |
```
sarvaje: I think all these items should include `AUTH`, `WEBHINT_AUTH_USER` `WEBHINT_AUTH_PASSWORD` and so on.

rfcs/chromium-connector.md
```diff
@@ -0,0 +1,213 @@
...
+| Configuration | `user`, `password` and `submit` properties <br> in `options` with `selector` and `value`. <br> See below for example |
+
+`.hintrc` example:
+
+```json
+{
+  "connector": {
+    "name": "chromium",
+    "options": {
+       "user": {
```
sarvaje: what about an `auth` property with all these information?

rfcs/chromium-connector.md
molant: If the user wants to analyze a bunch of pages in the same site behind authentication I think the CLI should support it. In fact, the Node.js API doesn't have any awareness of what `actions` is going to do so I don't see how we will be throwing an error in this case :/

rfcs/chromium-connector.md
molant: When we are running the tests basically 😊

rfcs/chromium-connector.md
molant: Yes, we already do that with the `chrome-connector` and `chrome-launcher`. It's just that finding the path to the executable might be problematic and we can probably improve the ux and have some well-known folders to look into so they can just decide.

rfcs/chromium-connector.md
molant: Not sure about sticking with puppeteer in this one, I think my watch has a higher resolution than that and 800x600... Don't know how representative that resolution is right now. Because we are running on the users machine I think we should use the whole real state and for the online scanner look into the most popular screen resolution and use that one.

rfcs/chromium-connector.md
molant: I like it, should be a bit cleaner even if a bit more nested 😊

rfcs/chromium-connector.md
molant: I don't understand your comment 🤔
This is how the table looks like:

image

sarvaje: What I mean is that I would include the word `AUTH` for the ENV variables:
so the env variables would be: WEBHINT_AUTH_USER, WEBHINT_AUTH_PASSWORD, WEBHINT_AUTH_USER_SELECTOR, WEBHINT_AUTH_PASSWORD_SELECTOR, WEBHINT_AUTH_SUBMIT

molant: OK, only for the `env` variables or do you want me to change the parameters too?

rfcs/chromium-connector.md
molant: @webhintio/contributors thoughts about this? ☝

rfcs/chromium-connector.md
molant: Synced with @sarvaje and only `env`

molant:
@antross can you take a look at my comments?
@sarvaje, I'll update the RFC with your feedback first thing tomorrow morning.

Let me know if there's any other change you want me to do. Otherwise I'll close this issue without merging and paste the markdown into a new Epic. Sounds good?

rfcs/chromium-connector.md
antross: FWIW `puppeteer` is planning to [raise the default size in 2.0](https://github.com/GoogleChrome/puppeteer/issues/2079):
> set initial viewport size to 1280 x 1024

That said, I'm okay with us defaulting to `null` and letting our users set an explicit size if they want one (e.g. for consistent results across test runs).

rfcs/chromium-connector.md
antross: Ok.

rfcs/chromium-connector.md
molant: Cool, I'm ok with reviewing the defaults once `puppeteer` 2 is out. The proposed value seems more reasonable 😊

rfcs/chromium-connector.md
molant: I've added the `auth` property

rfcs/chromium-connector.md
molant: Updated envirnment variable names!

rfcs/chromium-connector.md
sarvaje: @molant, I don't see the `auth` property, did you upload the changes?

molant:
Closing this and opening the Epic in #2200 🎉

PR #2175, opened 2019-04-04
Fix: Repeated formatter when extending from several configurations

### LINKED ISSUE #2077: [Bug] Extending several configurations can cause the same formatter to output multiple times
# 🐛 Bug report

## Description

```json
{   
    "extends": [
        "web-recommended",
        "development"
    ]
}
```

The above will output the `summary` twice. When extending we should make sure the contents are unique. This probably applies to other types of resources (`parsers`?)

The change should be [around here in `resource-loader`](https://github.com/webhintio/hint/blob/62df2451d64ecaea43325dafad7c5765bbc31273/packages/hint/src/lib/utils/resource-loader.ts#L417)

sarvaje:
Fix #2077

## Pull request checklist

Make sure you:


sarvaje:
> I think we should do the same with parsers.

We are already doing it.
```js
if (finalConfig.parsers) {
    finalConfig.parsers = Array.from(new Set(finalConfig.parsers));
}
```

PR #2178, opened 2019-04-04
Fix: Add `gzip` option to request to decompress content.

### LINKED ISSUE #2068: hint-sri doesn't decompress requests on manual downloads
Hi @molant, I tried to narrow down the problem today.

I can confirm that everything still works as expected with hint v4.4.0, @hint/hint-sri v3.0.2 and @hint/connector-chrome v3.0.0.

When I update everything to the latest version I still get an error although it's a different one now. Only one resource is reported to have a wrong hash. I think the problem is that this particular resource gets loaded with the `request` module of Node.js and not with the browser connector. It's the only resource which goes through this code path in my example.

https://github.com/webhintio/hint/blob/master/packages/hint-sri/src/hint.ts#L422

My server only returns the resources encoded with GZIP but the `request` module doesn't decode GZIP content. The browser on the other hand does.

Do you think it's feasible to replace the `request` module with the connector that is otherwise used to download resources?

_Originally posted by @chrisguttandin in https://github.com/webhintio/hint/issues/2035#issuecomment-471225905_

sarvaje:
Fix #2068

## Pull request checklist

Make sure you:

~~- [ ] Added/Updated related tests.~~

### DIFF
--- a/packages/hint-sri/src/hint.ts
+++ b/packages/hint-sri/src/hint.ts
@@ -416,6 +416,7 @@ Actual:   ${integrities.join(', ')}`;
 
         try {
             response.body.content = await requestAsync({
+                gzip: true,
                 method: 'GET',
                 url: resource
             });


sarvaje:
I don't know why pipelines is failing, this is the error:

```
/azure-pipelines.yml: Could not find /azure-pipelines/scheduled.yml in repository self hosted on https://api.github.com using commit 387e0ce645dc19ef15c21e03ab6b9d80bcb8b0ff. GitHub reported the error, "Not Found"
```

molant:
@sarvaje I've rebased with master and force pushed here. Looks like that fixed the issue.

sarvaje:
> @sarvaje I've rebased with master and force pushed here. Looks like that fixed the issue.

Thanks!!

PR #2203, opened 2019-04-11
Improvements for testing on macOS

molant:
## Pull request checklist

Make sure you:

Couple commits to hopefully make things better in macOS (and in general?):

`canvas` 2.4.1 has a problem on macOS where it crashes if an svg does
not have dimensions. This sets the dimensions to 1x1

Fix webhintio#2113
Ref Automattic/node-canvas#1400

On macOS Chrome does not fully exit even if all the tabs are closed.
This causes a timeout when running the tests with `nyc` of about 2
minutes (or if the browser is manually closed). This addresses this
issue.
Because during testing multiple workers are using the same instance of
Chrome, one could try to close the browser while another is opening a
new tab and thus causing communication issues. The solution is to use
the same `lock` to launch the browser and to close it with a twist: The
`close` method gets all the possible Targets of type `page` (tabs),
close the ones related to the current session and see if there are any
other left. If there are not, that means it is safe to completely close
the browser by sending a `SIGKILL` (no other methods works because the
debugging protocol `client` cannot communicate in this state and
gracefully close the browser).
This issue was easier to reproduce on `hint-compat-api` because of the
ammount of separate tests being run.

Fix webhintio#2197

### DIFF
--- a/packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
+++ b/packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
@@ -82,6 +82,10 @@ export class Connector implements IConnector {
     private _waitForTarget: Promise<null>;
     /** Function to call when the target is downloaded. */
     private _targetReceived!: Function;
+    /** Lock file. */
+    private _lockFile = 'cdp.lock';
+    /** Lock status. */
+    private _isLocked = false;
 
     public constructor(engine: Engine, config: object, launcher: ILauncher) {
         const defaultOptions = {
@@ -400,26 +404,42 @@ export class Connector implements IConnector {
         return loadCDP();
     }
 
-    /** Initiates Chrome if needed and a new tab to start the collection. */
-    private async initiateComms() {
-        const cdpLock = 'cdp.lock';
-
+    private async lockFile() {
         try {
-            await lock(cdpLock, {
+            debug(`Trying to acquire lock`);
+            await lock(this._lockFile, {
                 pollPeriod: 500,
                 retries: 20,
                 retryWait: 1000,
                 stale: 50000,
                 wait: 50000
             });
+            this._isLocked = true;
+            debug(`Lock acquired`);
         } catch (e) {
             /* istanbul ignore next */
             { // eslint-disable-line
-                debug('Error while locking', e);
+                debug(`Error while locking`, e);
 
                 throw e;
             }
         }
+    }
+
+    private async unlockFile() {
+        if (this._isLocked) {
+            debug(`Trying to unlock`);
+            await unlock(this._lockFile);
+            debug(`Unlock successful`);
+        } else {
+            debug(`No need to unlock`);
+        }
+    }
+
+    /** Initiates Chrome if needed and a new tab to start the collection. */
+    private async initiateComms() {
+
+        await this.lockFile();
 
         const launcher: BrowserInfo = await this._launcher.launch(this._options.useTabUrl ? this._options.tabUrl : 'about:blank');
         let client;
@@ -467,7 +487,7 @@ export class Connector implements IConnector {
             });
         }
 
-        await unlock(cdpLock);
+        await this.unlockFile();
 
         return client;
     }
@@ -478,8 +498,9 @@ export class Connector implements IConnector {
     }
 
     /** Handles when we have been disconnected from the browser. */
-    private onDisconnect() {
+    private async onDisconnect() {
         debug(`Disconnected`);
+        await this.unlockFile();
     }
 
     /** Enables the handles for all the relevant Networks events sent by the debugging protocol. */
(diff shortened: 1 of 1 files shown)


packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -421,6 +423,18 @@ export class Connector implements IConnector {
                 throw e;
             }
         }
+    }
+
+    private async unlock() {
+        if (this._isLocked) {
+            await unlock(this._lockFile);
```
antross: I had to stare at this a bit before I realized it wasn't a recursive call. Changing the names of the private `lock` and `unlock` methods would make this clearer.

packages/hint-content-type/tests/tests.ts
```diff
@@ -10,7 +10,7 @@ import * as hintRunner from '@hint/utils-tests-helpers/dist/src/hint-runner';
 const hintPath = getHintPath(__filename);
 
 const pngImage = fs.readFileSync(`${__dirname}/fixtures/image.png`); // eslint-disable-line no-sync
-const svgImage = '<svg xmlns="http://www.w3.org/2000/svg"><path d="M1,1"/></svg>';
+const svgImage = '<svg xmlns="http://www.w3.org/2000/svg" width="1px" height="1px"><path d="M1,1"/></svg>';
```
antross: Given you opened #2204 later which also includes this change, I'm assuming you intend to rebase this PR once that one is merged, correct?

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
```diff
@@ -712,70 +726,53 @@ export class Connector implements IConnector {
...
-        debug(`Pending tabs: ${this._tabs.length}`);
-
-        while (this._tabs.length > 0) {
-            const tab = this._tabs.pop();
-
             try {
-                await cdp.Close({ id: tab.id, port: (this._client as any).port }); // eslint-disable-line new-cap
+                (this._client as any).close();
+
+                if (targets.size === 0 && this._pid) {
```
antross: For future reference we should put some version of your comment from the PR description either here or at the top of `close`:
> On macOS Chrome does not fully exit even if all the tabs are closed.
This causes a timeout when running the tests with nyc of about 2
minutes (or if the browser is manually closed). This addresses this
issue.
Because during testing multiple workers are using the same instance of
Chrome, one could try to close the browser while another is opening a
new tab and thus causing communication issues. The solution is to use
the same lock to launch the browser and to close it with a twist: The
close method gets all the possible Targets of type page (tabs),
close the ones related to the current session and see if there are any
other left. If there are not, that means it is safe to completely close
the browser by sending a SIGKILL (no other methods works because the
debugging protocol client cannot communicate in this state and
gracefully close the browser).

packages/hint-content-type/tests/tests.ts
molant: Yes, this PR started failing in CI (on my mac it runs pretty well) and I wanted to get this change sooner rather than later.

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
molant: Sounds good. I might have to tweak it after the castatrophic failure in CI. Maybe ava is doing something clever underneath when it detects that is CI that doesn't work nice with my changes (less workers?). I'm going to try to run this PR locally on Windows and see if I have to enable the kill signal only outside CI and/or macs

molant: Looks like this only repros in CI only. Tests are running fine on my Windows machine too 😒

molant: So the plot thickens...

If you set the environemt variable `SYSTEM_TEAMFOUNDATIONCOLLECTIONURI` to something so `is-ci` returns `true`, tests start blocking. At the beginning I thought it was something related to `chrome` and my changes but if I disable the `chrome` connector on `hint-amp-validator` it looks like the problem is after finishing the first `jsdom` test 🤯

```
Debugger attached.

  - [chrome] Valid AMP HTML passes
  - [chrome] Resource is not an HTML document
  - [chrome] Invalid AMP HTML fails
  - [chrome] Error downloading HTML doesn't fail
  √ [jsdom] Resource is not an HTML document (2s)
```

I have no idea what is going on

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
molant: changed them to `(un)lockFile` because I'm not very original today. I gladly accept better suggestions (and even worse ones if they are funny 😛)

molant:
This is ready for another review and hopefully things will be working now on CI.

molant:
CI passed and it tool less than 1h in mac and Linux 🎉

packages/utils-debugging-protocol-common/src/debugging-protocol-connector.ts
antross: Good enough for me. 👍

Issue #2211, opened 2019-04-11
[Bug] Source code snippets can be too large


antross:
Source code snippets in reports from hints include the entire `outerHTML` of a tag. This works well for items like `<link>`, but when a hint reports something on `<html>` a serialization of the entire document gets included.

These should be fixed to at least be truncated, but perhaps to only include the start tag of the element instead.

molant:
I think this should be done at the `report` level. Otherwise we will have to update all the packages to repeat the code.

What about:

* Include only the opening tag (regular expression?)
* Check the size and if it's still to long truncate it
  * How can truncate this correctly? I don't think we are providing the position relative in the element so if the tag is too long we might have to do something like `<div [...] failingAttribute="blah">` because otherwise we might crop it

Or maybe we just use the opening tag because it should still be a lot shorter than what we are providing right now.

@webhintio/contributors, thoughts?

antross:
I'm okay just including the opening tag regardless of length for now.

antross:
Thinking about this some more, I do think if the overall content is small enough we could still include the outerHTML. This may be useful for leaf items like `<button>`. Maybe we do a size check before switching to just the opening tag?

antross:
Also because `report` deals with different types of source code (e.g. CSS in addition to HTML) I'm not sure it's the right place to put this logic. We'd quickly end up with sniffing logic to determine what type of source code is being reported (though perhaps we should be able to specify the source code type anyway when calling `report`)...

Maybe a helper on our `HTMLElement` type to get the snippet would be better, and then updating hints to call this instead of `outerHTML`.

molant:
> We'd quickly end up with sniffing logic to determine what type of source code is being reported (though perhaps we should be able to specify the source code type anyway when calling report)...

That will be a bit of a change. The printing of the HTML is done already in `report`, the interesting part looks like this:

```js
if (element) {
     // When element is provided, position is an offset in the content.
    position = this.findProblemLocation(element, position);
    sourceCode = element.outerHTML.replace(/[\t]/g, '    ');
}
```

So if there's an `element` we do the `outerHTML`. Haven't checked what we are doing for JS or CSS but maybe `element` should be a bit more abstract and have a `toString()` that's smart enought to know what to do? And we will have an implementation for each type of resource we need.

The other option is to have other properties in the `options` the same way we have `element` (`css`,  `js`, etc.) but that doesn't scale very well...

sarvaje:
[...]

PR #2247, opened 2019-04-15
Fix: Include source code snippets in `hint-compat-api`

### LINKED ISSUE #2210: [Bug] `hint-compat-api` doesn't include source code snippets [1.7]
A number of other hints include snippets of source code along with reports to make it easier to find what needs to be changed. However, `hint-compat-api` does not include these in its reports, making it more difficult to get context.

sarvaje:
Fix #2210

## Pull request checklist

Make sure you:


packages/hint-compat-api/src/helpers/compat-css.ts
```diff
@@ -26,7 +26,7 @@ export class CompatCSS extends CompatBase<StyleEvents, StyleParse> {
     }
 
     public searchFeatures(parse: StyleParse): void {
-        this.walk(parse.ast);
+        this.walk(parse.ast, parse.code);
```
antross: I think we want something narrower than the entire CSS file. We should either omit this for CSS or only include code which is near the problem. 

If we want to serialize something custom I think having the selector plus the relevant CSS property/value (ignoring all other surrounding properties) would be good.

@webhintio/contributors Thoughts?

molant: Definitely, somehow I thought `walk` did this automatically and it was just a feature we weren't using 😖

sarvaje: I thought `pase.code` had only the feature hehehe. I will look into it.

sarvaje: done!

antross:
One more thing. This gave me the idea that we could link to the Elements panel for this too in the browser extension by passing along the selector and showing the first element which matches (which would get the target CSS active in the "Styles" sub-panel). 

Probably not worth doing in this PR, but could be a good follow-up task if we agree we like it.

packages/hint-compat-api/src/helpers/compat-html.ts
```diff
@@ -23,18 +23,20 @@ export class CompatHTML extends CompatBase<Events, Event> {
         this.walk((elementFound: ElementFound) => {
             const { element, resource } = elementFound;
             const location = element.getLocation();
+            const codeSnippet = element.outerHTML;
```
antross: Also, this can end up including a serialization of most of the page if an element like `<main>` is targeted. Let's add a `utils` helper to do the start-tag-only behavior we've been discussing in #2211. No need to update other hints in this PR, but I'd like to avoid adding new large serializations.

sarvaje: @antross done!

sarvaje:
I still need to update the documentation and add some tests.

sarvaje:
@molant @antross this should be ready to review.

packages/utils/src/report/get-css-code-snippet.ts
```diff
@@ -0,0 +1,111 @@
...
+    let numberOfSpaces = 4;
+    const grandpa = node.parent ? node.parent.parent : null;
+
+    if (grandpa && grandpa.type === 'atrule') {
+        numberOfSpaces = 8;
+    }
+
+    const spaces = new Array(numberOfSpaces + 1).join(' ');
+
+    return `${spaces}${content}`;
```
antross: This should be a little clearer (also requires deleting `const spaces = ...` above).

```suggestion
    return content.padStart(numberOfSpaces, ' ');
```

packages/utils/src/report/get-css-code-snippet.ts
```diff
@@ -0,0 +1,111 @@
...
+    let numberOfSpaces = 0;
+    const grandpa = node.parent ? node.parent.parent : null;
+
+    if (grandpa) {
+        numberOfSpaces = 4;
+    }
+
+    const spaces = new Array(numberOfSpaces + 1).join(' ');
+
+    return `${spaces}}`;
```
antross: Again, requires deleting `const spaces = ...` above.

```suggestion
    return '}'.padStart(numberOfSpaces, ' ');
```

packages/utils/src/report/get-css-code-snippet.ts
sarvaje: I didn't like the `new Array` solution, but I couldn't think in a shorter way to do it. Your solution is much better :)

antross: If only I had written it correctly the first time 🤦‍♂️ Thanks for debugging this with me!

```ts
return content.padStart(content.length + numberOfSpaces, ' ');
```

Issue #2139, opened 2019-03-29
[Feature] Migrate to puppeteer-core [0.1]


molant:
# 🚀 Feature request

## Description

Since puppeteer 1.7 there's [`puppeteer-core`](https://www.npmjs.com/package/puppeteer-core) that doesn't download chrome. 

 **What do you want to be added?**

We should probably use this package instead.

 **Are you willing to work on this?**

Yup

## What scenarios will this solve?

This will avoid having to download hundreds of MB and maybe #2113

molant:
We need to pass the executable path to `puppeteer` so I'm thinking about tackling this right after #2250 (which should be an util to find a `chromium` executable (via shortcut or default).

@webhint/contributors thoughts?

PR #2264, opened 2019-04-16
New: Util `chromium-finder`

### LINKED ISSUE #2260: New util to find chromium browser
This is a new util under `packages/utils` that allows to find the path of an installed chromium browser.
This will be based in[ `chrome-finder`](https://raw.githubusercontent.com/GoogleChrome/chrome-launcher/master/src/chrome-finder.ts) from [`chrome-launcher`](https://raw.githubusercontent.com/GoogleChrome/chrome-launcher).

List of browser to find (in Win, macOS, Linux, wsl?):

* Chrome: Stable, Canary
* Chromium
* Edge: Dev, Canary
* Opera: ?
* Brave: ?
* Others: ?

It will have a list of aliases (e.g.: edge-canary) that will automatically try to find the executable for that browser for the current platform.

It should also have a method to run the first one found. I think it should follow the same priority as chrome-launcher in case we end up contributing this upstream.

molant:
## Pull request checklist

Make sure you:

Fix #2260

I'll probably have to do more changes once I start working on #2139 and #2248 but this should be a good starting point.

### DIFF
--- a/packages/utils/src/chromium-finder.ts
+++ b/packages/utils/src/chromium-finder.ts
@@ -1,0 +1,389 @@
+/**
+ * @license Copyright 2016 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import * as path from 'path';
+import { isFile } from './fs/is-file';
+import { isDirectory } from './fs/is-directory';
+import { getVariable, setVariable } from './misc/environment';
+import { getPlatform } from './misc/get-platform';
+
+import { execSync, execFileSync } from 'child_process';
+
+const newLineRegex = /\r?\n/;
+
+export enum Browser {
+    Chrome,
+    Chromium,
+    Edge
+}
+
+const ERRORS = {
+    InvalidPath: 'The provided path is not accessible',
+    NoInstallationFound: 'No installation found',
+    NotSupportedBrowser: 'The provided browser is not supported in this platform',
+    UnsupportedPlatform: 'Unsupported platform'
+};
+
+/**
+ * Variables used to find installers sorted by priority:
+ * - Chrome: Canary, Stable
+ * - Chromium: Only one install
+ * - Edge: Canary, Dev (no current beta or stable channels)
+ */
+const browserVariables = new Map([
+    ['darwin', new Map([
+        [
+            Browser.Chrome, [
+                '/Contents/MacOS/Google Chrome Canary',
+                '/Contents/MacOS/Google Chrome'
+            ]
+        ],
+        [
+            Browser.Chromium, [
+                '/Contents/MacOS/Chromium'
+            ]
+        ]
+    ])],
+    ['linux', new Map([
+        [
+            Browser.Chrome, [
+                // TODO: Chrome Canary? Chrome Dev?
+                '(google-chrome|chrome)', // regex for desktop
+                'google-chrome-stable', // used with `which`
+                'google-chrome' // used with `which`
+            ]
+        ],
+        [
+            Browser.Chromium, [
+                '(chromium)', // regex for desktop
+                'chromium-browser', // used with `which`
+                'chromium' // used with `which`
+            ]
+        ]
+    ])],
+    ['win32', new Map([
+        [
+            Browser.Chrome, [
+                `${path.sep}Google${path.sep}Chrome SxS${path.sep}Application${path.sep}chrome.exe`,
+                `${path.sep}Google${path.sep}Chrome${path.sep}Application${path.sep}chrome.exe`
+            ]
+        ],
+        [
+            // TODO: It seems like there's no official installer for chromium?
+            Browser.Chromium, [
+                `${path.sep}Chromium${path.sep}Application${path.sep}chromium.exe`
+            ]
+        ],
+        [
+            // TODO: Not sure how to identify the browser correctly, I believe it depends on which one you install first?
+            Browser.Edge, [
+                // Canary suffix path
+                `${path.sep}Microsoft${path.sep}Edge SxS${path.sep}Application${path.sep}msedge.exe`,
+                `${path.sep}Microsoft${path.sep}Edge${path.sep}Application${path.sep}msedge.exe`,
+                // Dev suffix path
+                `${path.sep}Microsoft${path.sep}Edge Dev SxS${path.sep}Application${path.sep}msedge.exe`,
+                `${path.sep}Microsoft${path.sep}Edge Dev${path.sep}Application${path.sep}msedge.exe`
+            ]
+        ]
+    ])]
+]);
+
+
+/** Find chromium executables on macOS */
+const darwin = (browser: Browser) => {
+    const platformBrowserInfo = browserVariables.get('darwin')!;
+    const suffixes = platformBrowserInfo.get(browser);
+
+    if (!suffixes) {
+        throw new Error(ERRORS.NotSupportedBrowser);
+    }
+
+    /**
+     * The following generates a list of all the installed applications in macOS.
+     * The `grep` command makes sure we only get the lines that end with `.app`
+     * because `-dump` generates more information than the one needed.
+     * The `awk` commands makes sure we remove the initial part of the line so we
+     * only get the path.
+     *
+     * More info in:
+     * http://krypted.com/mac-security/lsregister-associating-file-types-in-mac-os-x/
+     */
+    const LSREGISTER = '/System/Library/Frameworks/CoreServices.framework' +
+        '/Versions/A/Frameworks/LaunchServices.framework' +
+        '/Versions/A/Support/lsregister';
+
+    const lines = execSync(
+        `${LSREGISTER} -dump` +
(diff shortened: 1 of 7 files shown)


molant:
I've changed the implementation of `is-file` based on the recommendations of the docs and that's probably why tests are failing. Probably a different error message? I'll take a look.

packages/utils/src/fs/is-file.ts
```diff
@@ -1,11 +1,11 @@
-import { statSync } from 'fs';
+import { accessSync } from 'fs';
 
 /** Check if a path is a file and exists. */
 export const isFile = (filePath?: string): boolean => {
     try {
-        const stats = statSync(filePath!);
+        accessSync(filePath!);
```
sarvaje: This is not equivalent to `statSync`. This method will work for a file or a directory, and `isFile` should check if the path is a file.

packages/utils/tests/chromium-finder.ts
```diff
@@ -0,0 +1,235 @@
...
+test(`Invalid custom path throws an exception`, (t) => {
+    const sandbox = t.context.sandbox;
+
+    sandbox.stub(t.context.isFile, 'isFile')
+        .onFirstCall()
+        .returns(false);
+
+    const chromiumFinder = loadDependency(t.context);
+
+    t.throws(() => {
```
sarvaje: check the exception thowed is the right one.

packages/utils/src/chromium-finder.ts
```diff
@@ -0,0 +1,391 @@
...
+
+    for (const folder of desktopInstallationFolders) {
+        const executable = findChromeExecutable(folder, executables);
+
+        if (executable) {
+            return executable;
+        }
+    }
+
+    // Look for executables by using the which command
```
molant:
```suggestion
    // 2. Look for executables by using the which command
```

packages/utils/src/chromium-finder.ts
```diff
@@ -0,0 +1,391 @@
...
+};
+
+/**
+ * Searches for installation paths for the given `browser` on a Windows platform.
+ * @param browser The type of browser to search for
+ */
+const win32 = (browser: Browser) => {
+    const info = browserVariables.get('win32');
+
+    if (!info) {
```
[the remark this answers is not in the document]
molant: You are right

packages/utils/src/chromium-finder.ts
```diff
@@ -0,0 +1,391 @@
...
+
+    if (!suffixes) {
+        throw new Error(ERRORS.NotSupportedBrowser);
+    }
+
+    const prefixes = [
+        getVariable('LOCALAPPDATA'),
+        getVariable('PROGRAMFILES'),
+        getVariable('PROGRAMFILES(X86)')
+    ].filter(Boolean);
```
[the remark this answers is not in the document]
molant: A lot of this file is [copy pasta from chrome-finder in the chrome-launcher project](https://github.com/GoogleChrome/chrome-launcher/blob/master/src/chrome-finder.ts). In this specific case it's the `Boolean` constructor which will filter any environment variable that is not defined or returns `''`.

It also took me a moment to figure it out when I was looking the other day but because we might contribute changes upstream I didn't want to change too much their logic.

packages/utils/src/fs/is-file.ts
molant: Are you sure? 

[The docs say](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_stat_path_options_callback):

> To check if a file exists without manipulating it afterwards, fs.access() is recommended.

I know that some tests are failing probably because of this but haven't dig into those yet.

molant: I've reverted the changes just in case this was the issue with the babel parser.

Also from the docs:

> Using fs.stat() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Instead, user code should open/read/write the file directly and handle the error raised if the file is not available.

Not sure why it's not recommended though 🤔

packages/utils/src/chromium-finder.ts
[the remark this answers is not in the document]
sarvaje: Ahh, ok, it is the first time I see that xD

packages/utils/src/fs/is-file.ts
sarvaje: > To check if a file exists without manipulating it 

The thing is that isFile is not checking if a file exists, isFile is checking if the path is a file.

sarvaje: > Not sure why it's not recommended though

I don't know neither. I don't like exceptions hehehe

molant: Well, I've reverted the code so tests are passing again.

packages/utils/src/chromium-finder.ts
```diff
@@ -0,0 +1,388 @@
...
+
+    return '';
+};
+
+/**
+ * Searches for installation paths for the given `browser` on a Windows platform.
+ * @param browser The type of browser to search for
+ */
+const win32 = (browser: Browser) => {
+    const info = browserVariables.get('win32');
```
antross:
```suggestion
    const info = browserVariables.get('win32')!;
```

packages/utils/src/chromium-finder.ts
```diff
@@ -0,0 +1,388 @@
...
+
+    /** The order in which to search for browsers in case none are provided. */
+    const browsers = [Browser.Chrome, Browser.Chromium, Browser.Edge];
+    let browserFound = '';
+
+    while (browsers.length > 0 && !browserFound) {
+        const br = browsers.shift();
+
+        try {
+            browserFound = findBrowserPath(br!);
```
molant:
```suggestion
            browserFound = findBrowserPath(br);
```

packages/utils/src/chromium-finder.ts
```diff
@@ -0,0 +1,388 @@
...
+const resolveChromiumPath = () => {
+    const chromiumPaths = [
+        getVariable('WEBHINT_CHROMIUM_PATH'),
+        getVariable('CHROME_PATH')
+    ];
+
+    while (chromiumPaths.length > 0) {
+        const browserPath = chromiumPaths.shift();
+
+        if (canAccess(browserPath!)) {
```
molant:
```suggestion
        if (canAccess(browserPath)) {
```

packages/utils/src/chromium-finder.ts
```diff
@@ -0,0 +1,388 @@
...
+};
+
+/**
+ * Searches for installation paths for the given `browser` on a Windows platform.
+ * @param browser The type of browser to search for
+ */
+const win32 = (browser: Browser) => {
+    const info = browserVariables.get('win32');
+
+    const suffixes = info!.get(browser);
```
molant:
```suggestion
    const suffixes = info.get(browser);
```

packages/utils/src/chromium-finder.ts
```diff
@@ -0,0 +1,388 @@
...
+
+/** Find chromium executables on Linux. */
+const findChromeExecutable = (folder: string, regExes: string[]) => {
+    const argumentsRegex = /(^[^ ]+).*/; // Take everything up to the first space
+
+    for (const browserRegex of regExes) {
+
+        const chromeExecRegex = `^Exec=/.*/${browserRegex}-.*`;
+
+        if (canAccess(folder)) {
```
sarvaje: if you have restored `isFile`, then is not going to work as expected, right?

packages/utils/tests/chromium-finder.ts
```diff
@@ -0,0 +1,237 @@
...
+        chromiumFinder.getInstallationPath({ browser: 'Chrome' });
+    });
+
+    t.true(setVariableSpy.calledThrice, `setVariable isn't called 3 times`);
+    t.is(setVariableSpy.firstCall.args[0], 'LOCALAPPDATA', `First 'setVariable' call isn't LOCALAPPDATA`);
+    t.is(setVariableSpy.secondCall.args[0], 'PROGRAMFILES', `Second 'setVariable' call isn't PROGRAMFILES`);
+    t.is(setVariableSpy.thirdCall.args[0], 'PROGRAMFILES(X86)', `Third 'setVariable' call isn't PROGRAMFILES(X86)`);
```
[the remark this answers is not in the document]
molant: This is more for when we are running tests so we know faster where things are breaking. Not sure what the output will be with `ava` here.
Want me to remove the messages of all of the above as well?

packages/utils/src/chromium-finder.ts
molant: It's going to work as expected. I'm just using the same name of the variable as `chrome-finder` upstream. In their code they assume everything they check for is a file so this is fine.

packages/utils/tests/chromium-finder.ts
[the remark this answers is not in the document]
antross: My understanding was `ava` included sufficient context in the output, [some details here](https://github.com/avajs/ava/blob/master/docs/03-assertions.md#enhanced-assertion-messages). 

I thought I had also read an `ava` FAQ a few months ago that recommended omitting these for this reason, but I can't seem to dig it up at the moment.

And yes, I was hoping to have the messages removed from all of the above as well.

packages/utils/src/chromium-finder.ts
sarvaje: so `folder` is a file?

molant: It should be a file, don't know why they call it folder but now I'm hesitating. This is for Linux so not easy for me to test 🤔

molant: So this should probably use `is-dir`. Nice catch.

molant:
Changed applied. Ready for a last round. I'll start working on `puppeteer-core` in `extension-browser` from this now that things seem to be pretty stable.

PR #2266, opened 2019-04-16
Chore: Inline analytics script

### LINKED ISSUE #2230: [browser extension] Inline analytics package [0.25]
Firefox extensions don't like loading external resources so we have to inline our analytics package.

antross:
## Pull request checklist

Make sure you:

Fix #2230

Also switch to using the "basic" v2 version of Application Insights.

Raising the bundle size limit by 10KB to account for the additional code. I'll lower this again after #2097.

### DIFF
--- a/packages/extension-browser/src/devtools/utils/analytics.ts
+++ b/packages/extension-browser/src/devtools/utils/analytics.ts
@@ -1,33 +1,53 @@
-import { AppInsights } from 'applicationinsights-js';
+import { ApplicationInsights } from '@microsoft/applicationinsights-web-basic';
+
+const instrumentationKey = '8ef2b55b-2ce9-4c33-a09a-2c3ef605c97d';
+
+let appInsights: ApplicationInsights;
+
+const trackEvent = (name: string, properties?: { [key: string]: string }, measurements?: { [key: string]: number }) => {
+    if (!appInsights) {
+        return;
+    }
+
+    appInsights.track({
+        baseData: {
+            measurements,
+            name,
+            properties
+        },
+        baseType: 'EventData',
+        name: `Microsoft.ApplicationInsights.${instrumentationKey}.Event`
+    });
+};
 
 import { ErrorData } from '../../shared/types';
 
 /** Called to initialize the underlying analytics library. */
 export const setup = () => {
-    AppInsights.downloadAndSetup!({ instrumentationKey: '8ef2b55b-2ce9-4c33-a09a-2c3ef605c97d' });
+    appInsights = new ApplicationInsights({ instrumentationKey });
 };
 
 /** Called when analysis was canceled by the user. */
 export const trackCancel = (duration: number) => {
-    AppInsights.trackEvent('f12-cancel', undefined, { 'f12-cancel-duration': duration });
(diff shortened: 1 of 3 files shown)


packages/extension-browser/src/devtools/utils/analytics.ts
```diff
@@ -1,26 +1,46 @@
...
+
+let appInsights: ApplicationInsights;
+
+const trackEvent = (name: string, properties?: { [key: string]: string }, measurements?: { [key: string]: number }) => {
+    if (!appInsights) {
+        return;
+    }
+
+    appInsights.track({
+        baseData: {
```
molant: Not sure what `baseData` can have but I think we should indicate somehow where the extension is running (Edge, Chrome, Firefox...).  That will help us know if we have reliability issues in one more than another and which one is more popular.

packages/extension-browser/src/devtools/utils/analytics.ts
```diff
@@ -1,26 +1,46 @@
...
 /** Called when the "Hints" tab was opened by the user. */
 export const trackShow = () => {
-    AppInsights.trackEvent('f12-show');
+    trackEvent('f12-show');
 };
 
 /** Called when analysis was started by the user. */
 export const trackStart = () => {
-    AppInsights.trackEvent('f12-start');
+    trackEvent('f12-start');
```
molant: Can we have an "install" event? Or something that gets send the first time the user goes to the `hint` pannel.

packages/extension-browser/src/devtools/utils/analytics.ts
antross: No need in this case. Application Insights already includes data about the browser and OS in submissions (though its logic does need to be updated to account for the new Edge).

packages/extension-browser/src/devtools/utils/analytics.ts
antross: We could do a "first-show" event, with the caveat that it would be reset by clearing data for or re-installing the extension. 

For true "installs" we should rely on the dashboards for whichever extension stores we're published in. I'd probably prefer going this route and not adding another event at the moment. Thoughts?

antross: I should add that application insights also tags unique "users" so we can probably get a rough count of installs that way as well without adding another event.

molant: OK, then let's not do anything else here :)

packages/extension-browser/src/devtools/utils/analytics.ts
molant: Nice

molant:
@antross there is also some code coverage problems in `extension-browser`:

> ERROR: Coverage for branches (78.07%) does not meet global threshold (80%)

antross:
@molant I decided to raise the size limit by 10KB for now with the expectation that I'll lower it again once I remove the `caniuse` dependency. That way we're not holding this PR in the queue longer than it needs to be.

molant:
> I decided to raise the size limit by 10KB for now

SGTM!

PR #2269, opened 2019-04-17
Fix: Error location for HTTP headers and added code snippet.

### LINKED ISSUE #2209: [Bug] Bogus error locations for HTTP Headers
Hints like `hint-content-type` which point to HTTP headers also include a location, however this location appears to point to a location in the HTML file that triggered the request (as opposed to the listed resource that contains the HTTP header).

This should be fixed to omit line/column information in these cases as it doesn't make sense.

Actual:
```
'content-type' header charset value should be 'utf-8'
https://www.google-analytics.com/analytics.js:1:17293
```

Expected:
```
'content-type' header charset value should be 'utf-8'
https://www.google-analytics.com/analytics.js
```

sarvaje:
Fix #2209

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint/src/lib/hint-context.ts
+++ b/packages/hint/src/lib/hint-context.ts
@@ -33,6 +33,8 @@ export type ReportOptions = {
     location?: ProblemLocation | null;
     /** The `Severity` to report the issue as (overrides default settings for a hint). */
     severity?: Severity;
+    /** Indicate the language of the codeSnippet */
+    codeLanguage?: string;
 };
 
 /** Acts as an abstraction layer between hints and the main hint object. */
@@ -117,8 +119,8 @@ export class HintContext<E extends Events = Events> {
     /** Reports a problem with the resource. */
     public report(resource: string, message: string, options: ReportOptions = {}) {
         const { codeSnippet, element, severity } = options;
-        let position = options.location || null;
         let sourceCode: string | null = null;
+        let position = options.location || null;
 
         if (element) {
             // When element is provided, position is an offset in the content.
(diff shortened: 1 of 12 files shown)


sarvaje:
@molant @antross can you take a look if you like the solution. If so, I will continue working to apply it to other hints.

molant:
I like where this is going :)

packages/hint/src/lib/hint-context.ts
```diff
@@ -138,7 +140,8 @@ export class HintContext<E extends Events = Events> {
             codeSnippet || sourceCode || '',
             position,
             message,
-            resource
+            resource,
+            options.codeLanguage
```
antross: Not sure we need to do it in this PR, but it'd be nice to refactor `engine.report` to take something similar to `ReportOptions` instead of continuing to add extra parameters...

molant: I've opened https://github.com/webhintio/hint/issues/2276

sarvaje:
> One request: let's capitalize header names in the code snippets (e.g. Content-Type: ... instead of content-type: ...) to align with how they're typically displayed in the Network tab of the devtools

Ok! I will do it.

packages/hint/src/lib/hint-context.ts
sarvaje: I think this is a breaking change, so it would be nice to have it sooner rather than later.

sarvaje:
@antros @molant ready to review again

packages/hint/src/lib/hint-context.ts
molant: I agree, we have quite a few breaking changes already. The blogpost is going to be quite big...

antross: "How we rebuilt webhint from the ground up..." 🤣

PR #2273, opened 2019-04-18
Fix/puppeteer core

### LINKED ISSUE #2139: [Feature] Migrate to puppeteer-core [0.1]

molant:
## Pull request checklist

Make sure you:

`puppeteer-core` does not include a bundled version of Chromium so
installation should be faster.
Also there is a different `@types` package so reference needs to be
updated as well.

Fix #2139

- - - - - - - - - - - - - - - - - - -

This PR depends on #2264, that one should get merged first and then I'll rebase.
Also works on my machine™ but it will be good if someone could test it locally.
There seems to be a problem with tests coverage in my machine with `extension-browser` that I don't understand as I've only modified the tests (I might have to rebuild something else) 🤔


packages/extension-browser/tests/devtools.ts
```diff
@@ -98,7 +98,8 @@ test('It builds a configuration, starts a scan, and displays results', async (t)
      * Launch the browser and get a reference to the initial page.
      * Note: Uncomment config options to see page content (aids debugging).
      */
-    const browser = await launch(/* { defaultViewport: null, headless: false } */);
+    const executablePath = chromiumFinder.getInstallationPath();
+    const browser = await launch(/* { defaultViewport: null, headless: false } */{ executablePath});
```
antross:
```suggestion
    const browser = await launch(/* { defaultViewport: null, headless: false } */{ executablePath });
```

molant: Need to find a setting to make sure this doesn't happen. We've had quite a few of these in the past and eslint doesn't flag them 🤔

molant:
I've rebased with master. @sarvaje want to take a look before I merge?

PR #2280, opened 2019-04-18
Chore: Properly close resources on failure

### LINKED ISSUE #2236: [Bug] Some failing tests hang CI and don't stop the process [0.1]
# 🐛 Bug report

## Description

The test script might be capturing too eagerly when a test fails or something might be going on with `nyc` and `ava` when something fails. This seems to happen more often on macOS (😞) than other platforms.

[Example here](https://dev.azure.com/webhint/webhint/_build/results?buildId=3182&view=logs&jobId=790316ed-8aa2-568f-3859-3e923bc648be&taskId=4d28356e-2121-5462-4447-b94ded044ac4&lineStart=3431&lineEnd=3432&colStart=1&colEnd=1)

We should add a safegard in the test script that if running a command takes more than XX minutes it fails on top of trying to figure out what's going on too.

molant:
## Pull request checklist

Make sure you:

Fix #2236

### DIFF
--- a/packages/utils-tests-helpers/src/hint-runner.ts
+++ b/packages/utils-tests-helpers/src/hint-runner.ts
@@ -222,11 +222,16 @@ export const testHint = (hintId: string, hintTests: HintTest[], configs: { [key:
 
     /** Runs a test for the hint being tested */
     const runHint = async (t: ExecutionContext<HintRunnerContext>, hintTest: HintTest, connector: string) => {
+        let engine,
+            server;
+
         try {
-            const server = await Server.create({ configuration: hintTest.serverConfig, isHTTPS: configs.https });
+            server = await Server.create({ configuration: hintTest.serverConfig, isHTTPS: configs.https });
+
             const { serverUrl, reports } = hintTest;
             const target = serverUrl ? serverUrl : `${configs.https ? 'https' : 'http'}://localhost:${server.port}/`;
-            const engine = await createConnector(t, hintTest, connector);
+
+            engine = await createConnector(t, hintTest, connector);
             const results = await engine.executeOn(new URL(target));
 
             const sources = new Map<string, string>();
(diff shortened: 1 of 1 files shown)


packages/utils-tests-helpers/src/hint-runner.ts
```diff
@@ -244,6 +249,14 @@ export const testHint = (hintId: string, hintTests: HintTest[], configs: { [key:
         } catch (e) {
             console.error(e);
 
+            if (server) {
+                server.stop();
```
antross: Should this also be an `await`?
```suggestion
                await server.stop();
```

packages/utils-tests-helpers/src/hint-runner.ts
```diff
@@ -244,6 +249,14 @@ export const testHint = (hintId: string, hintTests: HintTest[], configs: { [key:
...
+            }
+
+            if (engine) {
+                await engine.close();
+            }
+
             return t.fail(`${hintTest.name} throwed an exception:\n${e.message}\n${e.stack}`);
```
antross: Should some version of `stopConnector` also be called here? Not sure if `hintTest.after` ever performs cleanup, but if it does it won't get run.

If so, perhaps we should just move all of the cleanup to a `finally` block.

molant: Not super concerned about not calling `hintTest.after`. The test has already failed so if there's ever a `test.after` action it doesn't matter that much.

packages/utils-tests-helpers/src/hint-runner.ts
molant: Yup

packages/utils-tests-helpers/src/hint-runner.ts
antross: ok

PR #2288, opened 2019-04-21
New: Add hint for manifest-icons

### LINKED ISSUE #462: Add rule to check the usage of web app manifest's `icons`
- [x] Check if at least required sizes icon are specified (`512x512` and `192x192`) 
- [x] Check if icons actually exists.
- [x] Check if [`type` is specified](https://github.com/h5bp/html5-boilerplate/pull/1629#r30987965).
- [x] Check if `sizes` is accurate.

shivangg:
## Pull request checklist

Make sure you:

Fixes #462


molant:
@shivangg you have a few linting errors. Can you please fix them?

Thanks!

molant:
I think that runnning `yarn lint:js --fix` should fix a bunch of those

shivangg:
Was not getting the eslint errors in VS Code as I had only added the 2 individual folders rather than the full monorepo. Resolved now.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+ */
+
+export default class ManifestIconHint implements IHint {
+    public static readonly meta = meta;
+    public constructor(context: HintContext<ManifestEvents>) {
+        /**
+         * Try to see if the `icon` file actually
+         * exists and is accessible.
+         */
+        const iconExists = async (IconPath: string) => {
```
molant:
```suggestion
        const iconExists = async (iconPath: string) => {
```

You'll have to rename it everywhere.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
+/**
+ * @fileoverview Checks usage of Web App Manifest icons
+ */
+import { URL } from 'url';
+import * as getImageData from 'image-size';
+import { HintContext } from 'hint/dist/src/lib/hint-context';
+// The list of types depends on the events you want to capture.
```
molant: Remove comment
```suggestion
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+
+            try {
+                image = getImageData(response.body.rawContent);
+            } catch (e) {
+                if (e instanceof TypeError) {
+                    const message = `'${IconPath}' should be a valid PNG image.`;
+
+                    context.report(IconPath, message);
+                } else {
+                    debug(`'getImageData' failed for '${IconPath}'`);
```
molant: Should we still report that something happened with the image file? Under what situations can `getImageData` fail?

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+                }
+
+                return;
+            }
+
+            // Check if the image is a PNG.
+            if (image.type !== 'png') {
+                const message = `'${IconPath}' should be a PNG image but found be a
```
molant: What about:
```suggestion
                const message = `Manifest icon should be of type "png" but found "${image.type}"`;
```

We are already providing `iconPath` in `report` so no need to put it in the message too.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+             */
+        };
+
+        /**
+         *
+         * @param icons array of the icons properties
+         * @param hostnameWithProtocol
+         */
+        const validIconsURLs = (
+            icons: ManifestImageResource[],
```
molant: Let's put all the parameters in the same line.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+            hostnameWithProtocol: string
+        ) => {
+            icons.forEach(async (iconProps: { src: string }) => {
+                await iconExists(`${hostnameWithProtocol}/${iconProps.src}`);
```
molant: IIRC this is not going to work as expected. `forEach` doesn't `await`.
There are  a few options on how to solve this:
1. Using `map` an `Promise.all`:

```ts
const iconsPromises = icons.map((iconProps) => {
  return iconExists(`${hostnameWithProtocol}/${iconProps.src}`);
});

await Promise.all(icons);
```

2. Using `for ...of...`:

```ts
for(const icon of icons) {
  await iconExists(`${hostnameWithProtocol}/${iconProps.src}`);
}
```

`1` has the advantage of being faster if there are multiple `icons`. The second is easier to read. Don't have a strong preference. @webhintio/contributors ?

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+        /**
+         *
+         * @param icons array of the icons properties
+         * @param hostnameWithProtocol
+         */
+        const validIconsURLs = (
+            icons: ManifestImageResource[],
+            hostnameWithProtocol: string
+        ) => {
+            icons.forEach(async (iconProps: { src: string }) => {
```
molant: Don't think you need to put the type
```suggestion
            icons.forEach(async (iconProps) => {
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+
+            debug(`Validating hint manifest-icon`);
+            /**
+             * console.log(`\n--------
+             * ${JSON.stringify(hostnameWithProtocol, null, '\t')}
+             *--------`);
+             */
+
+            if (icons !== undefined) {
+                validIconsURLs(icons, hostnameWithProtocol);
```
molant: You'll have to `await` here with the changes I mention earlier.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,150 @@
...
+            /*
+             * This is where all the magic happens. Any errors found should be
+             * reported using the `context` object. E.g.:
+             * context.report(resource, 'Add error message here.');
+             *
+             * More information on how to develop a hint is available in:
+             * https://webhint.io/docs/contributor-guide/hints/
+             */
+
+            if (Math.ceil(Math.random()) === 0) {
```
molant: Remove this code so you can actually test this correctly :)

shivangg: Yeah, wasn't testing anything that time.

packages/hint-manifest-icons/src/hint.ts
shivangg: Okay.

packages/hint-manifest-icons/src/hint.ts
shivangg: Removed.

packages/hint-manifest-icons/src/hint.ts
shivangg: Going with the 2nd approach for now. Validating for other points in the `for` loop as well.

packages/hint-manifest-icons/src/hint.ts
shivangg: okay.

packages/hint-manifest-icons/src/hint.ts
shivangg: Yeah, that makes sense.

packages/hint-manifest-icons/src/hint.ts
shivangg: Removed.

packages/hint-manifest-icons/src/hint.ts
shivangg: Done.

packages/hint-manifest-icons/src/hint.ts
shivangg: I thought it might be failing in the case of corrupt `response.body.rawContent`. This was also done in the `apple-touch-icons`

molant: Sounds good. Can you please add a comment with the reasoning?

shivangg: Okay.

packages/hint-manifest-icons/src/meta.ts
```diff
@@ -0,0 +1,21 @@
+import { Category } from 'hint/dist/src/lib/enums/category';
+import { HintScope } from 'hint/dist/src/lib/enums/hint-scope';
+import { HintMetadata } from 'hint/dist/src/lib/types';
+
+const meta: HintMetadata = {
+    docs: {
+        category: Category.pwa,
+        description: `Description for manifest-icons`
```
antross: Don't forget to fill this out. Also, add a `name: "Manifest has icons"` property too (not required in general, but we add it to all the hints we ship as part of webhint).

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            }
+
+            const specifiedType = iconType.split('/')[1];
+
+            /** Handling for the corrupt rawContent */
+            if (rawContent !== null) {
+                const image: ImageTypeResult | null = imageType(rawContent);
+
+                if (image !== null) {
+                    const isValidType: boolean = allowedTypes.some((item) => {
```
antross: This can be a bit simpler, though you'll also need to delete the next couple of lines... I'm eager for GitHub to support multi-line suggestions in PRs 😊
```suggestion
                    const isValidType = allowedTypes.includes(image.ext);
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+                context.report(iconPath, message);
+
+                return false;
+            }
+
+            const specifiedType = iconType.split('/')[1];
+
+            /** Handling for the corrupt rawContent */
+            if (rawContent !== null) {
+                const image: ImageTypeResult | null = imageType(rawContent);
```
antross: I usually try to avoid specifying types that get inferred correctly (keeps the code a bit closer to JS). Here we'll get the correct type from the function.
```suggestion
                const image = imageType(rawContent);
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+    public constructor(context: HintContext<ManifestEvents>) {
+        const getIconFullPath = (iconSrc: string, hostnameWithProtocol: string) => {
+            return `${hostnameWithProtocol}/${iconSrc}`;
+        };
+
+        /**
+         * See if the `icon` file actually
+         * exists and is accessible.
+         */
+        const iconExists = async (iconPath: string) => {
```
antross: While I generally prefer omitting types that can be inferred, return types in function signatures are one exception because they can easily become something you didn't expect without reading the full implementation.
```suggestion
        const iconExists = async (iconPath: string): Buffer | null => {
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            if (rawContent !== null) {
+                const image: ImageTypeResult | null = imageType(rawContent);
+
+                if (image !== null) {
+                    const isValidType: boolean = allowedTypes.some((item) => {
+                        return item === image.ext;
+                    });
+
+                    if (specifiedType !== image.ext) {
+                        const message = `Real image type (${image.ext}) do not match with specified type (${specifiedType})`;
```
antross: We should also validate that the `Content-Type` header we received matched `specifiedType`. This will likely require returning a bit more data from `iconExists` so we can check the headers here in addition to the raw content.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+
+            return response.body.rawContent;
+        };
+        /**
+         * Passes only for the PNG files
+         * @param iconType type specified in the manifest file
+         * @param rawContent raw datastream
+         * @param iconPath icon resource path
+         */
+        const validateImageType = (iconType: string | undefined, rawContent: Buffer, iconPath: string) => {
```
antross:
```suggestion
        const validateImageType = (iconType: string | undefined, rawContent: Buffer, iconPath: string): boolean => {
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            return false;
+        };
+
+        /**
+         *
+         * @param iconSizes Sizes specified in the manifest file
+         * @param iconRawData
+         * @param iconPath full path to the icon file
+         */
+        const validateSizes = (iconSizes: string | undefined, iconRawData: Buffer, iconPath: string) => {
```
antross:
```suggestion
        const validateSizes = (iconSizes: string | undefined, iconRawData: Buffer, iconPath: string): boolean => {
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            if (iconSizes === undefined) {
```
antross:
```suggestion
            if (!iconSizes) {
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+                return false;
+            }
+            /**
+             * sizes can be the string 'any' OR
+             * two non-negative integers without leading 0s and separated by 'x' like 144x144
+             */
+            if (iconSizes === 'any') {
+                return false;
+            }
+            const specifiedSize = iconSizes.split('x');
```
antross: I'd recommend a newline after the `if` (actually thought our linter required that...)
```suggestion

            const specifiedSize = iconSizes.split('x');
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+
```
antross: But the variable declarations don't need to be separated by blank lines.
```suggestion
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            const realImage = getImageData(iconRawData);
+
```
antross:
```suggestion
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            const realSize = [realImage.width.toString(), realImage.height.toString()];
+
```
antross:
```suggestion
```

packages/hint-manifest-icons/src/hint.ts
antross: Per https://developer.mozilla.org/en-US/docs/Web/Manifest#icons this can also be a space-separated list of sizes (mostly used for the `ico` format). We should probably warn if there's more than one size specified since our supported formats will only have a single size.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            const sizesMatch = JSON.stringify(specifiedSize) === JSON.stringify(realSize);
```
antross: This is a simple structure so let's compare each part separately (as numbers) so it's clear what's being checked.
```suggestion
            const specifiedWidth = parseInt(specifiedSize[0]);
            const specifiedHeight = parseInt(specifiedSize[1]);
            const sizesMatch = specifiedWidth === realImage.width && specifiedHeight === realImage.height;
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            const requiredSizes = ['192x192', '512x512'];
+
+            const requiredSizesNotFound = requiredSizes.filter((size) => {
+                return !validSizes.includes(size);
+            });
+
+            if (requiredSizesNotFound.length > 0) {
+                const message = `Required sizes ${JSON.stringify(requiredSizesNotFound)} not found.`;
+
+                context.report(icons, message);
```
antross: Each report that can reference a particular point in the manifest should include a location (see `hint-manifest-app-name` for an example of using the `getLocation` helper on the `ManifestParsed` event to do this). Also update the tests to validate this location. 
```suggestion
                context.report(icons, message, { location });
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            const realImage = getImageData(iconRawData);
+
+            const realSize = [realImage.width.toString(), realImage.height.toString()];
+
+            const sizesMatch = JSON.stringify(specifiedSize) === JSON.stringify(realSize);
+
+            if (!sizesMatch) {
+                const message = `Real image size (${JSON.stringify(realSize)}) do not match with specified size (${specifiedSize})`;
+
+                context.report(iconPath, message);
```
antross: Include the location in this report.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+                    }
+
+                    if (isValidType) {
+                        return true;
+                    }
+                }
+
+                const message = `Icon should be a valid image type ${JSON.stringify(allowedTypes)}`;
+
+                context.report(iconPath, message);
```
antross: Include the location in this report.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+
+                if (image !== null) {
+                    const isValidType: boolean = allowedTypes.some((item) => {
+                        return item === image.ext;
+                    });
+
+                    if (specifiedType !== image.ext) {
+                        const message = `Real image type (${image.ext}) do not match with specified type (${specifiedType})`;
+
+                        context.report(iconPath, message);
```
antross: Include the location in this report.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+         * @param rawContent raw datastream
+         * @param iconPath icon resource path
+         */
+        const validateImageType = (iconType: string | undefined, rawContent: Buffer, iconPath: string) => {
+            const allowedTypes = ['png', 'jpg'];
+
+            if (iconType === undefined) {
+                const message = `Icon type was not specifed`;
+
+                context.report(iconPath, message);
```
antross: Include the location in this report.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+
+                return null;
+            }
+            const response = networkData.response;
+
+            if (response.statusCode !== 200) {
+                const message = `Icon could not be fetched (status code: ${response.statusCode}).`;
+
+                context.report(iconPath, message);
```
antross: Include the location in this report.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+        const iconExists = async (iconPath: string) => {
+            let networkData: NetworkData;
+
+            try {
+                networkData = await context.fetchContent(iconPath);
+            } catch (e) {
+                debug(`Failed to fetch the ${iconPath} file`);
+                const message = `Icon could not be fetched (request failed).`;
+
+                context.report(iconPath, message);
```
antross: Include the location in this report.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+
+        /**
+         *
+         * @param iconSizes Sizes specified in the manifest file
+         * @param iconRawData
+         * @param iconPath full path to the icon file
+         */
+        const validateSizes = (iconSizes: string | undefined, iconRawData: Buffer, iconPath: string) => {
+            if (iconSizes === undefined) {
+                context.report(iconPath, `Sizes not specifed for icon`);
```
antross: Include the location in this report.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+                    }
+                }
+            }
+
+            debug(`Found ValidSizes: ${validSizes}`);
+
+            return validSizes;
+        };
+
+        const validate = async (parseEnd: ManifestParsed) => {
```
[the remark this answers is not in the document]
shivangg: Resolved.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            const hostnameWithProtocol = `${resourceURL.protocol}//${resourceURL.host}`;
+
+            debug(`Validating hint manifest-icon`);
+            /**
+             * console.log(`\n--------
+             * ${JSON.stringify(hostnameWithProtocol, null, '\t')}
+             *--------`);
+             */
+
+            if (icons !== undefined && icons.length > 0) {
```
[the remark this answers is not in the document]
shivangg: resolved.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+            } = parseEnd;
+            const resourceURL = new URL(resource);
+            const hostnameWithProtocol = `${resourceURL.protocol}//${resourceURL.host}`;
+
+            debug(`Validating hint manifest-icon`);
+            /**
+             * console.log(`\n--------
+             * ${JSON.stringify(hostnameWithProtocol, null, '\t')}
+             *--------`);
+             */
```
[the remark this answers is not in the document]
shivangg: Removed.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+         *
+         * @param icons array of the icons properties
+         * @param hostnameWithProtocol
+         */
+        const validateIcons = async (icons: ManifestImageResource[], hostnameWithProtocol: string) => {
+            const validSizes: string[] = [];
+
+            for (const icon of icons) {
+                const fullIconPath = getIconFullPath(icon.src, hostnameWithProtocol);
+
```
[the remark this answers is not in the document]
shivangg: Removed.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -0,0 +1,220 @@
...
+        /**
+         *
+         * @param icons array of the icons properties
+         * @param hostnameWithProtocol
+         */
+        const validateIcons = async (icons: ManifestImageResource[], hostnameWithProtocol: string) => {
+            const validSizes: string[] = [];
+
+            for (const icon of icons) {
+                const fullIconPath = getIconFullPath(icon.src, hostnameWithProtocol);
```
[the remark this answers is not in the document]
shivangg: Removed. At the time of writing, I anticipated it might be useful in more places.

packages/hint-manifest-icons/README.md
```diff
@@ -0,0 +1,232 @@
+# manifest-icons (`@hint/hint-manifest-icons`)
```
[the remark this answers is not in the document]
shivangg: `@hint/hint-manifest-icons` was the default from `create-hint` template. Shouldn't it be changed in the template also?

packages/hint-manifest-icons/src/meta.ts
shivangg: description and name added.

packages/hint-manifest-icons/src/hint.ts
shivangg: Resolved. Looks way cleaner now. :smiley:

packages/hint-manifest-icons/src/hint.ts
shivangg: Resolved. Learning TypeScript as I move forward with this codebase.

packages/hint-manifest-icons/src/hint.ts
shivangg: Added `Promise<Buffer | null>` as it returns a promise.

packages/hint-manifest-icons/src/hint.ts
shivangg: Removed.

packages/hint-manifest-icons/src/hint.ts
shivangg: Added line after if statements.

packages/hint-manifest-icons/src/hint.ts
shivangg: `ProblemLocation`s added to the reports.

packages/hint-manifest-icons/README.md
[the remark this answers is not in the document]
antross: Probably.

@molant What do you think? Should we just change this directly in the template or do you want to do something different for first-party vs. third-party hints? 

We may need to also update `create-hint` to ask for the friendly-name (e.g. "Manifest has icons") as I think it only asks for the ID and description currently.

packages/hint-manifest-icons/src/hint.ts
shivangg: @antross  Yes, it can be a space-separated list of sizes. I was thinking of reporting in case **none** of the specified sizes match the real size. Otherwise, pass. 

Or should it strictly report if a list of sizes is encountered?

antross: After double-checking the spec I think your plan of reporting only if **none** of the specified sizes match the real size is best.

Per https://w3c.github.io/manifest/#sizes-member:
> When multiple ImageResources are available, a user agent MAY use the value to decide which icon is most suitable for a display context (and ignore any that are inappropriate).

I read that as suggesting a developer may need to specify additional sizes that aren't exact matches to influence where the icon gets used, so we should allow it.

packages/hint-manifest-icons/src/hint.ts
antross: @shivangg Thanks! But it looks like you may still need to push those changes - I don't see them.

packages/hint-manifest-icons/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Why is this important?
+
+Icons are important for the user of your web app.
+
+When a user adds your site to their home screen, you can define
+a set of icons for the browser to use. These icons are used in
+places like the home screen, app launcher, task switcher, splash screen, etc.
+
+From Google's [Web App Manifest Guide][google-guide],
```
antross: Probably a good idea to include and lead with the MDN section for this as it contains more information about cross-browser support: https://developer.mozilla.org/en-US/docs/Web/Manifest#icons.

packages/hint-manifest-icons/README.md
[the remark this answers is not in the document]
molant: > Should we just change this directly in the template or do you want to do something different for first-party vs. third-party hints?

Don't think that's needed. We can have the same for both.

> We may need to also update `create-hint` to ask for the friendly-name

Right now the name will get transformed to an id. So if in the form you type `Manifest has icons` the id will be `manifest-has-icons`. I don't think it's worth asking for another input when we can easily change it in the readme. What we should do is remove the `@hint/hint-` part between parens.

packages/hint-manifest-icons/src/hint.ts
molant: We should link to this part of the spec in the code so we don't have to look it up again.

packages/hint-manifest-icons/README.md
molant: I'm not a big fan of copy pasting from other websites into the readme so we might want to rethink this section entirely before we publish it. Don't think it should be blocking. What do you think?

packages/hint-manifest-icons/src/hint.ts
shivangg: Added the check and the test case.

packages/hint-manifest-icons/src/hint.ts
shivangg: @antross  There is also `mediaType` returned from `fetchContent` which seems like a better candidate for comparison for strict comparison. 

Content-type [being a superset](https://stackoverflow.com/a/17949292/6118172) of MIME-type may contain more information besides media type (_like character encoding_).

antross: Yep, `mediaType` is what we're after. Good catch.

shivangg: Done.

packages/hint-manifest-icons/src/hint.ts
shivangg: Yes, pushing them. I thought pushing before completing the full set of proposed changes will unnecessarily startup the CI pipeline. And not resolving the comments makes it difficult to make sense of what changes still need to be handled by me.

Am I missing some best practice?

molant: Don't worry about CI, push when you have changes ready so conversation status are in sync with the code and we know when to review. If needed I can cancel scheduled builds.

packages/hint-manifest-icons/README.md
antross: Yeah, I'm okay leaving it for this PR and fine-tuning the wording, etc. in a follow-up change before publishing.

packages/hint-manifest-icons/tests/tests.ts
```diff
@@ -0,0 +1,204 @@
...
+    },
+    {
+        name: 'Inaccessible icon URL in the Web app manifest',
+        reports: [
+            {
+                message: `Icon could not be fetched (status code: 404).`,
+                position: {
+                    column: 29,
+                    line: 3
+                }
```
antross: To make test positions a bit easier to understand, we prefer the `"match"` format that provides a substring whose start should match the desired position. You can see an example of this in the tests for `hint-manifest-is-valid` (sorry I didn't realize `hint-manifest-app-name` is missing positions in its tests):

```js
        reports: [
            {
                message: `'root' should NOT have additional properties. Additional property found 'additionalProperty'.`,
                position: { match: 'additionalProperty' }
            }
        ]
```

shivangg: @antross This worked for the top level properties like `icons` but doesn't work for deeper properties like `src` or `type` **under** `icons` because the value of `position.match` is [checked with](https://github.com/webhintio/hint/blob/f934c5af80fedbdedc300597db126bbeb4df9016/packages/utils-tests-helpers/src/hint-runner.ts#L48) `indexOf` which returns `-1` for deeper checks. 

Is there a way to specify `match` for deeper properties?

shivangg: Specified the `match` without the preceding `"` like `match: 'src": "an-inaccessible-path.png"'` to make it work.
```js
{
        name: 'Inaccessible icon URL in the Web app manifest',
        reports: [
            {
                message: `Icon could not be fetched (status code: 404).`,
                position: { match: 'src": "an-inaccessible-path.png"' }
            }
        ],
        serverConfig: {
            '/': htmlWithManifestSpecified,
            '/fixtures/icon-192x192.png': generateImageData(icon192px),
            '/fixtures/icon-512x512.png': generateImageData(icon512px),
            '/site.webmanifest': {
                content: `{
                    "icons": [
                        {
                            "src": "an-inaccessible-path.png",
                            "sizes": "152x152",
                            "type": "image/png"
                        }
                    ]
                }`
            }
        }
    },
```

Is there a more readable way to specify the `match`?

packages/hint-manifest-icons/README.md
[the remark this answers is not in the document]
shivangg: Did the changes in the heading of Readme.

packages/hint-manifest-icons/README.md
shivangg: okay then, leaving it for this PR. To be continued in the follow-up PR.

packages/hint-manifest-icons/tests/tests.ts
antross: Nope. That's the best approach given the location we return is deliberately the first character of the name (just past the quote). You can include the trailing quote if you want as that won't hurt anything.

FYI The position is after the first quote to ensure we underline the full name in our VS Code extension since we currently only provide a start location. If we update in the future to provide a start and end range instead we may revisit this and include the opening quote, making the match text slightly less awkward.

molant:
> Any idea why CI didn't trigger properly on the latest push?

Looks like there are issues with AP not finding files if they come in a PR if it originates from a fork 🤔

PR #2296, opened 2019-04-22
Drop caniuse

antross:
## Pull request checklist

Make sure you:

Includes updating all hints which previously used `caniuse-api` to use the new `isSupported` helper based on MDN data. Also cuts `hint-compat-api` over to using the packaged MDN data from `@hint/utils` shaving ~30KB off the bundle size (bringing it to ~618KB).

I'll leave migrating `hint-compat-api` to the new `getUnsupported` helper for a follow-up PR (which should help simplify that hint significantly).


packages/utils/src/compat/cache.ts
```diff
@@ -0,0 +1,34 @@
+type Value = string[] | null;
+
+const cache = new Map<string, Map<string[], Value>>();
+
+const _hasCachedValue = (key: string, browsers: string[]): boolean => {
+    return cache.has(key) && cache.get(key)!.has(browsers);
+};
+
+const _getCachedValue = (key: string, browsers: string[]): Value => {
+    return cache.has(key) && cache.get(key)!.get(browsers) || null;
```
molant: Nit: `_getCachedValue` gets call always after `_hasCachedValue` so we could potentially drop `cache.has(key)` from this check.

Probably a micro optimization so I'm fine leaving as it is.

packages/utils/src/compat/css.ts
```diff
@@ -0,0 +1,172 @@
...
+    });
+};
+
+/**
+ * Determine if the provided CSS selector is supported (e.g. `:valid`).
+ *
+ * Can be passed any selector, but currently only validates pseudo-selectors
+ * as most other selector types already have broad support and certain other
+ * special cases (e.g. newer attribute and combinator selectors) need special
+ * handling to map to the MDN data (which hasn't been done yet).
```
molant: Is there an issue in MDN for this? If so maybe we should link it? Maybe create another issue in our side to know we have to do this and link to the MDN issue?

molant:
@antross there are conflicts with a `package.json`. Can you please fix it?

antross:
Yep, I'll get this updated shortly.

packages/utils/src/compat/cache.ts
antross: Yeah, I'd prefer to leave this for now.

antross:
> Also I'm thinking we should link to the video of the PR review in the commit message as a Ref (https://youtu.be/Y94fMxmbMFI). It helped me a lot. What do you think?

Yep, I like that - I'll add it.

PR #2409, opened 2019-05-08
Chore: Script to sign extensions for Firefox

molant:
## Pull request checklist

Make sure you:

The new script `sign-extension` automatically packages the extension and
uploads it for signing. The signed packaged is stored in
`web-ext-artifacts`.

In order for the script to work the following environment variables
need to exist and be valid:

* `WEB_EXT_API_KEY`
* `WEB_EXT_API_SECRET`


molant:
@antross this will generate a new entry into the account each time it's called unless we reuse the id. I think for now we should be OK but maybe we want to have an extra step that changes the name to `webhint-development` or something like that so it's easier to identify the ones that are for development and the official published ones?

packages/extension-browser/package.json
```diff
@@ -123,6 +123,7 @@
     "lint:dependencies": "node ../../scripts/lint-dependencies.js",
     "lint:md": "node ../../scripts/lint-markdown.js",
     "package-extension": "web-ext build -s ./dist/bundle -a ./dist --overwrite-dest && npm run lint-bundle-size",
+    "sign-extension": "web-ext sign -s ./dist/bundle",
```
antross: Can we have this output in the `./dist` folder just like `package-extension` does (if so the `.gitignore` update can be removed)? Also does it need the `--overwrite-dest` flag as well?
```suggestion
    "sign-extension": "web-ext sign -s ./dist/bundle -a ./dist --overwrite-dest",
```

molant: `--overwrite-dest` doesn't work in this case. Also this command doesn't require running `package-extension`

packages/extension-browser/package.json
```diff
@@ -112,7 +112,7 @@
   "repository": "webhintio/hint",
   "scripts": {
     "build": "npm-run-all build:*",
-    "build-release": "npm run clean && npm run build:assets && tsc --inlineSourceMap false --removeComments true && webpack --env.release && npm run package-extension",
+    "build-release": "npm run clean && npm run build:assets && tsc --inlineSourceMap false --removeComments true && webpack --env.release",
```
antross: If we remove `&& npm run package-extension` here, we need to add it to the end of the `test` script or we won't get bundle size validation.

PR #2518, opened 2019-05-28
Breaking: Refactor and streamline hint-compat-api

antross:
## Pull request checklist

Make sure you:

Posting as a draft due to a few remaining tasks, but feel free to review.

Tasks:
* [x] Restore support for user-provided ignore lists
* [x] Resolve multiple prefixed properties in the same block
* [x] Combine multiple versions of the same browser in reports
* [x] Pass relevant portions of the existing test suite
* [x] Trim test suite to focus on higher-level behavior (lower-level covered by helper unit tests)
* [x] Update documentation


packages/hint-compat-api/src/utils/parse-supports.ts
```diff
@@ -0,0 +1,184 @@
...
+ *         }
+ *     ]
+ * }
+ * ```
+ */
+export const parseSupports = (params: string): DeclarationGroup | null => {
+    try {
+        return parseParams(params);
+    } catch (e) {
+        return null;
```
antross: Looking at this again I'm thinking I should throw a specific type of error for known invalid states and only filter errors of that type out here. This would avoid silencing unexpected bugs is the parser...

packages/hint-compat-api/src/utils/ignore.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Determine the set of ignored features, given the provided defaults
+ * and user-specified `ignore` and/or `enable` options.
+ */
+export const resolveIgnore = (options: any, defaults: string[]) => {
+    const ignore = new Set(defaults);
+
+    if (options && options.ignore) {
+        options.ignore.forEach((entry: string) => {
```
antross: Not sure why I wrote this as a `forEach`. We've been preferring `for..of` loops lately.

packages/hint-compat-api/src/utils/ignore.ts
```diff
@@ -0,0 +1,21 @@
...
+            ignore.add(entry);
+        });
+    }
+
+    if (options && options.enable) {
+        options.enable.forEach((entry: string) => {
```
antross: `for..of`

packages/hint-compat-api/src/utils/filter-supports.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!root) {
+        return null;
+    }
+
+    try {
+        const supported = filterItem(root, hasAtSupports);
+
+        return supported.length ? supported : null;
+    } catch (e) {
+        return null;
```
antross: Similar to the parser, I should throw and catch a known error type here. Then other errors can be re-thrown to avoid silencing unexpected issues.

packages/hint-compat-api/src/utils/browsers.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const results: string[] = [];
+
+    // Sort and serialize version ranges for each browser name.
+    for (const [name, versions] of versionsByName) {
+        versions.sort((v1, v2) => {
+            return semver.compare(coerce(v1), coerce(v2));
+        });
+
+        const ranges = [];
```
antross: I'm surprised TypeScript didn't complain about an implicit `any` here...
```suggestion
        const ranges: string[] = [];
```

packages/hint-compat-api/src/css.ts
```diff
@@ -1,29 +1,227 @@
...
+    if (unsupported) {
+        context.report({ feature: `${node.selector}`, node, unsupported });
+    }
+
+    context.walk(node, context);
+};
+
+/**
+ * Filter reports for the same CSS property (prefixed or not) and only emit
+ * those for browsers that were unsupported in ALL reports (if any), favoring
```
antross:
```suggestion
 * those for browsers that were unsupported in ALL reports in the same block (if any), favoring
```

packages/hint-compat-api/docs/css.md
```diff
@@ -1,84 +1,68 @@
-# Deprecated CSS features (`compat-api/css`)
+# Supported CSS features (`compat-api/css`)
 
 ## What does the hint check?
 
-`compat-api/css` checks if the CSS features used are deprecated in the
-[targeted browsers][browser-context].
+`compat-api/css-next` checks if the CSS features used are
```
molant: Files are `css.(ts|md)`, this should be consistent.

```suggestion
`compat-api/css` checks if the CSS features used are
```

packages/hint-compat-api/docs/css.md
```diff
@@ -1,84 +1,68 @@
...
+and feature detection via `@supports` blocks.
 
 ### Examples that **trigger** the hint
 
-The [box-lines][box-lines] property
-was added with the `-webkit-` prefix for Chrome and removed from versions of
-Chrome 67 and onwards.
-Targeted Chrome browsers of versions 67 and up will trigger the hint.
+The [box-flex][box-flex] was
+never added for any version of Internet Explorer. If Internet Explorer is
```
molant: Should it be `added to`? I think there's another one later on in case it has to be changed. Also later in the HTML one you use `was never added on any version`.
 
```suggestion
never added to any version of Internet Explorer. If Internet Explorer is
```

packages/hint-compat-api/docs/css.md
```diff
@@ -1,84 +1,68 @@
...
-    }
+    display: grid;
 }
 ```
 
 ### Examples that **pass** the hint
 
-The [background][background] property was never
-removed for any browser. It should always pass the hint.
+The [charset][charset]
```
molant: I think we should add an `@supports` example in this section.

packages/hint-compat-api/docs/css.md
```diff
@@ -101,7 +95,7 @@ In the `.hintrc` file:
     "connector": {...},
     "formatters": [...],
     "hints": {
-        "compat-api/css": ["error", {
+        "compat-api/css-next": ["error", {
```
molant:
```suggestion
        "compat-api/css": ["error", {
```

packages/hint-compat-api/docs/css.md
```diff
@@ -120,7 +114,7 @@ In the `.hintrc` file:
...
```
molant:
```suggestion
        "compat-api/css": ["error", {
```

packages/hint-compat-api/docs/css.md
```diff
@@ -1,84 +1,68 @@
...
+does not have any information about when it was added to various browsers
+according to the  [MDN compat data][browser-compat]. No error is thrown
+if there is no information available on when, or if, the feature was added.
 
 ```css
 .example {
-    box-lines: single;
+    text-transform: capitalize;
 }
 ```
```
molant: We talked about listing the current limitations of the hint too.

packages/hint-compat-api/docs/html.md
```diff
@@ -1,75 +1,73 @@
-# Deprecated HTML features (`compat-api/html`)
+# Supported HTML features (`compat-api/html`)
 
 ## What does the hint check?
 
-`compat-api/html` checks if the HTML elements and attributes used are
-deprecated in the [targeted browsers][browser-context].
+`compat-api/html-next` checks if the HTML elements and attributes used are
```
molant:
```suggestion
`compat-api/html` checks if the HTML elements and attributes used are
```

packages/hint-compat-api/docs/html.md
```diff
@@ -92,7 +98,7 @@ In the `.hintrc` file:
     "connector": {...},
     "formatters": [...],
     "hints": {
-        "compat-api/html": ["error", {
+        "compat-api/html-next": ["error", {
```
molant:
```suggestion
        "compat-api/html": ["error", {
```

packages/hint-compat-api/docs/html.md
```diff
@@ -111,7 +117,7 @@ In the `.hintrc` file:
...
```
molant:
```suggestion
        "compat-api/html": ["error", {
```

packages/hint-compat-api/docs/css.md
antross: Sounds better to me - most of this section was pre-existing. I just moved things around a bit.

packages/hint-compat-api/src/css.ts
```diff
@@ -1,29 +1,227 @@
...
+
+    if (reportData) {
+        if (!reports.has(prop)) {
+            reports.set(prop, []);
+        }
+
+        // Decl had a report, add it to the list to resolve later.
+        reports.get(prop)!.push(reportData);
+    } else {
+        // Decl was fully supported; no need to test other decls in this block.
```
molant: It took me a while to understand what was going on here with this comment. Maybe something like the following?

> Decl was fully supported; no need to test other decls for same property in this block so setting to `null` to `return` at the beginning of the function.

packages/hint-compat-api/tests/html.ts
```diff
@@ -15,231 +13,80 @@ const generateHTMLConfig = (fileName: string) => {
...
+        },
+        {
+            name: 'Does not report manually ignored HTML features',
+            serverConfig: generateHTMLConfig('values')
+        }
+    ],
+    {
+        browserslist: targetBrowsers,
+        hintOptions: { enable: ['integrity'], ignore: ['input[type=color]'] },
+        parsers: ['css']
```
antross:
```suggestion
```

packages/hint-compat-api/tests/html.ts
```diff
@@ -15,231 +13,80 @@ const generateHTMLConfig = (fileName: string) => {
...
+                    message: 'input[type=color] is not supported by ie 10-11.',
+                    position: { match: 'input type="color"' }
+                }
+            ],
+            serverConfig: generateHTMLConfig('values')
+        }
+    ],
+    {
+        browserslist: targetBrowsers,
+        parsers: ['css']
```
antross:
```suggestion
```

packages/hint-compat-api/src/utils/browsers.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        // Outer loop runs once per start of each range.
+        for (let i = 0; i < versions.length; i++) {
+            const start = versions[i];
+
+            // Inner loop jumps to the end of a range.
+            while (parseInt(versions[i + 1]) - parseInt(versions[i]) <= 1) {
+                i++;
```
antross: @sarvaje You wanted to think about a different approach to doing this without modifying `i` inside the loop.

packages/hint-compat-api/src/css.ts
```diff
@@ -1,29 +1,227 @@
...
+    } else {
+        validateDeclValue(node, context);
+    }
+};
+
+const validateDeclInBlock = (node: Declaration, reports: ReportMap, context: Context): void => {
+    const prop = vendor.unprefixed(node.prop);
+
+    // If any matching decl was already fully supported, no need to evaluate further.
+    if (reports.get(prop) === null) {
```
antross: Use a constant or enum with a name like `Passed` instead of `null` for clarity.

packages/hint-compat-api/src/utils/browsers.ts
antross: @sarvaje Here's what I ended up changing it to:

```ts
for (let i = 0, start = versions[0]; i < versions.length; i++) {
    if (parseInt(versions[i + 1]) - parseInt(versions[i]) <= 1) {
        continue; // Continue until the end of a range.
    }

    // Format current range as either `start` or `start-end`.
    if (start === versions[i]) {
        ranges.push(start);
    } else {
        ranges.push(`${start}-${versions[i]}`);
    }

    // Remember the start of the next range.
    start = versions[i + 1];
}
```

sarvaje: I like this version more than the previous one hehehe.

packages/hint-compat-api/src/utils/filter-supports.ts
```diff
@@ -0,0 +1,78 @@
+import difference = require('lodash/difference');
+import intersection = require('lodash/intersection');
+import union = require('lodash/union');
+
+import { getSupported } from '@hint/utils/dist/src/compat/support';
+
+import { parseSupports, Declaration, DeclarationGroup } from './parse-supports';
+
+class InvalidInputError extends Error {
```
sarvaje: 👍

packages/hint-compat-api/src/utils/parse-supports.ts
```diff
@@ -0,0 +1,196 @@
...
+};
+
+export type DeclarationGroup = {
+    type: 'and' | 'or' | 'not';
+    nodes: (Declaration | DeclarationGroup)[];
+};
+
+type Token = '(' | ')' | 'not' | 'and' | 'or' | Declaration;
+
+class InvalidInputError extends Error {
```
sarvaje: You have the same `InvalidInputError` in the file ` packages/hint-compat-api/src/utils/filter-supports.ts`, maybe you can move this to its own file and share it.

antross: By keeping them distinct classes, if `parse-supports.ts` accidentally lets one escape, it won't get trapped by `filter-supports.ts` so we can see the issue and fix it.

I can give them unique names if we want, but I don't think we should share a single definition as each component should own it's own "ignored" error type.

antross: Alternatively I could just add comment clarifying why the error type is not shared if that works for you.

sarvaje: hehehe I saw the same name/code and I thought it was the same.
Maybe a comment for future refactors would be nice :)

PR #2515, opened 2019-05-28
Fix: Issue updating references

molant:
## Pull request checklist

Make sure you:

- [x] Signed the [Contributor License Agreement](https://cla.js.foundation/webhintio/hint)
- [x] Followed the [commit message guidelines](https://webhint.io/docs/contributor-guide/getting-started/pull-requests/#commit-messages)

For non-trivial changes, please make sure you also:

- ~~Added/Updated related documentation.~~
- ~~Added/Updated related tests.~~

PR #2520, opened 2019-05-29
Build: New incremental builds

### LINKED ISSUE #2078: [Feature] Use "--incremental" feature
TypeScript 3.4 (currently RC) has support for `--incremental` which can make build times a lot faster:

image

https://devblogs.microsoft.com/typescript/announcing-typescript-3-4-rc/

molant:
## Pull request checklist

Make sure you:

Store `tsconfig.tsbuildinfo` in a different directoy than `dist` to
avoid removing it when running `npm run clean`.

With this change running `yarn build` shows the following numbers:

* Cold build (no `.tsbuildinfo`): 786s
* Incremental build: 366s

Raw TypeScript time should be even better as `yarn build` does more
than just compilinng.

It is not necessary to add `"incremental": true` to the base
`tsconfig.json` because it is enabled by default with `composite`
configurations, but it is added here for clarity of intention.

Fix #2078

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+buildcache
 coverage
 dist
 hint-report
--- a/packages/connector-chrome/tsconfig.json
+++ b/packages/connector-chrome/tsconfig.json
@@ -1,7 +1,8 @@
 {
     "compilerOptions": {
         "outDir": "dist",
-        "strict": true
+        "strict": true,
+        "tsBuildInfoFile": "./buildcache/tsconfig.tsbuildinfo"
     },
     "exclude": [
         "dist",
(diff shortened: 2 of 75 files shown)


molant:
Keep in mind that this PR has also #2515 so we should probably merge that one first.

molant:
So it turns out that if the generated `.js` files are removed and nothing gets changed, TS will not recreate them if there's a `.tsbuildcache` file so this is actually not doing what we want.

There's an [open issue about this in the [TypeScript repo](https://github.com/microsoft/TypeScript/issues/30602). I'm going to leave this PR open until I get [clarity about my assumption](https://github.com/microsoft/TypeScript/issues/30602#issuecomment-497106142). If it turns out there's no benefit then there is no point in storing `.tsbuildcache` and what we are doing here.

molant:
Closing this as it looks like we will not be getting any perf benefit the way things are 😞

PR #2584, opened 2019-06-11
Fix: Parser javascript walk.full/fullAncestor

### LINKED ISSUE #2583: [Parser Javascript] walk.full/fullAncestor throws an exception [0.2]

sarvaje:
Fix #2583

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/parser-javascript/src/parser.ts
+++ b/packages/parser-javascript/src/parser.ts
@@ -54,6 +54,7 @@ export default class JavascriptParser extends WebhintParser<ScriptEvents> {
             const options: Options = { locations: true };
             const ast = parse(sourceCode, options) as ESTree.Node;
             const tokens = [...tokenizer(sourceCode, options)];
+            const defaultCallbacksProperty = 'callbacks';
 
             // Store a WalkArray for each method supported.
             const walkArrays: { [t: string]: WalkArray } = {};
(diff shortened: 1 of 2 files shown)


packages/parser-javascript/src/parser.ts
```diff
@@ -192,7 +193,7 @@ export default class JavascriptParser extends WebhintParser<ScriptEvents> {
                     let allVisitors: NodeVisitor | Function = {};
 
                     visitors.forEach((callbacks, name) => {
-                        if (typeof allVisitors !== 'function') {
+                        if (name !== defaultCallbacksProperty) {
```
antross: The change looks good, but I did want to share that it took me awhile to convince myself the existing code was safe and wouldn't result in us stomping on an existing `allVisitors` object in the assignment below this block. 

This stemmed from `WalkArray` being the *same* type for each variation of a walk whereas two of the methods only accept `NodeVisitor` and the other two only accept a callback function. This means we'll never deal with both types at once which makes this code safe. To be fair, that's already communicated in the comments in the file.

Still, I'm left wondering if we can improve the readability here through some better types or restructuring of the code to make this more obvious and reduce the amount of context needed. Perhaps something starting with a slightly modified `WalkArray` type.

```ts
type WalkArray = Array<[Key, Function[] | Map<string, Function[]>]>;
```

Granted this change alone requires updates through much of the rest of the code so I don't want to block the PR on it since most of this code was pre-existing. Something to keep in mind for a future refactor perhaps.

sarvaje: I have added some types.
I have some questions, so it would be nice if we can talk tomorrow :)

Issue #2415, opened 2019-05-09
[Feature] Don't allow Breaking commits to span multiple packages


antross:
To be implemented as a pre-commit hook. Possibly allowing some sort of flag in the commit message to override.

This should help us avoid accidentally bumping the major version number of dependent packages (e.g. hints) that need updated when we make a breaking change to one of their dependencies (e.g. parsers).

PR #2596, opened 2019-06-13
Fix: Correctly support `<base>` in all connectors

### LINKED ISSUE #2465: [Bug] Correctly support `<base>` in all connectors [3]
# 🐛 Bug report

We have a few issues with `<base>` not working properly. This can be verified in the `hint-sri` and `hint-no-broken-links`.

* `base` can be relative and our current code path with `new URL` does not accept it and fails. We should use the following code or something similar to fix that (currently available in https://github.com/webhintio/hint/pull/2468/commits/c3c8e2392ff62cf2c6f40eb822f785f95802773c but should be added to `no-broken-links` as well):
```ts
const createResourceURL = (resource: string) => {
    const targetURL = evt.resource;
    const pageDOM = this.context.pageDOM!;
    const baseTags = pageDOM.querySelectorAll('base');
    const hrefAttribute = (baseTags.length === 0) ? '' : (baseTags[0].getAttribute('href') || '');
    const baseURL = new URL(hrefAttribute, targetURL);

    /**
        * `new URL(path, base)` only uses `base.origin` to create
        * the new URL. E.g.:
        *
        * ```ts
        * const resourceURL = new URL('/style.css', 'https://localhost/nested/');
        *
        * resourceURL.href === 'https://localhost/style.css';
        * ```
        * What is needed is `https://localhost/nested/style.css`
        *
        * To solve this we use `path.join` with `base.pathname`
        * and we use the result to create a new `URL`.
        */
    const pathname = join(baseURL.pathname, resource);
    const resourceURL = new URL(pathname, baseURL);

    debug(`calculated "resourceURL": ${resourceURL}`);

    return resourceURL;
};
``` 
* `resourceLoader` for jsdom is not taking into account `base` correctly. Don't know if it's because we pass a `_finalHref` or what. Need to investigate a bit more. Chrome seems to be reporting the right downloaded URL.
* not sure that `getElementByUrl` is going to work correctly with `base`. We might be passing a different URL than expected. Need to double check this.

-------------------------------------------------

**OLD REPORT**

## Description

Hi 👋,
I noticed that the sri hint doesn't resolve URLs with the base href anymore in case it is used. Some time ago I submitted a pull request to fix this. But there seems to be no trace of those changes anymore in the current code.

Here is a link to one of the commits from last August:
https://github.com/webhintio/hint/commit/4817bc884ddd958cfbaa886ddc3ac6103e3d8040

## Environment

* __`hint` version:__ v5.0.1
* __`Node.js` version:__ v11.15.0
* __`OS`:__ OS X
* __URL for which webhint failed:__ https://chrisguttandin.github.io/angular-meetup-berlin-april-2019/slides/1

## `webhint` configuration

<details>
<summary>webhint’s configuration</summary>

```js
{
    "browserslist": [
        "last 2 Chrome versions",
        "last 2 Edge versions",
        "last 2 Firefox versions",
        "last 2 Opera versions",
        "last 2 Safari versions"
    ],
    "connector": {
        "name": "chrome",
        "options": {
            "waitFor": 1000
        }
    },
    "formatters": [
        "stylish"
    ],
    "hints": {
        "sri": [
            "error",
            {
                "originCriteria": "all"
            }
        ]
    },
    "hintsTimeout": 120000
}


```

</details>

## Debug output

Please include the raw output generated by `webhint`, or if possible,
the raw output from running `webhint` in debug mode: `hint --debug <URL>`.

<details>
<summary>webhint’s raw output</summary>

```text
https://chrisguttandin.github.io/web-audio-conference-2016/slides/1
line 0  col 888   Error  The hash in the "integrity" attribute in resource https://chrisguttandin.github.io/web-audio-conference-2016/slides/styles.0d342b82d21c340a7319.css doesn't match the received payload.
Expected: sha384-Ra84+SB14QcHtUcO+5znERKRYeENUMgGJadAsRAqwxAGf/istbLb/TgRhhvQXUsz
Actual:   sha384-91mdN8qjIiXZ0GEwH8MUuEbd7RCwdwesDZ0pHnF5f4OZCu7zvmBokMYjKbpxZ17X    sri
line 0  col 1065  Error  The hash in the "integrity" attribute in resource https://chrisguttandin.github.io/web-audio-conference-2016/slides/4.a0a92c5a6ca72d866ceb.js doesn't match the received payload.
Expected: sha384-Ra84+SB14QcHtUcO+5znERKRYeENUMgGJadAsRAqwxAGf/istbLb/TgRhhvQXUsz
Actual:   sha384-25mC50a1kqLhhBEcS+mw4DIBVoluiE22vMKqM2Go+auygPWdE+k2F4pZhyyFp1T5          sri
line 0  col 5244  Error  The hash in the "integrity" attribute in resource https://chrisguttandin.github.io/web-audio-conference-2016/slides/polyfills.61ddf36ababb35d368c0.js doesn't match the received payload.
Expected: sha384-Ra84+SB14QcHtUcO+5znERKRYeENUMgGJadAsRAqwxAGf/istbLb/TgRhhvQXUsz
Actual:   sha384-+N7evcl7zrc6o9kMNnhuSkeAgOTW8X1IJ9QNoUeFg1Nk2F9iePwtyeN23Xmrfvl8  sri
line 0  col 5440  Error  The hash in the "integrity" attribute in resource https://chrisguttandin.github.io/web-audio-conference-2016/slides/scripts.2e1ce7e02614f3c7a0cf.js doesn't match the received payload.
Expected: sha384-Ra84+SB14QcHtUcO+5znERKRYeENUMgGJadAsRAqwxAGf/istbLb/TgRhhvQXUsz
Actual:   sha384-yI6gqSjsIyAXA7wFcWkQ4wOL/tjbFiuFsNpCjJ8HTMnMqjIRVSzn3rALqetuHCfy    sri
line 0  col 5634  Error  The hash in the "integrity" attribute in resource https://chrisguttandin.github.io/web-audio-conference-2016/slides/main.d5fd304e58c763fb812c.js doesn't match the received payload.
Expected: sha384-Ra84+SB14QcHtUcO+5znERKRYeENUMgGJadAsRAqwxAGf/istbLb/TgRhhvQXUsz
Actual:   sha384-hoMI7obu/2kkxYGPqjXQxSxWT6qgUjNl5eks+Jla4W/FiWlpmAS/TeKQTHc7BdVa       sri
✖ Found 5 errors and 0 warnings

✖ Found a total of 5 errors and 0 warnings
```

</details>

It fails because it resolves the URLs incorrectly and downloads a different (the 404.html file) file instead.

sarvaje:
Fix #2465

## Pull request checklist

Make sure you:

Added a method `getElementFinalUrl` to return the final url for an element, taking into account if there is a base element in the dom.


antross:
@sarvaje Looks like tests for `connector-jsdom` are failing due to insufficient branch coverage with this change. My guess is you need to add a test using the `<base>` tag to resolve.

sarvaje:
@antross I'm working on that :)

packages/extension-browser/package.json
```diff
@@ -6,7 +6,7 @@
     ],
     "timeout": "1m"
   },
-  "bundleSize": 615000,
+  "bundleSize": 665000,
```
molant: 50K more? This doesn't make a lot of sense for the code that's being added. Maybe a wrong import that's pulling more stuff than needed?

sarvaje: yep, it is weird, I'm taking a look to the changes.

packages/connector-local/src/connector.ts
```diff
@@ -323,59 +330,76 @@ export default class LocalConnector implements IConnector {
...
-         *
-         * That's why we need to parse it to an URL
-         * and then get the path string.
-         */
-        const uri = getAsUri(target);
-        const filePath: string = uri ? asPathString(uri) : '';
-        const rawContent: Buffer = options && options.content ? Buffer.from(options.content) : await readFileAsBuffer(filePath);
-        const contentType = getContentTypeData(null as any, filePath, null, rawContent);
-        let content = '';
+        try {
```
sarvaje: Remove this try, I add it just to cache an error I was having.

packages/extension-browser/package.json
sarvaje: I revert this change. The change size was something unrelated with my PR

packages/connector-local/src/connector.ts
```diff
@@ -309,10 +312,14 @@ export default class LocalConnector implements IConnector {
     /* istanbul ignore next */
     private async onParseHTML(event: HTMLParse) {
         this._document = event.document;
+        this._base = getBaseUrl(this._document, event.resource);
         this._evaluate = this.createJsdom(event.html).window.eval;
 
         await traverse(this._document, this.engine, event.resource);
 
+        // Restoring base after traverse to avoid issues if there is more than one HTML file.
+        this._base = null;
```
antross: Isn't this unnecessary given we always assign `this._base` at the beginning of `onParseHTML`? If we had some sort of out-of-order issue we'd still have a bug if both HTML files had a `<base>` since `null` wouldn't be the right answer either.

packages/connector-local/src/connector.ts
```diff
@@ -365,17 +372,28 @@ export default class LocalConnector implements IConnector {
...
+
+        return href[href.length - 1] === '/' ? href : `${href}/`;
+    }
+
+    public async collect(target: URL, options?: IFetchOptions) {
         /** The target in string format */
-        const href: string = this._href = target.href;
+        const pathString = asPathString(target);
+        const file = isFile(pathString);
+        const href = this._href = this.getTargetString(target, file);
```
antross: What does this fix? And do we have a test that covers it?

packages/connector-jsdom/tests/get-element-final-url.ts
```diff
@@ -0,0 +1,100 @@
...
+        maxListeners: 0,
+        wildcard: true
+    }) as Engine;
+
+    const connector: IConnector = new JSDOMConnector(engine, {});
+    const server = await Server.create({ configuration: generateHTMLPage(`<base href="${baseUrl}">`, '') });
+
+    await connector.collect(new URL(`http://localhost:${server.port}/`));
+
+    await engine.emitAsync('element::base', {
```
antross: This should be able to be removed from the test since we now get the base URL from the DOM snapshot instead of during traversal.

packages/connector-jsdom/tests/get-element-final-url.ts
```diff
@@ -0,0 +1,100 @@
...
```
antross: Also can be removed.

packages/connector-jsdom/tests/get-element-final-url.ts
```diff
@@ -0,0 +1,100 @@
...
+    const server = await Server.create({ configuration: generateHTMLPage('', '') });
+    const siteUrl = `http://localhost:${server.port}/`;
+
+    await connector.collect(new URL(siteUrl));
+
+    await Promise.all([connector.close(), server.stop()]);
+
+    const expected = url.resolve(siteUrl, elementUrl);
+
+    const actual = connector.getElementFinalUrl(elementUrl);
```
antross: Not technically a problem, but feels wrong to access data off of `connector` after calling `connector.close()`. I'd prefer to have the `await Promise.all` cleanup at the end of the test function (for this and the other new tests).

packages/connector-puppeteer/src/connector.ts
```diff
@@ -43,6 +45,7 @@ type ConnectorOptions = {
 
 export default class PuppeteerConnector implements IConnector {
     private _auth: AuthConfig | null;
+    private _base: URL | null = null;
```
antross: What do you think about treating this more like `_finalHref` in all of the connectors and storing a `_baseHref` `string` instead of an `URL`? This would help eliminate some of the places where you're using `_base!` or falling back to `_finalHref` (which, if hit, should only occur when `_finalHref === ''` anyway).

packages/connector-puppeteer/tests/get-element-final-url.ts
```diff
@@ -0,0 +1,100 @@
...
+        maxListeners: 0,
+        wildcard: true
+    }) as Engine;
+
+    const connector: IConnector = new PuppeteerConnector(engine, {});
+    const server = await Server.create({ configuration: generateHTMLPage(`<base href="${baseUrl}">`, '') });
+
+    await connector.collect(new URL(`http://localhost:${server.port}/`));
+
+    await engine.emitAsync('element::base', {
```
antross: This can be removed.

packages/connector-puppeteer/tests/get-element-final-url.ts
```diff
@@ -0,0 +1,100 @@
...
```
antross: This can be removed.

sarvaje:
> In addition to my feedback below, what do you think about having getElementFinalUrl live as a method on HTMLDocument and/or HTMLElement instead of each Connector? That should reduce redundant code across connectors to handle this case as well as the number of interfaces that need changed.

What about storing `_base` in the `HTMLDocument`, and add to it a `resolveUrl` as you say, that will do what `getElementFinalUrl` does.

What do you think?

packages/connector-jsdom/tests/get-element-final-url.ts
sarvaje: yes, I just forgot to remove them from some tests.

packages/connector-local/src/connector.ts
sarvaje: That was necessary with the old code. Removing!

antross:
> What about storing _base in the HTMLDocument, and add to it a resolveUrl as you say, that will do what getElementFinalUrl does.

Yep, I think we're in agreement. We need the HTMLDocument anyway before this can work and putting the helper here will avoid having to add a new public API to every connector. 

To simplify further you may also want to roll in my suggestion of storing a `_baseHref: string` instead of a `_base: URL | null` from above.

sarvaje:
>To simplify further you may also want to roll in my suggestion of storing a _baseHref: string instead of a _base: URL | null from above.

Yep, but it will be in the HTMLDocument, not in the connector.

packages/connector-local/src/connector.ts
sarvaje: This was to get the right path/URL when we were analyzing a folder, instead of a file. I think this is not going to be necessary anymore.

molant:
image

This should probably be in different commits so not everything is breaking. See also https://github.com/webhintio/hint/issues/2415 😊

sarvaje:
> This should probably be in different commits so not everything is breaking. See also #2415 😊

yes, when the PR is approved I will rewrite the commits

packages/utils/src/dom/html.ts
```diff
@@ -199,4 +222,12 @@ export class HTMLDocument {
 
         return result;
     }
+
+    public resolveUrl(url: string) {
+        if (!this._base) {
```
antross: This check shouldn't be necessary anymore.

sarvaje: you are right!

sarvaje:
> One note for an if-check I don't think we need anymore, but otherwise looks good. Go ahead and re-work the commits when you're ready and tests have passed, then I'll take a final look.

Ok!

sarvaje:
The commits are reworked, waiting for the tests to pass

PR #2623, opened 2019-06-19
Fix: Don't allow Breaking commits to span multiple packages

### LINKED ISSUE #2415: [Feature] Don't allow Breaking commits to span multiple packages

sarvaje:
Fix #2415

## Pull request checklist

Make sure you:

### DIFF
--- a/scripts/check-commit-message.js
+++ b/scripts/check-commit-message.js
@@ -106,6 +106,36 @@ const checkSecoundLine = (line) => {
     return issues;
 };
 
+const checkPackagesModified = (files) => {
+    const issues = [];
+    const packages = files.reduce((pkgs, file) => {
+        const pathParts = file.split('/');
+
+        if (pathParts[0] === 'packages') {
+            pkgs.add(pathParts[1]);
+        }
+
+        return pkgs;
+    }, new Set());
+
+    if (packages.size > 1) {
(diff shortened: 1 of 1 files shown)


scripts/check-commit-message.js
```diff
@@ -106,6 +106,36 @@ const checkSecoundLine = (line) => {
...
+
+        if (pathParts[0] === 'packages') {
+            pkgs.add(pathParts[1]);
+        }
+
+        return pkgs;
+    }, new Set());
+
+    if (packages.size > 1) {
+        issues.push(`[Line 1] A breaking change cannot affect to more than one package. Packages affected:\n    - ${[...packages].join('\n    - ')}`);
```
antross: The `[Line 1]` is used to refer to the first line of the commit message. Given we're not technically validating the message itself I think we should omit that here.

```suggestion
        issues.push(`A breaking change cannot affect to more than one package. Packages affected:\n    - ${[...packages].join('\n    - ')}`);
```

sarvaje:
@antross done!

PR #2628, opened 2019-06-22
Docs/broken links

gesa:
## Pull request checklist

Make sure you:

A handful of fixes to the documentation, where either verbiage or links were incorrect.

### DIFF
--- a/packages/hint-compat-api/README.md
+++ b/packages/hint-compat-api/README.md
@@ -11,9 +11,7 @@ are compatible with the target browsers that you want to support.
 ## Hints
 
 * [compat-api/css][compat-api-css]
-* [compat-api/css-next][compat-api-css-next]
 * [compat-api/html][compat-api-html]
-* [compat-api/html-next][compat-api-html-next]
 
 ## How to use these hints?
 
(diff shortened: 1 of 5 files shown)


gesa:
! I had made a mistake with configuration package naming that I only got to the bottom of after running hint with `--debug` a few times. I think that section should be reworded, but I'm not entirely certain what it should say. This is _accurate_:

https://github.com/webhintio/hint/blob/0994c8857eef2fea3c7d89ef3ec8ca354565979f/packages/hint/docs/user-guide/concepts/configurations.md#L13-L17

but there are precisely zero examples in the docs of a .hintrc extending a non-official config, so figuring out what string to use on that line and what to name the package wasn't easy.

molant:
Hi @gesa 

You are right. We should at least have one example for extending a local configuration and a package that is not under the `@hint` namespace.

Thanks for catching this!

sarvaje:
@gesa Thanks for the PR!

> but there are precisely zero examples in the docs of a .hintrc extending a non-official config, so figuring out what string to use on that line and what to name the package wasn't easy.

Can you do this, if not, can you please open an issue so we can take a look to it later.

Thanks!!

PR #2633, opened 2019-06-24
New: Allow `application/javascript` for JS resources by default

### LINKED ISSUE #2621: Suppress warnings when application/javascript is used
Although `text/javascript` is the current best practice per discussion in #568 and [noted in the HTML standard](https://html.spec.whatwg.org/multipage/infrastructure.html#dependencies:mime-type), the practical value of switching from `application/javascript` to `text/javascript` is relatively low. 

More importantly this recommendation continues to cause confusion within the community as `application/javascript` was previously recommended by the IETF which hasn't yet updated [RFC 4329](https://tools.ietf.org/html/rfc4329) to recommend `text/javascript` (though [an active draft](https://tools.ietf.org/html/draft-ietf-dispatch-javascript-mjs) exists which will make that change).

Given the conflicting state of currently published standards, the level of confusion this causes, and the limited practical benefit of using one of these mime types over the other, I propose changing `hint-content-type` to not raise any issues when `application/javascript` is used. 

Note the recommended type would remain `text/javascript` with this change and warnings would still occur with this recommendation when any other mime-type is used (e.g. `application/x-javascript`). 

Also note that users would still be able to override the new behavior in `.hintrc` to require a single type of their choosing if desired.

antross:
Fix #2621

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint-content-type/README.md
+++ b/packages/hint-content-type/README.md
@@ -27,6 +27,20 @@ The hint checks if responses include the `Content-Type` HTTP response
 header and its value contains the appropriate media type and charset
 for the response.
 
+### A note about `application/javascript`
+
+This hint recommends using a `Content-Type` of `text/javascript` for
+JavaScript resources as [noted in the HTML standard][html js mime].
+However this hint also allows `application/javascript` because that
+value was previously recommended by the IETF in [RFC 4329][rfc 4329].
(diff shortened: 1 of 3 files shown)


molant:
Maybe `New` instead of `Fix` (although this is probably more of a breaking but...)

PR #2634, opened 2019-06-24
New: Localization

sarvaje:
## Pull request checklist

Make sure you:


scripts/create-i18n.js
```diff
@@ -0,0 +1,33 @@
...
+export type MessageName =
+${
+    Object.keys(messages).map((name) => {
+        return `    '${name}'`;
+    })
+        .join(' |\n')
+};
+
+export const getMessage = (message: MessageName, options?: GetMessageOptions) => {
+    return getMessageGlobal(message, options);
```
antross: I had another idea for allowing us to avoid specifying prefixes on every key. Since we're generating this method now, we could inject the package name here and leave it out of both the source `messages.json` file and callers of this method.

antross: That should also make `validate-i18n.js` unnecessary.

sarvaje: @antross to get the message, we also pass the language, so basically, for each report we are doing:

```js
context.report('resource', getMessage('MessageKey', {language: context.language, substitutions}, reportOptions));
```

I think it can be easy to forget to pass the language, I can think in three options:
1. Make options y language mandatory
1. Make language another parameter in `getMessage`
1. Make `context` a parameter in `getMessage`.

What do you think?

utsavized:
@antross Thanks for going through the changes on Teams ... changes look good to me at a macro-level, so I'm approving on my end. However, @sarvaje will probably have more high-level insight into the changes.

packages/parser-typescript-config/src/schema.json
```diff
@@ -89,12 +89,12 @@
...
               "description": "Enable incremental compilation.",
               "type": "boolean"
-            },            
+            },
               "tsBuildInfoFile": {
               "description": "Specify file to store incremental compilation information.",
               "type": "string"
             },
-              
+
```
sarvaje: Remove empty line.
```suggestion
```

PR #2639, opened 2019-06-26
Docs: Clarify custom configuration process

### LINKED ISSUE #2637: [Docs] Add example extending a local configuration that is not under the @hint namespace.
! I had made a mistake with configuration package naming that I only got to the bottom of after running hint with `--debug` a few times. I think that section should be reworded, but I'm not entirely certain what it should say. This is _accurate_:

https://github.com/webhintio/hint/blob/0994c8857eef2fea3c7d89ef3ec8ca354565979f/packages/hint/docs/user-guide/concepts/configurations.md#L13-L17

but there are precisely zero examples in the docs of a .hintrc extending a non-official config, so figuring out what string to use on that line and what to name the package wasn't easy.

_Originally posted by @gesa in https://github.com/webhintio/hint/pull/2628#issuecomment-504690590_

gesa:
## Pull request checklist

Make sure you:

Clarifies how to use a custom configuration. I don't know how, exactly, accurate this is, but it is the only way I was able to get mine to work. It does not answer the question of namespaced custom configs, which mine surely will be when it's actually published.

Fix #2637

### DIFF
--- a/packages/hint/docs/contributor-guide/how-to/configuration.md
+++ b/packages/hint/docs/contributor-guide/how-to/configuration.md
@@ -25,7 +25,7 @@ And in your `.hintrc` add:
 }
 ```
 
-One interesting thing that can be done with shared configurations is
+Shared configurations, using their package's `dependancies`, are able
 to download all the required `connector`s, `formatter`s, `hint`s, and
 `parser`s at the same time they get installed. This way you can guarantee
 that all dependencies are satisfied and there are no configuration issues.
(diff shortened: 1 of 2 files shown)


antross:
@gesa Thanks for fixing this!

I went digging through the code to confirm your updates and found conflicting use of [`webhint-`](https://github.com/webhintio/hint/blob/cfc6a086ff3eea00bb6cb5e3b85aa859a9d0b691/packages/hint/src/lib/utils/resource-loader.ts#L334) vs [`hint-`](https://github.com/webhintio/hint/blob/master/packages/hint/src/lib/utils/resource-loader.ts#L263) internally. I want to follow up on this with some of the other maintainers before we merge your PR in case it affects what we should document (it may require a separate bug fix of its own). 

As for
> It does not answer the question of namespaced custom configs, which mine surely will be when it's actually published.

To clarify, are you referring to publishing the package under your own [npm scope](https://docs.npmjs.com/misc/scope), much like we use `@hint/`? If so, from looking at the code it doesn't look like webhint currently handles third-party package names outside of the format `webhint-${type}-${packageName}`. Seems like something we should try to fix, though.

PR #2672, opened 2019-07-08
New: Show version in browser extension

### LINKED ISSUE #2643: Show version of browser extension in UI
From @JamoCA in #2641

> The WebHint version should be displayed in the output so it can be reported if necessary. (If I manage extensions, I see that it is v0.0.4.)

I'll open a separate issue for this as well. I could see us listing this next to the "Powered by webhint" text displayed at the bottom of the configuration and results pages.

pratiyush2706:
## Pull request checklist

Make sure you:

- [ ] ~Added/Updated related documentation.~
- [ ] ~Added/Updated related tests.~

Resolves #2643 
The webhint version used should be displayed in browser extension next to "powered by webhint" text.

### DIFF
--- a/packages/extension-browser/src/devtools/views/powered-by.tsx
+++ b/packages/extension-browser/src/devtools/views/powered-by.tsx
@@ -8,6 +8,8 @@ type Props = {
     className?: string;
 };
 
+const { version } = require('../../manifest.json');
+
 const PoweredBy = ({ className }: Props) => {
     return (
         <div className={className || ''}>
@@ -16,6 +18,8 @@ const PoweredBy = ({ className }: Props) => {
             <ExternalLink href="https://webhint.io">
                 webhint
             </ExternalLink>
+            {' '}
+            v{version}
         </div>
     );
 };


antross:
@pratiyush2706 When you're ready to have this reviewed, can you include a screenshot? Thanks!

packages/extension-browser/src/devtools/views/add-version.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import semver = require('semver');
+
+import * as pkg from '../../../package.json';
+
+console.log('Package: ', pkg.version);
+const Version = () => {
+    const versionToShow = semver.inc(pkg.version, 'patch') || pkg.version;
```
antross: FYI, you can get this from `manifest.json` - see `extension-browser/src/devtools/utils/analytics.ts` which also reads the current version.

This is preferred to using `package.json` because the whole file will end up in the bundle generated by webpack. `manifest.json` already needs to be in our output so getting the version from there avoids adding extra bytes.

packages/extension-browser/src/devtools/views/pages/config.tsx
```diff
@@ -89,6 +90,7 @@ const ConfigPage = ({ disabled, onStart }: Props) => {
             </Button>
             {' '}
             <PoweredBy className={styles.poweredBy} />
+            <Version />
```
antross: I think this may make more sense as a sub-components of `PoweredBy` so it shows everywhere we use that component.

packages/extension-browser/tsconfig.json
```diff
@@ -14,6 +15,7 @@
         "src/**/*.tsx",
         "tests/**/*.ts"
     ],
+    "files": ["package.json"],
```
antross: I don't think we need this.
```suggestion
```

packages/extension-browser/src/devtools/views/add-version.tsx
antross: Also that will avoid the need to pull in `semver` as `manifest.json` will already have been incremented appropriately as part of the build process when building for release.

packages/extension-browser/tsconfig.json
```diff
@@ -2,7 +2,8 @@
     "compilerOptions": {
         "jsx": "react",
         "outDir": "dist",
-        "strict": true
+        "strict": true,
+        "resolveJsonModule": true
```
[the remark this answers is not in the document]
pratiyush2706: @antross since I needed the ```package.json``` file it was needed because otherwise, it was not able to read the ```package.json``` file. But now if I pull it from ```manifest.json``` I think this would not be needed.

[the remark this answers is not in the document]
antross: It's really about which kind of import/require is being used. A raw require of JSON will work without this flag, but won't have any type information (which is fine for our limited use here). This flag allows an import to be used and types dynamically generated based on the contents of the JSON file.

pratiyush2706:
Screen Shot 2019-07-10 at 2 40 34 AM

pratiyush2706:
@antross  Probably the version should be 0.0.5 but when I import version from ```manifest.json``` it shows version as 0.0.4.

antross:
> @antross Probably the version should be 0.0.5 but when I import version from `manifest.json` it shows version as 0.0.4.

That doesn't sound right. What do you see when you open `manifest.json`? Perhaps you need to rebase against `master` to get the latest changes.

packages/extension-browser/src/devtools/views/powered-by.tsx
```diff
@@ -16,6 +17,7 @@ const PoweredBy = ({ className }: Props) => {
             <ExternalLink href="https://webhint.io">
                 webhint
             </ExternalLink>
+            <Version />
```
antross: Maybe this is simple enough to not bother with a separate component. If we move `const version = require('../../manifest.json').version;` to this file we should be able to replace this with the following with no additional classes or styles needed.
```suggestion
            {' '}
            v{version}
```

packages/extension-browser/tsconfig.json
```diff
@@ -2,7 +2,7 @@
     "compilerOptions": {
         "jsx": "react",
         "outDir": "dist",
-        "strict": true
+        "strict": true,
```
antross:
```suggestion
        "strict": true
```

pratiyush2706:
> > @antross Probably the version should be 0.0.5 but when I import version from `manifest.json` it shows version as 0.0.4.
> 
> That doesn't sound right. What do you see when you open `manifest.json`? Perhaps you need to rebase against `master` to get the latest changes.

@antross I did. It's still the same. My master is up-to-date. Any other possible reason for this?

antross:
> @antross I did. It's still the same. My master is up-to-date. Any other possible reason for this?

If `manifest.json` still has `0.0.4` then you don't have the latest changes. Note that your remote is likely set to your fork of  `hint` so you need to fetch from `upstream` and rebase from `upstream/master` in order to be current. 

Message me on Gitter tomorrow if you're still having trouble and I'll take a closer look.

pratiyush2706:
**Latest Preview:**

Screen Shot 2019-07-10 at 11 14 17 PM

molant:
I find a bit weird that we say webhint 0.0.5 when we are already at v6 or so?
Maybe we should clarify this is the extension version? Something like:

```
Powered by webhint
Extension version 0.0.5
```


antross:
> I find a bit weird that we say webhint 0.0.5 when we are already at v6 or so?

@molant Personally I feel we should favor simplicity here over absolute accuracy. 

We have so many versions floating around between packages I doubt even the version of `hint` itself is that meaningful (e.g. your target configuration could easily have added more hints without bumping the `hint` version). No need to bother the end user with the details.

molant:
You are lucky I'm on vacation and don't want to battle you over this right now 😛

antross:
> You are lucky I'm on vacation and don't want to battle you over this right now 😛

@molant I fully expect it to magically change one day after you get back, but for now I'll accept my minor victory. 😉

PR #2709, opened 2019-07-19
New: Add Localization to formatter-html

sarvaje:
## Pull request checklist

Make sure you:


packages/formatter-html/scripts/compile-locales.js
```diff
@@ -0,0 +1,46 @@
...
+const path = require('path');
+const util = require('util');
+
+const globby = require('globby');
+const fsExtra = require('fs-extra');
+
+const writeFileAsync = util.promisify(fs.writeFile);
+const readFileAsync = util.promisify(fs.readFile);
+
+
```
antross: Extra blank line.
```suggestion
```

packages/formatter-html/scripts/compile-locales.js
```diff
@@ -0,0 +1,46 @@
...
+const compile = async () => {
```
antross: Probably worth adding a comment explaining *why* we're compiling the locales, specifically what the generated `assets/js/scan/_locales/*/messages.js` files get used for.

packages/formatter-html/scripts/create-get-message.js
```diff
@@ -0,0 +1,43 @@
+const fs = require('fs');
+const path = require('path');
+
+const { format } = require('@hint/utils/dist/src/i18n/format');
+
+const filename = path.join(process.cwd(), 'src', 'assets', 'js', 'scan', 'get-message.js');
+
+const content = `// Autogenerated by scripts/create-get-message.js
```
antross: Same here with respect to a comment explaining why this exists. Specifically how this generated function is intended to be passed to `ejs` templates at evaluation time to resolve contained strings.

Also probably worth noting how it depends on `window.localeStrings` being defined via a `messages.js` file compiled by `scripts/compile-locales.js`.

Lastly clarify that this and the generated `message.js` files are only for supporting the online-scanner and will not be used when consuming `formatter-html` via the CLI.

packages/formatter-html/src/assets/js/scan/_locales/en/messages.js
```diff
@@ -0,0 +1,122 @@
+// Autogenerated by scripts/compile-locales.js
+/* eslint-disable quotes, quote-props */
+(function () {
```
antross: Just to clarify, do these particular generated files have to be committed (e.g. the scanner reads them from `src`)? Or can the scanner read these from the published `npm` package so we only need to generate them on build (and can omit them from `git`)?

packages/formatter-html/src/assets/js/scan/get-message.js
```diff
@@ -0,0 +1,34 @@
+// Autogenerated by scripts/create-get-message.js
+
+(function () {
```
antross: Same question here as to whether we need to commit this generated file.

packages/formatter-html/src/views/partials/scan-category-summary.ejs
```diff
@@ -2,15 +2,15 @@
     <div class="rule-tile">
         <img src="<%= category.image %>" class="rule-icon" alt="" role="presentation">
         <div class="rule-tile__info">
-            <p class="rule-tile__category"><%= category.name.toUpperCase() %></p>
+            <p class="rule-tile__category"><%= category.localizedName.toLocaleUpperCase(); %></p>
```
antross: I think we should tackle part of webhintio/webhint.io#657 here and drop converting the name to upper case. As that issue notes, this doesn't always work correctly for localized strings and we should just used the sentence-cased category name instead.

packages/formatter-html/scripts/create-get-message.js
sarvaje: I will add the comment but, this files are going to be used also for the formatter-html when it is used via the CLI. Only for a couple of strings, but they are also necessary.

packages/formatter-html/src/assets/js/scan/_locales/en/messages.js
sarvaje: For now, we need these generated files to be commited because we are using them when generating the website. These files will not be necessary once https://github.com/webhintio/webhint.io/issues/757 is done.

There is an issue to track this: #2722.

packages/formatter-html/src/assets/js/scan/get-message.js
sarvaje: Same response as before :)

packages/formatter-html/scripts/create-get-message.js
antross: Ok, good clarification.

PR #2734, opened 2019-07-25
New: Configuration All

sarvaje:
## Pull request checklist

Make sure you:


packages/configuration-all/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+```json
+{
+    "extends": ["all"],
+    "formatters": ["codeframe"]
+}
+```
+
+Note: Not all the hints can be use for all connectors,
+you will be warning when a hint is going to be ignored
```
antross:
```suggestion
you will be warned when a hint is going to be ignored
```

release/tasks/update-configuration-all.ts
```diff
@@ -0,0 +1,100 @@
...
+            if (isHint(pkg)) {
+                index.hints = Object.assign(index.hints, await getHints(pkg));
+            }
+        }
+
+        pkgConfigAll.updated = true;
+        pkgConfigAll.content.version = calculatePackageNewVersion(pkgConfigAll, currentBump);
+
+        // Step 6: Save `index.json`.
+        await fs.writeFile(indexPath, JSON.stringify(index, null, 4), 'utf-8');
```
antross: Probably should include a newline at the end of the generated file for consistency with our other source files.
```suggestion
        await fs.writeFile(indexPath, `${JSON.stringify(index, null, 4)}\n`, 'utf-8');
```

release/tasks/update-configuration-all.ts
```diff
@@ -0,0 +1,100 @@
...
+            currentBump = Math.max(currentBump, pkg.bump || Bump.none);
+            configAll.dependencies[pkg.name] = `^${pkg.content.version}`;
+
+            if (isHint(pkg)) {
+                index.hints = Object.assign(index.hints, await getHints(pkg));
+            }
+        }
+
+        pkgConfigAll.updated = true;
+        pkgConfigAll.content.version = calculatePackageNewVersion(pkgConfigAll, currentBump);
```
antross: Shouldn't this happen automatically elsewhere in the existing release script? I feel we're duplicating some work vs. what happens when updating packages for the other configurations... 

Perhaps if you run this before `calculateNewVersions` it won't be necessary? It should be ok if you're actually writing in the old versions of the dependency packages - they'll get updated by `calculateNewVersions`. 

Then you can drop the refactor that moved `calculatePackageNewVersion` around and the added `pkg.bump` property.

release/tasks/update-configuration-all.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Bumps the version of each package:
+ *
+ * 1. Load current `configuration-all` package.
+ * 2. Load current `index.json`.
+ * 3. Clean current dependencies in `package.json`.
+ * 4. Clean current hints in `index.json`.
+ * 5. Add packages to `package.json` and rules to `index.json`.
+ * 6. Save `index.json`.
+ */
+export const updateConfigurationAll = (ignoredPackages: string[]) => {
```
antross: Doesn't look like `ignoredPackages` is used anywhere in the implementation. Was it meant to be used somewhere?

release/tasks/update-configuration-all.ts
sarvaje: How do you know if it the update is 'major', 'minor' or 'patch'?

release/tasks/update-configuration-all.ts
sarvaje: good catch!

release/tasks/update-configuration-all.ts
molant: I'm answering bad on the comments, haven't had the time to check the code from my phone 🙄

I think the epic with the release script had the reasoning for versions for configs. IIRC we assumed they were going to be patches for pkg update, minor for new packages, and major for breaking (hintrc format change done manually). Maybe this script can add a commit if a pkg is missing with a "New:" and let the release do the rest?

sarvaje: sorry, I forgot to push the changes hehehe. I talked with @antross yesterday and I modified the order of the task, so this line doesn't exists anymore :) and it is using the current code to generate the version.

packages/configuration-all/index.json
```diff
@@ -0,0 +1,78 @@
...
+        "webpack-config/config-exists": "error",
+        "webpack-config/is-installed": "error",
+        "webpack-config/is-valid": "error",
+        "webpack-config/module-esnext-typescript": "error",
+        "webpack-config/modules-false-babel": "error",
+        "webpack-config/no-devtool-in-prod": "error",
+        "x-content-type-options": "error"
+    },
+    "hintsTimeout": 120000,
+    "parsers": []
```
antross: I missed this before, but I think we need to include all the parsers here if we want this configuration to fully enable all the hints. Otherwise those that are waiting for a parse event from a specific parser will never run.

sarvaje: ups, you are right!

packages/configuration-all/README.md
```diff
@@ -0,0 +1,70 @@
...
+etc.) you can do something like:
+
+```json
+{
+    "extends": ["all"],
+    "formatters": ["codeframe"]
+}
+```
+
+Note: Not all the hints can be use for all connectors,
```
utsavized:
```suggestion
Note: Not all the hints can be used for all connectors,
```

packages/configuration-all/package.json
```diff
@@ -0,0 +1,84 @@
...
+    "@hint/parser-babel-config": "^2.1.3",
+    "@hint/parser-css": "^3.0.3",
+    "@hint/parser-html": "^3.0.2",
+    "@hint/parser-javascript": "^3.0.3",
+    "@hint/parser-manifest": "^2.2.1",
+    "@hint/parser-package-json": "^1.1.3",
+    "@hint/parser-typescript-config": "^2.3.0",
+    "@hint/parser-webpack-config": "^2.1.3"
+  },
+  "description": "webhint's all configuration that includes all the published packages",
```
utsavized:
```suggestion
  "description": "webhint's configuration that includes all the published packages",
```

release/tasks/update-configuration-all.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (isHint(pkg)) {
+            const hints = await getHints(pkg);
+
+            index.hints = { ...index.hints, ...hints };
+        }
+    }
+
+    // Step 6: Save `index.json`.
+    await fs.writeFile(indexPath, `${JSON.stringify(index, null, 4)}\n`, 'utf-8');
+};
```
utsavized: Do we need any exception handling in this file for a bit more logging info, then re-throw? Or, do we generally expect things to be handled wherever this is being used?

sarvaje: I think the script will handle the exception. There are more places in the script that writes files that doesn't have a try/catch.

packages/configuration-all/index.json
antross: Also, random thought about future behavior: should we even specify parsers via `.hintrc` anymore? It just seems like a way to have hints not work by forgetting them. I'd prefer having hints specify which parser(s) they need and having those automatically pulled in based on which hints are enabled.

@sarvaje If that sounds reasonable to you, I'll file this as an issue so we can discuss with the broader team later. What do you think?

sarvaje: Sounds reasonable, but for now, we need to fill that list :)

PR #2745, opened 2019-07-27
Fix: Use `language` from `CLI Options` if present

jaspreet57:
## Pull request checklist

Make sure you:

- [ ] ~~Added/Updated related documentation.~~
- [ ] ~~Added/Updated related tests.~~

Earlier, `language` of Analyzer was being taken from config file or OS Locale.
Now it is using CLI Options also, if `language` is present in them.

## Details
This bug was caught when `packages/hint` test case failed. 
In below screenshot, test case is expecting `en-US` as language of Analyzer. But it happens to be `en-GB`. This is because current logic of loading config in Analyzer is taking language either from config file or OS Locale. This makes the test case execution OS dependent as config file is always null here.

actions-language-bug

Actually Analyzer should also consider CLI Options being passed. In test cases, we are already passing `language = 'en-US'` as actions (CLIOptions).

### DIFF
--- a/packages/hint/src/lib/cli/analyze.ts
+++ b/packages/hint/src/lib/cli/analyze.ts
@@ -189,7 +189,13 @@ const loadUserConfig = async (actions?: CLIOptions): Promise<UserConfig> => {
     }
 
     // If language is not in the config file or CLI options, get the language configure in the OS.
-    userConfig.language = (actions && actions.language) || userConfig.language || await osLocale();
+    if (actions && actions.language) {
+        userConfig.language = actions.language;
+        debug(`Using language option provided from command line: ${actions.language}`);
+    } else {
+        userConfig.language = userConfig.language || await osLocale();
+    }
+
     userConfig = mergeEnvWithOptions(userConfig) as UserConfig;
 
     return userConfig;
@@ -281,7 +287,6 @@ const actionsToOptions = (actions: CLIOptions): CreateAnalyzerOptions => {
     const options: CreateAnalyzerOptions = {
         formatters: actions.formatters ? actions.formatters.split(',') : undefined,
         hints: actions.hints ? actions.hints.split(',') : undefined,
-        language: actions.language,
         watch: actions.watch
     };
 
--- a/packages/hint/src/lib/cli/analyze.ts
+++ b/packages/hint/src/lib/cli/analyze.ts
@@ -188,8 +188,8 @@ const loadUserConfig = async (actions?: CLIOptions): Promise<UserConfig> => {
         userConfig = getDefaultConfiguration();
     }
 
-    // If language is not in the config file, get the language configure in the OS.
-    userConfig.language = userConfig.language || await osLocale();
+    // If language is not in the config file or CLI options, get the language configure in the OS.
+    userConfig.language = (actions && actions.language) || userConfig.language || await osLocale();
     userConfig = mergeEnvWithOptions(userConfig) as UserConfig;
 
     return userConfig;
--- a/packages/hint/src/lib/cli/analyze.ts
+++ b/packages/hint/src/lib/cli/analyze.ts
@@ -181,6 +181,27 @@ const askUserToInstallDependencies = async (resources: HintResources): Promise<b
     return answer;
 };
 
+/** Get language. If language is not in the config file or CLI options, get the language configure in the OS. */
+const getLanguage = async (userConfig?: UserConfig, actions?: CLIOptions): Promise<string> => {
+    if (actions && actions.language) {
+        debug(`Using language option provided from command line: ${actions.language}`);
+
+        return actions.language;
+    }
+
+    if (userConfig && userConfig.language) {
+        debug(`Using language option provided in user config file: ${userConfig.language}`);
+
+        return userConfig.language;
+    }
+
+    const osLanguage = await osLocale();
+
+    debug(`Using language option configured in the OS: ${osLanguage}`);
+
+    return osLanguage;
+};
+
 const loadUserConfig = async (actions?: CLIOptions): Promise<UserConfig> => {
     let userConfig = getUserConfig(actions && actions.config);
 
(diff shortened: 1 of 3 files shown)


jaspreet57:
@antross I have executed all other test cases to check for similar issue. This was the only one.
And as this is my first PR, please let me know if I am not following any guideline in commit message or PR template. I will fix it. Thanks

sarvaje:
@antross now that I'm thinking, let me double check if this change is necessary. I was changing the language using the CLI options when I was doing this.

antross:
@sarvaje Interesting, but for what it's worth the location we were passing in mocked CLI options for the test suite definitely wasn't getting picked up. Maybe something regressed before your original PR merged?

sarvaje:
The language from the CLI it is used for the scan, but for some reason, I did over-complicate how it is used.

@jaspreet57 can you clean the code for how it is the language used right now?

Remove language from the type: `CreateAnalyzerOptions`, and then follow the errors xDD

jaspreet57:
@sarvaje ok I will try it now. I will let you know soon when I found something.

sarvaje:
> @sarvaje Interesting, but for what it's worth the location we were passing in mocked CLI options for the test suite definitely wasn't getting picked up. Maybe something regressed before your original PR merged?

Maybe it depends on what machine is running the tests.

sarvaje:
> @sarvaje ok I will try it now. I will let you know soon when I found something.

Let me know if you have any problem

jaspreet57:
Well @sarvaje you are right. `hint` picks up the language from CLI, if passed.
So flow is (inside `analyze.ts`):
1. Load user config from file or get default one (`loadUserConfig`)
2. `createAnalyzerOptions` from give actions. (which also have `language`)
3. Pass above two to `getAnalyzer`.
4. And then in `Analyzer.create` method override `userConfig` with `AnalyzerOptions`. ( This is where Analyzer gets langauage from CLI.)

This just becomes bit complicated because logic of overriding loaded `userConfig` with `actions` went deeper inside the `Analyzer.create` method.

And in test cases, we have mocked the `Analyzer` itself with `FakeAnalyzer` whose `create` method doesn't do this overriding.

Hack could be to update `FakeAnalyzer`, but solution would be to clean this process.

**Also** there is another issue `loadUserConfig` needs `filePath` only but we are passing full `actions` to it. For first time user, this one looks like we are using `actions` inside loadUserConfig also and that leads to this PR. Thanks for pointing out early.

sarvaje:
@jaspreet57 maybe the name of `loadUserConfig` is a little confusing, but it doesn't just load the config from a path. It is the one in charge to get the final "UserConfig". That is why it gets the `actions` parameter instead of just a path.

jaspreet57:
@sarvaje marking it WIP.
I would need some suggestions. So changes would be to take logic out of `Analyzer.create` method? code like below one (creating configuration)
create-config-analyzer

And then `getAnalyzer` will accept the created `configuration: Configuration` only.
suggestions?

jaspreet57:
But I see `Analyzer.create` is being used in other packages as `hintModule.createAnalyzer`. This would be big change.
Thinking of some other solution.

antross:
If I understood @sarvaje correctly, I think he was just suggesting to largely keep your current change as-is and remove the other pre-existing code that was trying to apply the `language` from the CLI arguments in favor of just letting your new code provide the override.

jaspreet57:
> If I understood @sarvaje correctly, I think he was just suggesting to largely keep your current change as-is and remove the other pre-existing code that was trying to apply the `language` from the CLI arguments in favor of just letting your new code provide the override.

Oh ok. I understood it incorrectly. Yeah I can clean up where language is getting overridden at other places.

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -188,8 +188,14 @@ const loadUserConfig = async (actions?: CLIOptions): Promise<UserConfig> => {
         userConfig = getDefaultConfiguration();
     }
 
-    // If language is not in the config file, get the language configure in the OS.
-    userConfig.language = userConfig.language || await osLocale();
+    // If language is not in the config file or CLI options, get the language configure in the OS.
+    if (actions && actions.language) {
```
jaspreet57: Added this `if-else` to keep debug log entry, we were already having at other places from where `language` overriding code is removed.

jaspreet57:
@antross @sarvaje I just pushed changes as requested. Please review and let me know.
Test cases of `hint` and `extension-vscode` passed where this method was being used directly.
If any other package test case failed in CI, I will check it.
Thanks

packages/hint/src/lib/cli/analyze.ts
sarvaje: I like this, maybe we should have the same for the other two options. Also, maybe you can move the logic to their own `getLanguage` method.

jaspreet57: By other two options you `formatters` and `hints` right?

jaspreet57: I think if we move `updateConfigWithOptionsValues` to this place. It should work. All `debug` and all CLI options overriding logic will remain in one function.

sarvaje: Sorry, I didn't explain myself very well, for other two options I meant add a debug message when we are using the language from the userconfig or the language using `osLocale`.

jaspreet57: Oh ok. Got it. Thanks

jaspreet57: @sarvaje made the change :)

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -188,8 +188,19 @@ const loadUserConfig = async (actions?: CLIOptions): Promise<UserConfig> => {
...
```
sarvaje: Can you move all of this to a method `getLanguage`?

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -188,8 +188,19 @@ const loadUserConfig = async (actions?: CLIOptions): Promise<UserConfig> => {
...
+        userConfig.language = actions.language;
+        debug(`Using language option provided from command line: ${actions.language}`);
+    } else if (userConfig.language) {
+        debug(`Using language option provided in user config file: ${userConfig.language}`);
+    } else {
+        const osLanguage = await osLocale();
```
sarvaje: You don't need to assign the value to a new variable, you can do:
```js
userConfig.language = await osLocale();
debug(`Using language option configured in the OS: ${userConfig.language}`);
```

packages/hint/src/lib/cli/analyze.ts
jaspreet57: yeah moving it. Forgot in previous commit.

jaspreet57: @sarvaje move language logic to `getLanguage`. Please have a look. `userConfig` is made optional in case we use `getLanguage` function somewhere else in file. So it will give `osLocale` in case both `userConfig` and `actions` are `undefined`

PR #2781, opened 2019-08-02
Fix: False positives for React.createElement

### LINKED ISSUE #2706: [Bug] hint-create-element-svg false positives for React.createElement
Run a scan on a page using React to create SVG elements.
(e.g. https://codepen.io/acusti/pen/qNWMGL)

Expected:
No reports from `hint-create-element-svg`.

Actual:
The hint raises a warning:
image

-----

I see two possible fixes here:
1. Require the left-hand side to be `document`.

   This has the disadvantage of missing cases where a local variable was used to reference the document (e.g. as `doc` or `d`).

2. Require that only one argument was provided to the function.

   Technically a second argument is allowed in browsers that support custom elements, but this is rare, especially with SVG elements so the missed cases should be few.

   In theory this could still false positive in some `React.createElement` cases if only one argument is provided (since the second argument is technically optional). However current transpilation from JSX seems to always include a `null` if no attributes are provided and most SVG elements are likely to include attributes when used in JSX anyway. 

Personally I'm leaning toward option 2.

sarvaje:
Fix #2706

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint-create-element-svg/src/hint.ts
+++ b/packages/hint-create-element-svg/src/hint.ts
@@ -42,7 +42,11 @@ export default class CreateElementSvgHint implements IHint {
 
                     const arg = node.arguments[0];
 
-                    if (arg && 'value' in arg && typeof arg.value === 'string' && svgElements.has(arg.value.toLowerCase())) {
+                    /*
+                     * We are checking also the number of arguments to avoid some false positives.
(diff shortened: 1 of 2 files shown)


antross:
@utsavized Can you take a look at this? Thanks!

Issue #2657, opened 2019-07-02
[Bug] Keep create hint packages up-to-date


pratiyush2706:
## Description
On running ```npm create hint``` it installs deprecated version or outdated version of the dependencies required in the hint folder.

antross:
Thanks @pratiyush2706. I'd have to double check the `create-hint` package, but I suspect there's an opportunity here to automatically increment these as part of our release process.

pratiyush2706:
@antross Yes, automatically incrementing these would be good as developers tend to overlook the versions of the dependencies if the ```package.json``` file is created automatically.

PR #2783, opened 2019-08-05
Fix: Load latest versions from package in `create-hint` templates

jaspreet57:
## Pull request checklist

Make sure you:

- ~~[ ] Added/Updated related tests.~~

This PR improves the process of creating new hint in main repo in following two ways:

1. It loads package versions from `create-hint/package.json` while creating new hint.
As package versions in `package.json` will always remain updated, it seems good idea to load versions from it instead of updating manually in template file.
Thoughts?

2. ~~`webpack` is added to create-hint build script. When user run `npm create hint` in `hint/packages` of main repo, it doesn't work. It requires to run `webpack` after `yarn && yarn build` to create executable `create-hint` binary.~~
This is removed as test cases are failing if we add webpack step in build script.

Fix: #2657

### DIFF
--- a/packages/create-hint/package.json
+++ b/packages/create-hint/package.json
@@ -12,6 +12,8 @@
   "dependencies": {},
   "description": "webhint's hint initializer package",
   "devDependencies": {
+    "@hint/utils-tests-helpers": "^5.0.3",
+    "@types/debug": "^4.1.4",
     "@typescript-eslint/eslint-plugin": "^1.13.0",
     "@typescript-eslint/parser": "^1.12.0",
     "@types/mkdirp": "^0.5.2",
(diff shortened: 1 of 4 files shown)


jaspreet57:
@antross I have not added `i18n` support in create-hint template in this PR. It requires a bit more work so will add another PR.
`i18n` requires a script `create-i18n` from root `hint/scripts`. I doubt if it gets included in release of create-hint or not.

antross:
> i18n requires a script create-i18n from root hint/scripts. I doubt if it gets included in release of create-hint or not.

Also, when making this update `i18n` support (and the related script) should be conditional to when we're using `create-hint` to create a first party package. We don't want to require third party packages to use our localization approach. This should also mean `create-hint` doesn't need to bundle that script as it will already be in the repo for first party hints.

jaspreet57:
> > i18n requires a script create-i18n from root hint/scripts. I doubt if it gets included in release of create-hint or not.
> 
> Also, when making this update `i18n` support (and the related script) should be conditional to when we're using `create-hint` to create a first party package. We don't want to require third party packages to use our localization approach. This should also mean `create-hint` doesn't need to bundle that script as it will already be in the repo for first party hints.

Thanks. I got it.

PR #2797, opened 2019-08-07
Chore: Move `load-resource` and its dependencies to `utils`

sarvaje:
## Pull request checklist

Make sure you:


packages/utils/src/packages/has-multiple-resources.ts
```diff
@@ -0,0 +1,15 @@
+import { ResourceType } from '../types/resource-type';
+
+/**
+ * Check if it is a package with multiple resources.
+ */
+export const hasMultipleResources = (resource: any, type: ResourceType) => {
+    switch (type) {
```
utsavized: Why use `switch` if there is only one case?

packages/hint/src/lib/enums/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from '@hint/utils/dist/src/types/category';
 export * from './error-status';
 export * from './hint-scope';
-export * from './resource-type';
+export * from '@hint/utils/dist/src/types/resource-type';
+export * from '@hint/utils/dist/src/types/resource-error-status';
```
antross: This is unnecessary as `./error-status` already re-exports these types.
```suggestion
```

packages/hint/src/lib/enums/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from '@hint/utils/dist/src/types/category';
 export * from './error-status';
 export * from './hint-scope';
-export * from './resource-type';
+export * from '@hint/utils/dist/src/types/resource-type';
```
antross: Similarly, shouldn't this still just export from `./resource-type` as it already re-exports the necessary types?
```suggestion
export * from './resource-type';
```

packages/hint/src/lib/cli/version.ts
```diff
@@ -1,6 +1,6 @@
-import { logger } from '@hint/utils';
-import { loadHintPackage } from '../utils/packages/load-hint-package';
+import { logger, packages } from '@hint/utils';
 
+const { loadHintPackage } = packages;
```
antross: Do any of the imports in this file need to change since `../utils/packages/load-hint-package` still re-exports `loadHintPackage`?

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -17,7 +17,6 @@ import {
     HintResources,
     UserConfig
 } from '../types';
-import { loadHintPackage } from '../utils/packages/load-hint-package';
```
antross: Same here, seems we could still use the old references. No need for the added diff.

packages/utils/src/config/to-absolute-path.ts
```diff
@@ -0,0 +1,80 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+import { UserConfig, HintsConfigObject } from '../types/config';
+
+export const toAbsolutePaths = (config: UserConfig | null, configRoot: string): UserConfig | null => {
```
antross: Given the method is `toAbsolutePaths`, let's call this file `to-absolute-paths.ts` to include the "s" as well for consistency.

packages/utils/src/packages/has-multiple-resources.ts
sarvaje: Good question, IIRC we use to have more options hehehe.

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -6,7 +6,7 @@ import * as isCI from 'is-ci';
 import { default as ora } from 'ora';
 import * as osLocale from 'os-locale';
 
-import { appInsights, configStore, debug as d, fs, logger, misc, network, npm, ConnectorConfig, normalizeHints, HintsConfigObject, HintSeverity } from '@hint/utils';
+import { appInsights, configStore, debug as d, fs, logger, misc, network, npm, ConnectorConfig, normalizeHints, HintsConfigObject, HintSeverity, packages } from '@hint/utils';
```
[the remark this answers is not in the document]
sarvaje: That is not part of this PR, and change it will be a breaking change.

packages/utils/src/packages/load-resource.ts
```diff
@@ -0,0 +1,303 @@
...
+import * as path from 'path';
+
+import * as globby from 'globby';
+import * as semver from 'semver';
+
+import { cwd, loadJSONFile } from '../fs';
+import { debug as d } from '../debug';
+import { isFullPackageName } from './is-full-package-name';
+import { ResourceType } from '../types/resource-type';
+import { loadPackage as getPackage } from './load-package';
```
antross: Since we renamed the conflicting `loadPackage` to `requirePackage`, let's just drop this alias to `getPackage` and use the original helper name.

packages/create-hint/tsconfig.json
```diff
@@ -14,6 +14,7 @@
     ],
     "references": [
         { "path": "../hint" },
-        { "path": "../utils" }
+        { "path": "../utils" },
+        { "path": "../utils-tests-helpers" }
```
antross: This probably came from re-running one of our scripts, but since it's cross-package let's either drop it from this PR or include it as a separate commit for merging.

sarvaje: ok

PR #2802, opened 2019-08-09
Feat: Added localization support to create-hint package.

jaspreet57:
## Pull request checklist

Make sure you:

Added localization support to `create-hint` package.
Now if new hint is create in main repo (is official), it will have all necessary `i18n` scripts and files.

* `meta` will have `getDescription` and `getName` method
* `_locale` folder will get created with hint `description` and `name` for `en` language.

### DIFF
--- a/packages/create-hint/src/templates/locales-messages.json.hbs
+++ b/packages/create-hint/src/templates/locales-messages.json.hbs
@@ -1,0 +1,16 @@
+{
+    {{#each hints}}
+    "{{#if ../isMulti}}{{toCamelCase normalizedName}}_{{/if}}description": {
+        "description": "Metadata description",
+        "message": "{{description}}"
+    },
+    "{{#if ../isMulti}}{{toCamelCase normalizedName}}_{{/if}}name": {
+        "description": "Metadata name",
+        "message": "{{normalizedName}}"
+    },
+    {{/each}}
+    "reportMessage": {
+        "description": "Metadata reportMessage",
+        "message": "Add localized error message here."
+    }
+}
--- a/packages/create-hint/src/templates/meta.ts.hbs
+++ b/packages/create-hint/src/templates/meta.ts.hbs
@@ -1,13 +1,35 @@
 import { Category } from 'hint/dist/src/lib/enums/category';
 import { HintScope } from 'hint/dist/src/lib/enums/hint-scope';
 import { HintMetadata } from 'hint/dist/src/lib/types';
+{{#if packageData.official}}
+
+import { getMessage } from './i18n.import';{{/if}}
 
 const meta: HintMetadata = {
     docs: {
-        category: Category.{{category}},
-        description: `{{description}}`
+        category: Category.{{hint.category}},{{#if packageData.official}}
+        description: getMessage('{{#if packageData.isMulti}}{{toCamelCase hint.normalizedName}}_{{/if}}description', 'en'),
+        name: getMessage('{{#if packageData.isMulti}}{{toCamelCase hint.normalizedName}}_{{/if}}name', 'en'){{else}}
+        description: `{{hint.description}}`,
+        name: `{{hint.normalizedName}}`{{/if}}
+    },{{#if packageData.official}}
+    /* istanbul ignore next */
+    getDescription(language: string) {
+        return getMessage('{{#if packageData.isMulti}}{{toCamelCase hint.normalizedName}}_{{/if}}description', language);
     },
-    id: '{{#if parentName}}{{parentName}}/{{/if}}{{normalizedName}}',
+    /* istanbul ignore next */
+    getName(language: string) {
+        return getMessage('{{#if packageData.isMulti}}{{toCamelCase hint.normalizedName}}_{{/if}}name', language);
+    },{{else}}
+    /* istanbul ignore next */
+    getDescription(language: string) {
+        return `{{hint.description}}`;
+    },
+    /* istanbul ignore next */
+    getName(language: string) {
+        return `{{hint.normalizedName}}`;
+    },{{/if}}
+    id: '{{#if hint.parentName}}{{hint.parentName}}/{{/if}}{{hint.normalizedName}}',
     schema: [
         /*
          * If you want to allow the user to configure your hint
@@ -15,7 +37,7 @@ const meta: HintMetadata = {
          * https://webhint.io/docs/contributor-guide/hints/#themetaproperty
          */
     ],
-    scope: HintScope.{{scope}}
+    scope: HintScope.{{hint.scope}}
 };
 
 export default meta;
--- a/packages/create-hint/src/templates/partial-event-code.hbs
+++ b/packages/create-hint/src/templates/partial-event-code.hbs
@@ -1,6 +1,6 @@
 const { resource } = {{event}};
 
-debug(`Validating hint {{name}}`);
+debug(`Validating hint {{hint.name}}`);
 
 /*
  * This is where all the magic happens. Any errors found should be
@@ -12,5 +12,9 @@ debug(`Validating hint {{name}}`);
  */
 
 if (Math.ceil(Math.random()) === 0) {
+    {{#if packageData.official}}
+    context.report(resource, getMessage('reportMessage', context.language));
+    {{else}}
     context.report(resource, 'Add error message here.');
+    {{/if}}
 }
--- a/packages/hint/docs/contributor-guide/guides/create-custom-hint.md
+++ b/packages/hint/docs/contributor-guide/guides/create-custom-hint.md
@@ -24,9 +24,16 @@ a resource file.
 The recommended way to create a new hint is to use `npm create hint` from
 the CLI.
 
-**Note for core hints**: If you are creating a hint for the main repo you should
-use the same command from the `packages` folder. The generated files will be
-slighthly different (references and task will change a bit).
+>**Note for core hints**: If you are creating a hint for the main repo you should
+>follow the below steps:
+>
+>* Make sure to run `yarn` and `yarn build` in root directory.
+>* Then go to `create-hint` package ( `cd packages/create-hint` ).
+>* Make executable release. run `yarn webpack`.
+>* Now run `npm create hint` from `packages` folder.
+>
+>The generated files will be
+>slightly different (references and task will change a bit).
 
 At this point, a wizard will ask a series of questions relevant to the new hint,
 including the name, description, category, etc. In particular, you will be asked
--- a/packages/create-hint/src/create-hint.ts
+++ b/packages/create-hint/src/create-hint.ts
@@ -385,7 +385,11 @@ const generateHintFiles = async (destination: string, data: HintPackage) => {
     }
 
     for (const hint of data.hints) {
-        const [hintContent, testContent, metaContent] = await Promise.all([compileTemplate(hintFile.path, hint), compileTemplate(testFile.path, hint), compileTemplate(metaFile.path, hint)]);
+        const [hintContent, testContent, metaContent] = await Promise.all([
+            compileTemplate(hintFile.path, { hint, packageData: data }),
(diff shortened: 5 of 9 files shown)


packages/hint/docs/contributor-guide/guides/create-custom-hint.md
```diff
@@ -24,9 +24,15 @@ a resource file.
...
-slighthly different (references and task will change a bit).
+>**Note for core hints**: If you are creating a hint for the main repo you should
+>follow the below steps:
+>
+>* Make sure to run `yarn` and `yarn build` in root directory.
+>* Then go to `create-hint` package ( `cd packages/create-hint` ).
+>* Make executable release. run `yarn webpack`.
+>* Now run `npm create hint` from `packages` folder.
+>
+>The generated files will be slighthly different (references and task will change a bit).
```
antross: The length of this line is causing the PR to fail during markdown linting in CI. Just need to wrap it to reduce the length below 80 characters.

jaspreet57: Oh ok. Thanks. I am fixing it.

packages/create-hint/src/templates/meta.ts.hbs
```diff
@@ -1,21 +1,43 @@
...
+import { getMessage } from './i18n.import';{{/if}}
 
 const meta: HintMetadata = {
     docs: {
-        category: Category.{{category}},
-        description: `{{description}}`
+        category: Category.{{hint.category}},{{#if packageData.official}}
+        description: getMessage('{{#if packageData.isMulti}}{{hint.normalizedName}}_{{/if}}description', 'en'),
+        name: getMessage('{{#if packageData.isMulti}}{{hint.normalizedName}}_{{/if}}name', 'en'){{else}}
+        description: `{{hint.description}}`,
```
sarvaje: any reason to use ` instead of '?

jaspreet57: Not any specific reason. We can remove this one.

jaspreet57: It was already there earlier, So I just left it as it is. But I think, it should not have any role here.

sarvaje: I was just curious. The only thing is that we should escape the character used (I don't know if we are already doing that or not)

packages/create-hint/src/templates/locales-messages.json.hbs
```diff
@@ -0,0 +1,12 @@
+{
+    {{#each hints}}
+    "{{#if ../isMulti}}{{normalizedName}}_{{/if}}description": {
```
sarvaje: We are not using the `normalizedName` here. We are using the `camelcase`, I think we already have a helper for that "toCamelCase"

packages/create-hint/src/templates/meta.ts.hbs
```diff
@@ -1,21 +1,43 @@
...
+{{#if packageData.official}}
+
+import { getMessage } from './i18n.import';{{/if}}
 
 const meta: HintMetadata = {
     docs: {
-        category: Category.{{category}},
-        description: `{{description}}`
+        category: Category.{{hint.category}},{{#if packageData.official}}
+        description: getMessage('{{#if packageData.isMulti}}{{hint.normalizedName}}_{{/if}}description', 'en'),
```
sarvaje: Same as before, it is not `normalizedName`. There are more places where you need to change it.

packages/hint/docs/contributor-guide/guides/create-custom-hint.md
jaspreet57: done

packages/create-hint/src/templates/meta.ts.hbs
jaspreet57: Hmm right. Checking now if we are escaping or not. However backtick should be still safe to use in such cases.

packages/create-hint/src/templates/locales-messages.json.hbs
sarvaje: You can see an example here: https://github.com/webhintio/hint/blob/master/packages/hint-typescript-config/src/_locales/en/messages.json

packages/create-hint/src/templates/meta.ts.hbs
jaspreet57: we are doing `escape` while creating `NewHint`
```
this.description = escapeSafeString(hintData.description);
```

Should I keep single-quote?

sarvaje: nah, it is ok.

packages/create-hint/src/templates/locales-messages.json.hbs
jaspreet57: Yes right. good catch. Thanks. I didn't see for that long names.

So should I use `toCamelCase` helper on `normalizedName` or any other name?

jaspreet57: I got it. It is helper registered there...

packages/create-hint/src/templates/locales-messages.json.hbs
```diff
@@ -0,0 +1,12 @@
+{
+    {{#each hints}}
+    "{{#if ../isMulti}}{{toCamelCase normalizedName}}_{{/if}}description": {
+        "description": "Metadata description",
+        "message": "{{description}}"
+    },
+    "{{#if ../isMulti}}{{toCamelCase normalizedName}}_{{/if}}name": {
+        "description": "Metadata name",
+        "message": "{{normalizedName}}"
```
jaspreet57: @sarvaje key name is in `camelCase`. But actual name value seems to be `space` delimited with first char capital. Right?
Do we have helper for that in utils? For me it seems like `normalize-string-with-delimiter` will work... But should I create one for same?

sarvaje: can you give me an example?

jaspreet57: Like in `hint-axe` locale file...
https://github.com/webhintio/hint/blob/master/packages/hint-axe/src/_locales/en/messages.json

see `nameRoleValue_name` > `message` value
```
"nameRoleValue_name": {
        "description": "Metadata name",
        "message": "Name role value"
 },
```

I just see similar pattern in many hints though it is just a value, user can configure it.
I was asking for default value when hint gets created.

sarvaje: Ahhh, yes, the actual name is like you said. I was talking only about the key. Sorry for the misunderstanding

jaspreet57: ok that is fine. So my changes are done. Please have a review once.

packages/hint/docs/contributor-guide/guides/create-custom-hint.md
```diff
@@ -32,7 +32,8 @@ the CLI.
 >* Make executable release. run `yarn webpack`.
 >* Now run `npm create hint` from `packages` folder.
 >
->The generated files will be slighthly different (references and task will change a bit).
+>The generated files will be
+>slightly different (references and task will change a bit).
```
antross: If you can, you may want to split this markdown fix into its own commit. 

Context: When we're ready to merge this I'd like to squash down to two commits - one for everything under `packages/create-hint` with the `New:` label and another for the doc updates under `packages/hint` with the `Docs:` label. This will help our release script bump versions appropriately.

jaspreet57: Ahh yes right. Let me do this.

antross:
@sarvaje Do you recall the intent behind the `templates` versus `shared-templates` folders? Perhaps I'm missing something, but I don't see a logical separation between how the two are being used.

packages/hint/docs/contributor-guide/guides/create-custom-hint.md
jaspreet57: @antross I have re-based this PR to 2 commits. One is for docs and another one for updates.

jaspreet57:
> Current changes look good 
> 
> One new request: I noticed we don't currently emit a `context.report` example in the generated `hint.ts`. I think that'd be good to add as part of this PR so we can show how it should be localized when generating an official hint. We should also include the string for that message in the generated `_locales/en/messages.json` file in that case.

Good Idea :)... I will add that..

molant:
Before breaking everything into packages the create- tasks shared some templates.
To speed up the process we decided to copy paste the folder instead of finding something else.

That's the history behind. It should be safe to merge both folders even though there will still be some duplication among the create packages.

Maybe a "good first issue"?

packages/hint/docs/contributor-guide/guides/create-custom-hint.md
antross: Thanks! One minor tweak, let's use `New:` instead of `Chore:` for the commit with the functionality updates. We're adding support for localization of official hints to `create-hint` so I'd consider it a new feature.

jaspreet57:
> > Current changes look good
> > One new request: I noticed we don't currently emit a `context.report` example in the generated `hint.ts`. I think that'd be good to add as part of this PR so we can show how it should be localized when generating an official hint. We should also include the string for that message in the generated `_locales/en/messages.json` file in that case.
> 
> Good Idea :)... I will add that..

@antross Added i18n support in `hint.ts` code also for official packages. `context.report` is now using `getMessage` of `./i18n-imports`.
I just had to pass `{ hint, packageData }` to `hint.ts.hbs` template to differentiate between hint and root data.

And `New` tag also added to feat commits :)

packages/create-hint/src/templates/locales-messages.json.hbs
```diff
@@ -0,0 +1,16 @@
...
+    },
+    "{{#if ../isMulti}}{{toCamelCase normalizedName}}_{{/if}}reportMessage": {
```
antross: Other than `name` and `description`, we don't generally prefix messages in `messages.json`, even for multi-hints. The name and description metadata is an exception because we need those names to follow a well-defined format so we can use them when generating documentation.
```suggestion
    "reportMessage": {
```

packages/create-hint/src/templates/partial-event-code.hbs
```diff
@@ -12,5 +12,9 @@ debug(`Validating hint {{name}}`);
  */
 
 if (Math.ceil(Math.random()) === 0) {
+    {{#if packageData.official}}
+    context.report(resource, getMessage('{{#if packageData.isMulti}}{{toCamelCase hint.normalizedName}}_{{/if}}reportMessage', context.language));
```
antross:
```suggestion
    context.report(resource, getMessage('reportMessage', context.language));
```

packages/create-hint/src/templates/locales-messages.json.hbs
jaspreet57: @antross  that is right if user create new messages in locales file himself. But here I am auto-generating, and for demo purpose, I am keeping name of key as `reportMessage`. Now what should I do for multiple hints here? Key name has to be prefixed with something.

jaspreet57: Or should I just add only one message as `reportMessage` and use it in all hints of multi hint package

antross: Good point. I like your suggestion of using a common message in this case. Otherwise I'm concerned new users would think they have to prefix all messages.

jaspreet57: Thanks @antross. And I have made the change. Please review last commit.

jaspreet57:
> Looks good.
> 
> Can you squash this back down to two commits? Then I'll go ahead and merge. Thanks!

Rebased and squashed. And had fun :)

PR #2805, opened 2019-08-09
New: Improve VS Code first-run experience

### LINKED ISSUE #1942: [vscode]Don't prompt install if .hintrc is missing?
The VS Code extension is prompting to install webhint each time I open a new folder even though I don't want it. Maybe we should be more "polite" and only do this check if a `.hintrc` file is found?

antross:
* Only prompt to install `hint` to a project when `.hintrc` is present
* Fall back to a shared copy of `hint` from extension's global folder
* Use VS Code settings to track telemetry preferences
* Refactor dependencies to reduce package size

- - - - - - - - - -

Fix #1942

## Pull request checklist

Make sure you:

Creating as a draft pending test updates. Feel free to start reviewing core changes.

### DIFF
--- a/packages/extension-vscode/src/extension.ts
+++ b/packages/extension-vscode/src/extension.ts
@@ -1,4 +1,4 @@
-import { workspace, ExtensionContext } from 'vscode';
+import { window, workspace, ExtensionContext, WorkspaceConfiguration } from 'vscode';
 
 import {
     LanguageClient,
@@ -7,7 +7,11 @@ import {
     TransportKind
 } from 'vscode-languageclient';
 
-import * as notifications from './notifications';
+import * as notifications from './utils/notifications';
+
+type TelemetryState = 'ask' | 'disabled' | 'enabled';
+
+const telemetryKey = 'enableTelemetry';
 
 // List of document types the extension will run against.
 const supportedDocuments = [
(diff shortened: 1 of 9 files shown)


packages/extension-vscode/src/utils/config.ts
```diff
@@ -0,0 +1,33 @@
+const defaultConfig: import('hint').UserConfig = { extends: ['development'] };
```
sarvaje: Can you remind me what are the benefits of using `import('hint')` instead of the tipical: `import { UserConfig } from 'hint'` :)

antross: The advantage is that it can only import types, not implementation. Here I want to be absolutely certain we don't accidentally depend on having a copy of `hint` bundled with the extension so I've used the `import('hint')` syntax for all types from `hint` to make that obvious.

packages/extension-vscode/src/extension.ts
```diff
@@ -7,7 +7,11 @@ import {
     TransportKind
 } from 'vscode-languageclient';
 
-import * as notifications from './notifications';
+import * as notifications from './utils/notifications';
+
+type TelemetryState = 'ask' | 'disabled' | 'enabled';
```
sarvaje: You are using this in at least a couple of files, maybe it is better to move this to a place to share it.

PR #2838, opened 2019-08-21
Fix: Correctly calculate nested dependency versions

### LINKED ISSUE #2823: [Bug] `configuration-all` doesn't get the right version of the other configurations.
e.g. Last version of `configuration-all` has this dependencies:
```json
"@hint/configuration-accessibility": "^1.0.3",
"@hint/configuration-development": "^7.0.3",
"@hint/configuration-progressive-web-apps": "^6.0.3",
"@hint/configuration-web-recommended": "^7.0.3",
```

but the latest version of those packages are : `1.0.4`, `^7.0.4`, `^6.0.4` and `^7.0.3` respectively

### LINKED ISSUE #2830: [Bug] Release script results in incorrect version references for nested dependencies
Bumping something like `@hint/utils` can cause a configuration to reference an incorrect version number of one of its hints (since the hint will get a dependency-only version bump and the configuration will reference the outdated version of the hint, if it gets updated at all). We noticed this in particular with the new `@hint/configuration-all`, but it appears others are affected.

antross:
Fix #2830
Fix #2823

## Pull request checklist

Make sure you:


release/lib/update-dependencies.ts
```diff
@@ -47,30 +122,49 @@ const updateProperty = (property: string) => {
...
+            }
+
+            pkg.content[property][dependencyName] = dependencyVersion;
+            pkg.updated = true;
+
+            /*
+             * Update the version of this package (`patch`) if it has not been bumped yet
+             * (because it has changed versions in its dependencies).
+             */
+            if (pkg.oldVersion === pkg.content.version && pkg.updated) {
```
sarvaje: You don't need `&& pkg.updated` because is going to be always true.

release/lib/update-dependencies.ts
```diff
@@ -15,6 +19,77 @@ const propertiesToUpdate = [
     'peerDependencies'
 ];
 
+const nonOptionalDependencyTypes = [
```
sarvaje: Nit: They are properties, what about change the name to `nonOptionalDependencyProperties`? (or any reduce version of that name hehehe)

release/lib/update-dependencies.ts
antross: Good point. I'm actually kind of surprised neither TypeScript nor ESLint flagged that 🤔

release/lib/update-dependencies.ts
antross: Sure.

Issue #2033, opened 2019-03-07
[Feature] Improve selector query in highest-available-document-mode


molant:
Right now we are doing some weird filtering when we can probably use a more direct selector once #1943 is merged. We should look into improving it to avoid iterating that much.

PR #2846, opened 2019-08-22
Chore: Remove iterating over elements and use direct selector

jaspreet57:
## Pull request checklist

Make sure you:

In `hint-highest-available-document-mode`, `getXUACompatibleMetaElements` function is defined to filter out elements from given `elements` that have attribute `http-equiv` with value `x-ua-compatible`.

It can be replaced by using direct selectors. Selectors with attributes and values are now supported by `querySelectorAll`.

So we can use `meta[http-equiv=x-ua-compatible i]` selector to get `XUACompatibleMetaElements`.

`i` in the selector is to ignore case.

Resolves: #2033

### DIFF
--- a/packages/hint-highest-available-document-mode/src/hint.ts
+++ b/packages/hint-highest-available-document-mode/src/hint.ts
@@ -34,20 +34,6 @@ export default class HighestAvailableDocumentModeHint implements IHint {
         let requireMetaElement: boolean = false;
         let suggestRemoval: boolean = false;
 
-        /*
-         * This function exists because not all connector (e.g.: jsdom)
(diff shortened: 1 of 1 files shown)


jaspreet57:
@molant I created this PR as this issue was in open state since a long time.
This might not increase the performance, as `css-select` internally is doing the same iteration.
But it removes the unnecessary filtering being done in our code.

Please let me know if this PR could still be useful or not?

antross:
> Please let me know if this PR could still be useful or not?

@jaspreet57 This is definitely still useful. While you're correct it may not result in a performance improvement, it helps simplify our code which is much appreciated.

PR #2848, opened 2019-08-22
New: Prompt user to opt in to telemetry in browser extension

### LINKED ISSUE #2818: [Feature] Prompt user to opt in to telemetry in browser extension
# 🚀 Feature request

Prompt user to opt in to telemetry in browser extension.

sarvaje:
Fix #2818

## Pull request checklist

Make sure you:


sarvaje:
This is how it looks:

image

image

image

image

image

image

sarvaje:
And here are the bundles
Chromium: 
[webhint-0.0.6.zip](https://github.com/webhintio/hint/files/3531824/webhint-0.0.6.zip)

Firefox:
[webhint-0.0.6.zip](https://github.com/webhintio/hint/files/3531826/webhint-0.0.6.zip)

ststimac:
@sarvaje 
[icons.zip](https://github.com/webhintio/hint/files/3535715/icons.zip)

sarvaje:
Thanks @ststimac!!

antross:
> we probably still need to expose some way to change the telemetry choice after the initial prompt.

@molant suggested adding a collapsed details/summary section to the configuration page for settings (maybe just below the "Restore defaults" button). Then we could put a toggle inside there for the telemetry setting. That would avoid needing to create a separate settings page until we have more things to include. 

@ststimac Does that sound reasonable?

ststimac:
@antross that sounds good to me!

molant:
I've created #2920 to track the details/summary but maybe we want to add it into this PR?

antross:
> I've created #2920 to track the details/summary but maybe we want to add it into this PR?

Yeah, @sarvaje and I were discussing just including it in this PR yesterday. That seems better to ensure both of these pieces go out together in the next release.

molant:
> That seems better to ensure both of these pieces go out together in the next release.

I'll wait then until that change is in to provide feedback

sarvaje:
This is how the new settings looks:
image

image

antross:
@sarvaje - can you also add an `f12-telemetry` event that logs when a user first opts-in to telemetry? Thanks!

packages/extension-browser/src/devtools/views/app.base.css
```diff
@@ -5,6 +5,7 @@
     --base-color: #000;
 
     --code-bg: var(--base-bg);
+    --popup-bg: #efefef;
```
antross: Should we call this `--notification-bg` for consistency with how the component is named?

packages/extension-browser/src/devtools/views/app.tsx
```diff
@@ -30,12 +32,20 @@ export type Props = {
     results?: ResultsData;
 };
 
+/*
+ * If we need to show the notification, we don't want it to be removed
+ * from the dom after the status change so we don't lose the animation.
+ */
+const addNotification = showOptIn();
```
antross: Since this never changes if it's `true` when we start, why not always leave the notification in the DOM? Seems simpler.

packages/extension-browser/src/devtools/views/controls/settings.css
```diff
@@ -0,0 +1,3 @@
+.root {
+    margin-top: 1.5rem;
```
antross: Could we achieve this with a grid and gap in the parent component (or at least with a passed through class name)? I don't like having child-components control layout outside of their contents.

packages/extension-browser/src/devtools/views/controls/settings.tsx
```diff
@@ -0,0 +1,44 @@
...
+import Checkbox from './checkbox';
+import LabelText from './label-text';
+
+import * as styles from './settings.css';
+
+const settings = { telemetry: 'telemetry' };
+
+type Props = {
+    isTelemetryEnabled: boolean;
+    onTelemetryChange: (enable?: boolean) => void;
```
antross: Why is `enable` optional? Looks like we always pass `true` or `false`. Narrowing the type here should also simplify checks in components which consume this.

packages/extension-browser/src/devtools/views/controls/settings.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+const Settings = ({isTelemetryEnabled, onTelemetryChange}: Props) => {
+
+    const onEnabledChange = useCallback((event: FormEvent<HTMLInputElement>) => {
+        const input = (event.target as HTMLInputElement);
+        const setting = input.value;
+
+        if (setting === settings.telemetry) {
```
antross: This check seems unnecessary. You're only assigning this handler to a checkbox which is intended for telemetry, so you shouldn't need to provide or check `input.value`.

packages/extension-browser/src/devtools/views/controls/telemetry-notification.tsx
```diff
@@ -0,0 +1,43 @@
...
+const TelemetryNotification = ({ onTelemetryChange, show }: Props) => {
+    const onEnableTelementry = useCallback(() => {
+        onTelemetryChange(true);
+    }, [onTelemetryChange]);
+
+    const onDisableTelemetry = useCallback(() => {
+        onTelemetryChange(false);
+    }, [onTelemetryChange]);
+
+    const getActions = () => {
```
antross: No need for this to be a function. Just store the array of `Button`s here directly and reference it below.
```suggestion
    const actions = ([
```

packages/extension-browser/src/devtools/views/controls/telemetry-notification.tsx
```diff
@@ -0,0 +1,43 @@
...
+            <Button key="1" primary={true} onClick={onEnableTelementry}>{getMessage('enable')}</Button>,
+            <Button key="2" primary={true} onClick={onDisableTelemetry}>{getMessage('noThanks')}</Button>
+        ]);
+    };
+
+    return (
+        <Notification show={show} actions={getActions()}>
+            <span className={styles.message}>
+                {getMessage('helpUs')}
+                &nbsp;<ExternalLink href="https://webhint.io/docs/user-guide/telemetry/summary/">{getMessage('learnMore')}</ExternalLink>
```
antross: Formatting nit (also a regular space should be fine here):
```suggestion
                {' '}
                <ExternalLink href="https://webhint.io/docs/user-guide/telemetry/summary/">
                    {getMessage('learnMore')}
                </ExternalLink>
```

packages/extension-browser/src/devtools/views/app.tsx
sarvaje: Because of the animations. If we keep it always in the dom, the "hidden" animation will always work and it will make a weird effect.

packages/extension-browser/src/devtools/views/controls/settings.tsx
sarvaje: I thought I will need that for "close" the notification, but that is not necessary and I forgot to change that :)

packages/extension-browser/src/devtools/views/controls/settings.tsx
sarvaje: This was thinking in the future with more settings, but I can remove it if you want to.

antross: Yes, I would expect each setting to have its own callback so let's remove it.

sarvaje: Then this shouldn't be here. I should create a new component for this setting in particular.

sarvaje:
@molant this should be ready to review

packages/extension-browser/src/devtools/views/app.tsx
antross: Ah yes. Add a `transition` for `visibility` from `visible` to `hidden` over the same duration as the animation (we should have this end up with `visibility: hidden` anyway to avoid getting keyboard focus after it's gone).

antross:
@molant you're up 😉

packages/extension-browser/src/_locales/en/messages.json
```diff
@@ -111,6 +127,10 @@
         "message": "None",
         "description": "Label for option to not ignore any resources on the page."
     },
+    "noThanks": {
+        "message": "No Thanks",
```
molant:
```suggestion
        "message": "No, thanks",
```

antross: Whoa, the comma feels weird here 😵(though possibly more technically correct). Personally I prefer the existing treatment of "No Thanks".

PR #2851, opened 2019-08-22
Fix: PWA manifest-file-extension is wrong

sarvaje:
## Pull request checklist

Make sure you:

- [ ] Signed the [Contributor License Agreement](https://cla.js.foundation/webhintio/hint)


packages/utils/src/fs/file-extension.ts
```diff
@@ -14,7 +14,7 @@ export const fileExtension = (resource: string): string => {
          */
         url = new URL(resource);
     } catch (err) {
-        return extname(resource).split('.')
+        return extname(resource ? resource.split('?')[0] : '').split('.')
```
antross: Should we have the additional spot fix here? We certainly don't need it to fix this specific bug since you've already updated that caller to pass an absolute URL. And if other callers have the same problem, wouldn't we prefer to update them to pass absolute URLs as well?

sarvaje: > We certainly don't need it to fix this specific bug since you've already updated that caller to pass an absolute URL

I agree.

> And if other callers have the same problem, wouldn't we prefer to update them to pass absolute URLs as well

I prefer to have this fix for those scenarios. Passing a relative URL shouldn't be a problem, and that probably wasn't a problem at the beginning, because I think the old URL API supported relative urls.

antross: If you want to keep relative URLs supported, wouldn't it be less special-case to supply a dummy base URL on line 15? That would make relative URLs resolve so we can use the existing code path to get the file extension without having to maintain special handling in the `catch` clause.
```ts
url = new URL(resource, 'http://example.com');
```

sarvaje: Sure, I can do that

sarvaje: @antross done!

PR #2852, opened 2019-08-23
Enable `hint-scoped-svg-styles` in default configuration

jaspreet57:
## Pull request checklist

Make sure you:

Enabling `hint-scoped-svg-styles` and `private: true` removed from the hint.

### DIFF
--- a/packages/configuration-development/index.json
+++ b/packages/configuration-development/index.json
@@ -21,6 +21,7 @@
         "meta-viewport": "error",
         "no-bom": "error",
         "no-protocol-relative-urls": "error",
+        "scoped-svg-styles": "error",
         "sri": "error",
         "typescript-config/consistent-casing": "error",
         "typescript-config/import-helpers": "error",
--- a/packages/configuration-development/package.json
+++ b/packages/configuration-development/package.json
@@ -15,6 +15,7 @@
     "@hint/hint-meta-viewport": "^4.1.3",
     "@hint/hint-no-bom": "^4.1.3",
     "@hint/hint-no-protocol-relative-urls": "^2.3.3",
+    "@hint/hint-scoped-svg-styles": "^1.0.0",
     "@hint/hint-sri": "^3.2.3",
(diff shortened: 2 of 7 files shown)


jaspreet57:
@antross Should I require to push `private: true` removal change first?

I think build failure is because this `hint` is not published yet.. Could you please check once.

antross:
Ah, actually these failures appear to be due to the webpack build for the browser extension pulling in more packages than we want it to. This happens because we still compile to the `node` modules format so we don't benefit from tree shaking yet. To work around that we'll need to update some of the `import` statements in your new hint to be more scoped to help webpack out. I'll look through and flag the ones I see.

antross:
Looks like it's just the `debug` method from `@hint/utils`. `HTMLElement` is only used as a type so it gets stripped automatically after the TypeScript build runs, but still wouldn't hurt to use a scoped import for it too.
https://github.com/webhintio/hint/blob/3aba15082be60e6f1a5743850901227751ff1a8b/packages/hint-scoped-svg-styles/src/hint.ts#L7

```diff
- import { debug as d, HTMLElement } from '@hint/utils';
+ import { debug as d } from '@hint/utils/dist/src/debug';
+ import { HTMLElement } from '@hint/utils/dist/src/dom/html';
```

jaspreet57:
> Looks like it's just the `debug` method from `@hint/utils`. `HTMLElement` is only used as a type so it gets stripped automatically after the TypeScript build runs, but still wouldn't hurt to use a scoped import for it too.
> https://github.com/webhintio/hint/blob/3aba15082be60e6f1a5743850901227751ff1a8b/packages/hint-scoped-svg-styles/src/hint.ts#L7
> 
> ```diff
> - import { debug as d, HTMLElement } from '@hint/utils';
> + import { debug as d } from '@hint/utils/dist/src/debug';
> + import { HTMLElement } from '@hint/utils/dist/src/dom/html';
> ```

Thanks @antross. I have changed it.

PR #2893, opened 2019-08-30
Fix: Using `formatter-stylish` in CI by default.

### LINKED ISSUE #2892: [Feature] Improve logging in CI
Right now, we are using `formatter-summary` by default for the configurations. This formatter only says what "needs to be fixed" but not where.

During today's retrospective we agree that we should give more information by default in CI.

Even if the user is able to use its own configuration, the idea is to offer to the user more information by default.

To achieve this we have at least two options:

1. Check if we are in CI and set by default another formatter like `stylish`.
1. Modify the configurations to use `stylish` instead of `summary`.

I'm saying `stylish` but we can use any other. (But I prefer stylish :P)

sarvaje:
Fix #2892

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint/package.json
+++ b/packages/hint/package.json
@@ -74,8 +74,7 @@
   },
   "optionalDependencies": {
     "@hint/configuration-development": "^7.1.0",
-    "@hint/configuration-web-recommended": "^7.1.0",
-    "@hint/formatter-stylish": "^3.1.5"
+    "@hint/configuration-web-recommended": "^7.1.0"
   },
   "repository": {
     "directory": "packages/hint",
--- a/packages/hint/package.json
+++ b/packages/hint/package.json
@@ -73,8 +73,9 @@
     "extends": "../../.nycrc"
   },
   "optionalDependencies": {
-    "@hint/configuration-development": "^7.0.3",
-    "@hint/configuration-web-recommended": "^7.0.3"
+    "@hint/configuration-development": "^7.1.0",
+    "@hint/configuration-web-recommended": "^7.1.0",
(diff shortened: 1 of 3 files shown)


molant:
@antross can you please open an issue to discuss that? Probably part of the next version discussion we should have.

antross:
@sarvaje want to add the test case @molant suggested so we can unblock and merge this PR?

sarvaje:
ok

sarvaje:
I'm taking a look to the code and that test case is already there:
https://github.com/webhintio/hint/pull/2893/files#diff-41369f12d00e1d73e14241156907f6ffR207
https://github.com/webhintio/hint/pull/2893/files#diff-41369f12d00e1d73e14241156907f6ffR207

antross:
@sarvaje maybe I'm missing something, but I don't see a test that validates it only uses `html` and `summary` when *NOT* in CI...

sarvaje:
no literally, but we are checking that the configuration is { extends: ["web-recommended"]} and {extends: ["development"]}

molant:
It's weird to me that we have a test to check the formatters in one environment and not another.

sarvaje:
The test is there, if it is not CI, the configuration is { extends: ["web-recommended"]} if it is in CI, the configuration is {extends: ["web-recommended"], formatters: ["stylish", "html"]}

If we are not in CI, the property formatter is not initialized and we are using whatever comes from the extends.

sarvaje:
I can't check the property formatter, because when I'm not in CI, I only use the extends. And I don't think we should test here what the extends has by default.

molant:
OK, then 👍

PR #2933, opened 2019-09-05
New: Add parser for SASS and SCSS files

### LINKED ISSUE #2887: Support SASS files

antross:
Fix #2887

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/extension-vscode/src/extension.ts
+++ b/packages/extension-vscode/src/extension.ts
@@ -10,16 +10,14 @@ import {
 import { TelemetryState } from './utils/analytics';
 import * as notifications from './utils/notifications';
 
+const { activationEvents } = require('../../package.json');
+
 const telemetryKey = 'enableTelemetry';
 
 // List of document types the extension will run against.
-const supportedDocuments = [
-    'css',
-    'html',
-    'javascript',
-    'json',
-    'jsonc'
-];
+const supportedDocuments = activationEvents.map((event: string) => {
+    return event.split(':')[1];
+});
 
 // Keep a reference to the client to stop it when deactivating.
 let client: LanguageClient;
--- a/packages/parser-sass/README.md
+++ b/packages/parser-sass/README.md
@@ -1,0 +1,62 @@
+# SASS (`@hint/parser-sass`)
+
+The `SASS` parser is built on top of [PostCSS][postcss] so hints can
+analyze `SASS` and `SCSS` files. It emits the same events as
+`@hint/parser-css` so existing hints targeting `CSS` files will work
+without modification.
+
+To use it you will have to install it via `npm`:
+
+```bash
+npm install @hint/parser-sass
+```
+
+Note: You can make `npm` install it as a `devDependency` using the
+`--save-dev` parameter, or to install it globally, you can use the
+`-g` parameter. For other options see [`npm`'s
+documentation](https://docs.npmjs.com/cli/install).
+
+And then activate it via the [`.hintrc`][hintrc] configuration file:
+
+```json
+{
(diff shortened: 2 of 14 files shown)


packages/extension-vscode/src/extension.ts
```diff
@@ -10,16 +10,14 @@ import {
...
 
 // List of document types the extension will run against.
-const supportedDocuments = [
-    'css',
-    'html',
-    'javascript',
-    'json',
-    'jsonc'
-];
+const supportedDocuments = activationEvents.map((event: string) => {
```
molant: 😍

packages/parser-sass/README.md
```diff
@@ -0,0 +1,62 @@
+# SASS (`@hint/parser-sass`)
+
+The `SASS` parser is built on top of [PostCSS][postcss] so hints can
+analyze `SASS` and `SCSS` files. It emits the same events as
+`@hint/parser-css` so existing hints targeting `CSS` files will work
```
molant: >  so existing hints targeting `CSS` files will work

🎉

packages/parser-sass/README.md
```diff
@@ -0,0 +1,62 @@
...
+without modification.
+
+To use it you will have to install it via `npm`:
+
+```bash
+npm install @hint/parser-sass
+```
+
+Note: You can make `npm` install it as a `devDependency` using the
+`--save-dev` parameter, or to install it globally, you can use the
```
molant: > or to install it globally

Did we decide to go against global installs? I believe that individual packages installed globally will not work, only if they are in a configuration...
Do we have a task to update all the docs to remove this?

packages/parser-sass/package.json
```diff
@@ -0,0 +1,84 @@
...
+  "license": "Apache-2.0",
+  "main": "dist/src/parser.js",
+  "name": "@hint/parser-sass",
+  "nyc": {
+    "extends": "../../.nycrc"
+  },
+  "peerDependencies": {
+    "hint": "^5.2.2"
+  },
+  "repository": {
```
molant: Do we want a `"private": true` here or are we ready to publish it directly?

In the past we were doing it because of the documentation, right? But now that we get everything from npm it should be fine?

packages/parser-sass/tests/interface-tests.ts
```diff
@@ -0,0 +1,165 @@
...
+    const args = engineEmitAsyncSpy.thirdCall.args;
+    const data = args[1] as StyleParse;
+
+    t.is(args[0], 'parse::end::css');
+    t.is(data.code, code);
+    t.is(data.element, null);
+    t.is(data.resource, 'styles.scss');
+});
+
+test('If fetch::end::* is received for CSS, nothing should happen', async (t) => {
```
molant: What about the following? It took me a second to understand what the intention was:

```suggestion
test('If fetch::end::css is emitted, nothing should happen', async (t) => {
```

packages/parser-sass/README.md
antross: That seems to be the growing consensus. No existing task to go forth and change the docs yet, though. I'm on board if you want to create one.

packages/parser-sass/package.json
antross: Yeah, I don't think we need to keep packages `private` anymore unless we're expecting to do a release without them. This one should be ready to go once we merge.

antross:
@molant since every commit here is a "New" I propose just squashing them down to one when merging. I originally kept them separate so we could see what touched which packages more easily, but I don't think that's necessary going forward. Sound good to you?

PR #2936, opened 2019-09-06
New: Add parser for LESS files

### LINKED ISSUE #2886: Support LESS files

antross:
## Pull request checklist

Make sure you:

Fix #2886

### DIFF
--- a/packages/extension-vscode/package.json
+++ b/packages/extension-vscode/package.json
@@ -5,6 +5,7 @@
     "onLanguage:javascript",
     "onLanguage:json",
     "onLanguage:jsonc",
+    "onLanguage:less",
     "onLanguage:sass",
     "onLanguage:scss"
   ],
--- a/packages/parser-less/src/parser.ts
+++ b/packages/parser-less/src/parser.ts
@@ -1,0 +1,56 @@
+const parser = require('postcss-less');
+
+import * as postcss from 'postcss';
+
+import { debug as d } from '@hint/utils/dist/src/debug';
+import { normalizeString } from '@hint/utils/dist/src/misc/normalize-string';
+import { HTMLElement } from '@hint/utils/dist/src/dom/html';
+import { Parser } from 'hint/dist/src/lib/types';
+import { StyleEvents } from '@hint/parser-css';
+import { Engine } from 'hint';
+
+const debug = d(__filename);
+
+export default class CSSParser extends Parser<StyleEvents> {
+    public constructor(engine: Engine<StyleEvents>) {
+        super(engine, 'less');
+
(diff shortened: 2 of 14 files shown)


packages/extension-vscode/package.json
```diff
@@ -4,7 +4,8 @@
     "onLanguage:html",
     "onLanguage:javascript",
     "onLanguage:json",
-    "onLanguage:jsonc"
+    "onLanguage:jsonc",
```
molant: I foresee a conflict here soon 😜

packages/parser-less/src/parser.ts
```diff
@@ -0,0 +1,55 @@
...
+            const code = fetchEnd.response.body.content;
+            const resource = fetchEnd.resource;
+
+            if (fetchEnd.response.mediaType === 'text/less') {
+                await emitLESS(code, resource, null);
+            }
+        });
+
+        engine.on('element::style', async ({ element, resource }) => {
+            const rel = normalizeString(element.getAttribute('rel'));
```
molant: I'm looking at the [docs and `rel` is when using a `.less` file](http://lesscss.org/usage/#using-less-in-the-browser):

```html
<link rel="stylesheet/less" type="text/css" href="styles.less" />
```

so this won't trigger with the current event you are subscribing to.

The docs doesn't seem to document it, but it looks like any `<style type="">` that matches `/^text\/(x-)?less$/` should work:

https://github.com/less/less.js/blob/cae5021358a5fca932c32ed071f652403d07def8/lib/less-browser/index.js#L64

packages/parser-less/tests/interface-tests.ts
```diff
@@ -0,0 +1,133 @@
...
+    const args = engineEmitAsyncSpy.thirdCall.args;
+    const data = args[1] as StyleParse;
+
+    t.is(args[0], 'parse::end::css');
+    t.is(data.code, code);
+    t.is(data.element, null);
+    t.is(data.resource, 'styles.less');
+});
+
+test('If fetch::end::* is received for CSS, nothing should happen', async (t) => {
```
molant: Same comment as for SASS:

```suggestion
test('If fetch::end::css is received, nothing should happen', async (t) => {
```

packages/extension-vscode/package.json
antross: Indeed 😉

packages/parser-less/src/parser.ts
antross: `<link>`, `<style>`, what's the difference?🤦‍♂️

I've also been wondering if `<style lang="less">` works with Vue since that seems to be the approach there for single file components once we're ready to analyze them. I see examples for SASS, but haven't found anything concrete for LESS yet.

PR #2937, opened 2019-09-06
Fix: Clear cache and service worker data on reload

### LINKED ISSUE #2928: Force a cache clean and bypass service workers when possible
Windows 8.1
Canary 78 and Chrome 76
Webhint Extension 0.0.7

I'm using the suggested Apache .htaccess configuration, yet this hint is thrown for the webmanifest. It's once again appending the charset to the content-type; reason unknown.

> hint 1: 'content-type' header value should not contain 'charset=utf-8'.
> https://thomasbrodhead.com/favicons/site.20190904080058.webmanifest
> Content-Type: application/manifest+json; charset=utf-8

Is this related to Issue #2927?

molant:
## Pull request checklist

Make sure you:

Explicitly delete the cache and any service worker data before reloading. 

Fix #2928

Bundles for testing:

[chromium 2019-09-09](https://github.com/webhintio/hint/files/3592792/webhint-0.0.7.zip)
[firefox 2019-09-09](https://github.com/webhintio/hint/files/3592794/webhint-0.0.7.zip)

### DIFF
--- a/packages/extension-browser/src/content-script/connector.ts
+++ b/packages/extension-browser/src/content-script/connector.ts
@@ -4,7 +4,7 @@ import { Engine } from 'hint';
 import { getElementByUrl, HTMLDocument, HTMLElement, traverse } from '@hint/utils/dist/src/dom';
 import { createHelpers, restoreReferences } from '@hint/utils/dist/src/dom/snapshot';
 import { DocumentData } from '@hint/utils/dist/src/types/snapshot';
-import { getContentTypeData, getType } from '@hint/utils/dist/src/content-type';
+
 import {
     ConnectorOptionsConfig,
     IConnector,
@@ -15,6 +15,7 @@ import {
 import { browser, document, eval, location, MutationObserver, window } from '../shared/globals';
 import { Events } from '../shared/types';
 import { Fetcher } from './fetcher';
+import { setFetchType } from './set-fetch-type';
 
 export default class WebExtensionConnector implements IConnector {
     private _document: HTMLDocument | undefined;
@@ -124,15 +125,6 @@ export default class WebExtensionConnector implements IConnector {
         }
     }
 
-    private setFetchType(event: FetchEnd): string {
-        const { charset, mediaType } = getContentTypeData(null, event.response.url, event.response.headers, null as any);
-
-        event.response.charset = charset || '';
-        event.response.mediaType = mediaType || '';
-
-        return getType(mediaType || '');
-    }
-
     private async notifyFetch(event: FetchEnd) {
         /*
          * Delay dispatching FetchEnd until we have the DOM snapshot to populate `element`.
@@ -145,7 +137,7 @@ export default class WebExtensionConnector implements IConnector {
         }
 
         this.setFetchElement(event);
-        const type = this.setFetchType(event);
+        const type = setFetchType(event);
 
         await this._engine.emitAsync(`fetch::end::${type}` as 'fetch::end::*', event);
     }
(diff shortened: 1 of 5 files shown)


molant:
Yes, I need to update tests. I wanted to validate it was working for @bmt-systems but it looks like more work is needed.

molant:
I've updated this PR and right now I'm just ignoring requests coming from service workers. Seems to be more reliable.

molant:
Also the bundles are updated

packages/extension-browser/src/devtools/utils/network.ts
```diff
@@ -79,10 +79,45 @@ const generateFetchEnd = async (entry: chrome.devtools.network.Request) => {
     hops.delete(url);
 };
 
+/**
+ * Checks if a request is comming from a Service|Web worker using the
+ * `referer` header to identify it.
```
antross: Probably good to include *why* we want to ignore requests from workers somewhere in the comments. Otherwise looks good.

molant: This is explained in line 118

molant:
This is ready for another review.
I've updated the bundles in the first comment as well.

packages/extension-browser/src/content-script/connector.ts
```diff
@@ -147,7 +161,7 @@ export default class WebExtensionConnector implements IConnector {
         this.setFetchElement(event);
         const type = this.setFetchType(event);
 
-        await this._engine.emitAsync(`fetch::end::${type}` as 'fetch::end::*', event);
+        await this.handleFetch(`fetch::end::${type}` as 'fetch::end::*', { fetchEnd: event });
```
antross: Thinking about this some more I realized it only fixes the Chrome code path for the `Fetcher` class. When running in Firefox we return the response of a `fetch` call initiated by `Fetcher` directly. This means `setFetchElement` and `setFetchType` never get a chance to run in that context so we're still missing `element` and `mediaType` data.

antross: Arguably `element` isn't as relevant here since these were triggered by a `fetchContent` call to begin with. This means the caller should have the `element` already so we don't need to wait for it here. And `mediaType` is derived from the headers so we can call that anytime.

TL;DR - I think we can roll back this particular change, add the `mediaType` from within the `Fetcher` class itself, and require callers of `fetchContent` to set the `element`.

molant:
😒

packages/extension-browser/src/content-script/connector.ts
molant: Updated. The `element` is already added in `parser-manifest` so no need to re-add it.

packages/extension-browser/src/content-script/fetcher.ts
```diff
@@ -108,6 +110,8 @@ export class Fetcher {
     private resolveFetch(event: FetchEnd) {
```
antross: To make this work in the Firefox code-path, we should replace the following above (around line 70) with a call to this `resolveFetch` method instead.

```diff
-                if (this._fetches.has(target) && response.type === 'basic') {
+                if (response.type === 'basic') {
                    // Firefox code-path (can't use other path as fetch doesn't trigger onRequestFinished).
                    const data = await toNetworkData(target, headers, response);

-                    resolve(data);
-                    this._fetches.delete(target);
+                    this.resolveFetch(data);
                } else {
```

antross: Resolving as we discussed this data is already set by the `toNetworkData` helper.

PR #2954, opened 2019-09-09
Chore: Bundle VS Code extension with webpack

### LINKED ISSUE #2803: Bundle the VS Code extension using webpack
Currently we can't reference shared code from other packages (e.g. `@hint/utils`) without pulling in the entirety of that package and *ALL* of its dependencies into the final packaged extension (adding many MB to the file size). 

Adding a webpack step similar to what we do for the browser extension would help solve this problem by only bundling the parts of reference dependencies we actually use. It's also the approach now recommended by the VS Code team: https://code.visualstudio.com/api/working-with-extensions/bundling-extension

antross:
Reduces extension size and unblocks sharing utilities
with other parts of webhint.

- - - - - - - - - -

Fix #2803

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/extension-vscode/src/utils/packages.ts
+++ b/packages/extension-vscode/src/utils/packages.ts
@@ -28,7 +28,10 @@ export const installPackages = async (packages: string[], options?: InstallOptio
 
 /**
  * Load the provided packages from the specified location.
- * Always performs a dynamic load, even if bundled with webpack.
+ * Always performs a dynamic load, even if bundled with webpack, to ensure
+ * locally installed versions of packages such as `hint` are used instead of
+ * being bundle with the extension itself. This avoids needing to update the
+ * extension every time webhint updates.
  */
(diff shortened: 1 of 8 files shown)


packages/extension-vscode/src/utils/packages.ts
```diff
@@ -25,12 +28,13 @@ export const installPackages = async (packages: string[], options?: InstallOptio
 
 /**
  * Load the provided packages from the specified location.
+ * Always performs a dynamic load, even if bundled with webpack.
```
molant: I think you are missing the `why` here 🙂

antross: Indeed I am - good catch!

antross:
> Should we do something similar to the browser extension where we have a size budget?

Yes, I'll file an issue to track that.

Issue #2828, opened 2019-08-19
[Bug] Incorrect hint on location of charset defintion


tmb-github:
The browser extension complains:

'charset' meta element should be the first thing in '<head>'.

This is not true, and is widely misunderstood (and sadly so). The element containing the character encoding declaration must be within the first 1024 bytes of the document. There is nothing in the specs mandating that it be the first thing in the head element. It is therefore perfectly valid for it to follow the title element, for example, so long as it is still within the first 1024 bytes of the document.

See for reference:

https://www.w3.org/TR/html52/document-metadata.html#character-encoding-declaration

https://www.w3.org/International/questions/qa-html-encoding-declarations

Please revise the analysis engine to see if the charset is within the first 1024 bytes of the html document.

Many thanks!
Tom

# 🐛 Bug report

## Environment

* __webhint version:__ 
* __Browser version:__ 
* __URL for which webhint failed:__ 

## `webhint` configuration

<details>
<summary>webhint’s configuration</summary>

Categories:
* [x] Accessibility
* [x] Compatibility
* [x] Performance
* [x] Pitfalls
* [x] PWA
* [x] Security

Your target browsers:
* [x] Recommended settings
* [ ] Custom: 

Ignored resources:
* [x] None
* [ ] Different origin
* [ ] Custom: 

</details>

## Debug output

Please include the content of the `Error details` section if an error message was displayed.

<details>
<summary>webhint’s Error details</summary>

```text
<!-- ✍️ Paste the error details here -->


```

</details>

molant:
Thanks for opening this issue @tmb-github!

The [docs of this hint](https://webhint.io/docs/user-guide/hints/hint-meta-charset-utf-8/#why-is-this-important) explains why we advice to be the first thing:

> It needs to be inside the`<head>` element and within the first 1024 bytes of the HTML, as some browsers only look at those bytes before choosing an encoding.
Moreover, it is recommended that the meta tag be the first thing in the <head>. This ensures it is before any content that could be controlled by an attacker, such as a `<title>` element, thus avoiding potential encoding-related security issues (such as the one in old IE).

Please let us know if there's any information here that's not accurate or if you still believe changes should be done!

tmb-github:
[...]

PR #2968, opened 2019-09-11
Docs: Severity changes RFC

molant:
This is an RFC proposal to change how the severity works in webhint and how it should be working. I'm opening it as a draft as I don't have the intention to get the document merged. I'll use the final version as the text for the epic that will be created if we decide to pursue this changes.

This is a first draft and you can start reviewing it 🙂

### DIFF
--- a/rfcs/severity.md
+++ b/rfcs/severity.md
@@ -71,7 +71,7 @@ The new values are `hint` and `information`.
   decide if we swap them or if we are ok as that (in the description about it
   doesn't quite match the expectations I think).
 
-### Hint severity
+### Hint severity examples
 
 As mentioned previously, there are a few hints that report multiple things
 were the relevance is not the same. An example would be `http-compression`
@@ -89,7 +89,21 @@ Having all the reports with the same severity does not help the user focus on
 what's important (have compression enabled for only text based resources
 ([see GitHub comment](https://github.com/webhintio/hint/issues/2919#issuecomment-530190038))
 
-> Other examples here
+Another example will be with the `compat-api` hint.
+Any reports coming from this hint are marked as a "warning" in the default
+configurations. This will be the default with the new schema with the caveat
+that if we feel highly confident a feature may cause a compatibility issue,
+it could be upgraded to be an "error".
+
+On the other side, the hint has a list of "ignored feautes". These are
+features that are considered safe for graceful degradation (e.g. `integrity`
+attribute) and do not trigger any report. When one of this features is found,
+instead of ignoring it a report with the severity "hint" could be created.
+
+There is also an on going conversation to
+[annotate features which can be used safely without support](https://github.com/mdn/browser-compat-data/issues/3365).
+Having this information will help automate and make more accurate the
+scenarios described above for `compat-api`.
 
 ### User configuration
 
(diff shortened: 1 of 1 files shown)


rfcs/severity.md
```diff
@@ -0,0 +1,130 @@
...
+        }
+    },
+    "hintsTimeout": 120000,
+    "hints": {
+        "amp-validator": "off",
+        "apple-touch-icons": "on",
+        "button-type": "off",
+        "content-type": "on",
+        ...
+        "no-vulnerable-javascript-libraries": ["on", {
```
antross: If we're making this change, I think we should imply `"on"` when passing options to a hint.
```suggestion
        "no-vulnerable-javascript-libraries": {
            "severity": "high"
        }
```

rfcs/severity.md
```diff
@@ -0,0 +1,130 @@
...
+    "hintsTimeout": 120000,
+    "hints": [
+        "apple-touch-icons",
+        "content-type",
+        ...
+        "no-vulnerable-javascript-libraries"
+        ]
+    }
+}
+```
```
molant: One thing I forgot to add is that a user should be now capable of setting a threshold for the severity and how that impacts the exit code.
E.g.: A user might want to be notified on `warning` and higher and ignore everything else. The process exit code should be different than 0 even if there are only warnings.

What about a `threshold` property with `"error|warning|hint|information"` as possible values?

rfcs/severity.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Hint severity level
+
+The idea is to get closer to the severity options available in
+[VS Code](https://code.visualstudio.com/api/references/vscode-api#DiagnosticSeverity):
+
+| Severity | Description |
+| -------- | ----------- |
+| off      | Disabled, no reporting will come from it |
+| error    | Something that the user should get fixed right away |
```
molant: I'll add the following images to the doc so we have a better idea of how it will look like in VS Code

VS Code Error
vs code error

VS Code Warning
vs code warning

VS Code Information
vs code information

VS Code Hint
vs code hint

Elchi3:
(Chiming in from https://github.com/webhintio/hint/issues/2828#issuecomment-532326790)

Generally, I like the idea of more severity levels.

Would this also help with https://github.com/mdn/browser-compat-data/issues/3365?

If it does, what are the severity levels for "graceful" versus "ungraceful" degradation? Is it "warning" and "hint"? I think it would be nice to mention that in the doc, too, if we can agree on a level for it.

antross:
Thanks @Elchi3!

> Would this also help with mdn/browser-compat-data#3365?

Yes, without this change we'd be omitting notifying the user at all when a feature was marked for "graceful" degradation. We selectively do this today with a default "ignore" list that we're slowly adding to. 

I like the notion of still being able to inform the user where *all* features are actually supported, but with the appropriate severity.

> If it does, what are the severity levels for "graceful" versus "ungraceful" degradation? Is it "warning" and "hint"? 

Yes, we use "warning" today for anything we haven't ignored and would likely continue to do so. Most things we currently ignore would likely change to "hint" in the new scheme and we'd apply the same logic to any indication we get from MDN that a feature degrades gracefully. If we feel highly confident a feature may cause a compatibility issue I could see us upgrading some reports to "error" as well.

> I think it would be nice to mention that in the doc, too, if we can agree on a level for it.

@molant I agree these would be good to mention in the doc, maybe as some additional concrete examples of how we could apply the expanded set of severity levels. What do you think?

molant:
>  I agree these would be good to mention in the doc, maybe as some additional concrete examples

I'll work on adding this.

rfcs/severity.md
molant: I've added the 2 options to the RFC

molant:
@antross @Elchi3 I've added the `compat-api` examples to the doc.
Let me know if I missed anything!

molant:
Is there any more immediate feedback on this proposal or do we feel it's good enough to start coding something?

If there are no objections in the next few (working) days I'll close it and create a new Epic with the document above that we can update if needed.

molant:
Closing in favor of #3036

PR #2983, opened 2019-09-14
Chore: Validate size of VS Code extension bundle

### LINKED ISSUE #2955: Validate size of VS Code extension bundle
> Should we do something similar to the browser extension where we have a size budget?

Yes, I'll file an issue to track that.

_Originally posted by @antross in https://github.com/webhintio/hint/pull/2954#issuecomment-529633258_

This will help us ensure we don't add large dependencies unintentionally.

antross:
Fix #2955

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/extension-vscode/package.json
+++ b/packages/extension-vscode/package.json
@@ -16,6 +16,7 @@
     ],
     "timeout": "1m"
   },
+  "bundleSize": 75000,
   "categories": [
     "Linters"
   ],
@@ -61,6 +62,7 @@
     "rimraf": "^3.0.0",
     "sinon": "^7.4.2",
     "typescript": "^3.6.3",
+    "vsce": "^1.66.0",
     "vscode-languageclient": "^5.2.1",
     "vscode-languageserver": "^5.2.1",
     "webpack": "^4.39.3",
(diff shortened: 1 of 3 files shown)


packages/extension-vscode/package.json
```diff
@@ -16,6 +16,7 @@
     ],
     "timeout": "1m"
   },
+  "bundleSize": 75000,
```
molant: 😍

packages/extension-vscode/scripts/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
```
molant: Why do we need a new `tsconfig.json` under `scripts`? Is it for `checkJS`? If so, can't we just add it to the top of the file (or maybe add that option in the root of our project?)

antross: Yes, this is for `checkJS`. We've used this across a few of our packages now, so adding to the root of our project seems like the right approach going forward. I'll update the PR.

packages/extension-vscode/scripts/bundle-size.js
```diff
@@ -0,0 +1,20 @@
...
+const packageJSON = require('../package.json');
+
+const filename = `${path.resolve(__dirname, '..')}/vscode-webhint-${packageJSON.version}.vsix`;
+
+fs.stat(filename, (err, stats) => {
+    if (err) {
+        throw new Error(`Reading bundle failed: ${err}`);
+    }
+
+    const sizeKB = Math.round(stats.size / 1000);
```
sarvaje: Nit: Shouldn't this be `const sizeKB = Math.round(stats.size / 1024);`?

antross: Either that or a lower-case "k" according to wikipedia. https://en.wikipedia.org/wiki/Kilobyte

Honestly doesn't matter much in this case as we're using this just to keep tabs on bundle size growth. I went for consistency with what we were already doing for the browser extension.

sarvaje: hehehe, then `1000` is ok :)

packages/extension-vscode/scripts/tsconfig.json
molant: sgtm

molant:
looks like there are some merging conflicts. Otherwise this should be ready to merge.

PR #2993, opened 2019-09-17
Fix: Track `FirstRun` events in CI differently

### LINKED ISSUE #2962: [Bug] Don't log `FirstRun` when `isCI` is true
CI environments aren't typically going to preserve state between runs so if someone allows telemetry, we can end up sending `FirstRun` events on every run.

molant:
## Pull request checklist

Make sure you:

Send `CIRun` events when the user gives permission for telemetry instead
of `FirstRun` so it is easier to know when there are new users and how
many times it is used from CI.

Fix #2962

### DIFF
--- a/packages/hint/src/bin/hint.ts
+++ b/packages/hint/src/bin/hint.ts
@@ -12,33 +12,22 @@
  * ------------------------------------------------------------------------------
  */
 
-const tracking = (/--tracking[=\s]+([^\s]*)/i).exec(process.argv.join(' '));
+const telemetry = (/--tracking[=\s]+([^\s]*)/i).exec(process.argv.join(' '));
 
-import { configStore, appInsights } from '@hint/utils';
+import { appInsights } from '@hint/utils';
 
-const trackingEnv = process.env.HINT_TRACKING;
-let enableTracking;
+const telemetryEnv = process.env.HINT_TRACKING;
+let enableTelemetry;
 
-if (tracking) {
-    enableTracking = tracking[1] === 'on';
-} else if (trackingEnv) {
-    enableTracking = trackingEnv === 'on';
+if (telemetry) {
+    enableTelemetry = telemetry[1] === 'on';
+} else if (telemetryEnv) {
+    enableTelemetry = telemetryEnv === 'on';
 }
 
-if (typeof enableTracking !== 'undefined') {
-    if (enableTracking) {
-        const alreadyRun: boolean = configStore.get('run');
-        const configured = appInsights.isConfigured();
-
+if (typeof enableTelemetry !== 'undefined') {
+    if (enableTelemetry) {
         appInsights.enable();
-
-        if (!configured) {
-            if (!alreadyRun) {
-                appInsights.trackEvent('FirstRun');
-            } else {
-                appInsights.trackEvent('SecondRun');
-            }
-        }
     } else {
         appInsights.disable();
     }
--- a/packages/hint/src/lib/cli/analyze.ts
+++ b/packages/hint/src/lib/cli/analyze.ts
@@ -27,7 +27,7 @@ const { askQuestion, mergeEnvWithOptions } = misc;
 const { installPackages } = npm;
 const { cwd } = fs;
 const debug: debug.IDebugger = d(__filename);
-const configStoreKey: string = 'run';
+const alreadyRunKey: string = 'run';
 const spinner = ora({ spinner: 'line' });
 
 /*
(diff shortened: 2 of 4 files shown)


molant:
After syncing with @antross we've seen there are some inconsistencies with the `FirstRun` and `SecondRun` events being sent from multiple places.

I'll repurpose this PR to clean up a bit the code and make it more consistent.

molant:
This is ready for another review

packages/hint/src/bin/hint.ts
```diff
@@ -12,33 +12,22 @@
...
 
-const tracking = (/--tracking[=\s]+([^\s]*)/i).exec(process.argv.join(' '));
+const telemetry = (/--tracking[=\s]+([^\s]*)/i).exec(process.argv.join(' '));
 
-import { configStore, appInsights } from '@hint/utils';
+import { appInsights } from '@hint/utils';
 
-const trackingEnv = process.env.HINT_TRACKING;
-let enableTracking;
+const telemetryEnv = process.env.HINT_TRACKING;
```
sarvaje: Should we change also the env variable?

molant: There's an issue for that (#2995). I don't want to make a breaking change right now.

sarvaje: ok!

sarvaje:
You need to update the tests :)

molant:
> You need to update the tests :)

Yes, for whatever reason I had a comment saying that I'll update them once everyone was OK but I don't see it 🤔

packages/hint/src/bin/hint.ts
```diff
@@ -12,33 +12,22 @@
...
+let enableTelementry;
```
antross:
```suggestion
let enableTelemetry;
```

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -104,46 +104,64 @@ const pruneUserConfig = (userConfig: UserConfig) => {
...
+     * No need to check for CI as state is not persisted and
+     * `alreadyRun` will always be `false`.
+     */
+    if (alreadyRun) {
+        appInsights.trackEvent('SecondRun');
     }
 
-    appInsights.disable();
+    const prunedUserConfig = pruneUserConfig(userConfig);
+    const payload = { ...{ ci: isCI }, ...prunedUserConfig };
```
antross: No need to create an intermediate object here.

```suggestion
    const payload = { ci: isCI, ...prunedUserConfig };
```

molant:
I've done some changes. Now we don't have `FirstRun` nor `SecondRun`. I believe we were reporting `SecondRun` for all subsequent runs. I've added a new property that tells if it is been run before.

@hxlnt can you please take a look to make sure you are OK with these changes?

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -104,46 +104,55 @@ const pruneUserConfig = (userConfig: UserConfig) => {
...
+    if (!telemetryEnabled) {
         return;
     }
 
-    appInsights.disable();
+    appInsights.trackEvent('analyze', {
+        ...{
+            ci: isCI,
+            previouslyRun: alreadyRun
+        },
```
antross: Similar to my earlier feedback, there's no need to create an intermediate object that we just throw away:

```ts
    appInsights.trackEvent('analyze', {
        ci: isCI,
        previouslyRun: alreadyRun,
        ...pruneUserConfig(userConfig)
    });
```

packages/hint/src/lib/cli/analyze.ts
```diff
@@ -104,46 +104,55 @@ const pruneUserConfig = (userConfig: UserConfig) => {
...
 
+    if (!telemetryEnabled) {
         return;
     }
 
-    appInsights.disable();
+    appInsights.trackEvent('analyze', {
+        ...{
+            ci: isCI,
+            previouslyRun: alreadyRun
```
antross: I like having this switched to a property. 

It also got me thinking, for the browser and VS Code extensions we added a `*-telemetry` event (`f12-telemetry` or `vscode-telemetry` respectively) to report when a user opts-in to telemetry for the first time. Think we should do something similar in this PR or leave it as a separate task?

packages/hint/src/lib/cli/analyze.ts
molant: Done

packages/hint/src/lib/cli/analyze.ts
molant: I've renamed all the events to `cli-*` and added an opt-in event.

molant:
Ready for another round

PR #2996, opened 2019-09-17
New: Suggest alternate support when available

### LINKED ISSUE #2644: Update hint-compat-api to suggest using prefixed features
From self-hosting I've noticed a number of times hint-compat-api flags a feature like a CSS property or value it's because the CSS is missing a vendor prefixed version of the property needed for one of the target browsers. 

We could make fixing these a bit easier to fix by recognizing this situation and suggesting which prefix or prefixed versions to add.

antross:
## Pull request checklist

Make sure you:

Fix #2644

### DIFF
--- a/packages/hint-compat-api/src/css.ts
+++ b/packages/hint-compat-api/src/css.ts
@@ -9,19 +9,16 @@ import { HintContext } from 'hint/dist/src/lib/hint-context';
 import { IHint, ProblemLocation } from 'hint/dist/src/lib/types';
 import { StyleEvents } from '@hint/parser-css/dist/src/types';
 import { getUnsupportedDetails } from '@hint/utils/dist/src/compat';
-import { AlternateDetails, UnsupportedBrowsers } from '@hint/utils/dist/src/compat/browsers';
+import { UnsupportedBrowsers } from '@hint/utils/dist/src/compat/browsers';
 import { getCSSCodeSnippet } from '@hint/utils/dist/src/report';
 
-import { filterBrowsers, formatSupported, joinBrowsers } from './utils/browsers';
+import { formatAlternatives } from './utils/alternatives';
+import { filterBrowsers, joinBrowsers } from './utils/browsers';
 import { filterSupports } from './utils/filter-supports';
 import { resolveIgnore } from './utils/ignore';
 
 import meta from './meta/css';
 import { getMessage } from './i18n.import';
-
-type BrowserAlternateDetails = AlternateDetails & {
-    browser: string;
-};
 
 type ReportData = {
     feature: string;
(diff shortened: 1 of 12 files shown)


antross:
Some pre-built versions of the browser extension with these changes enabled:
* [Chromium webhint-0.1.0.zip](https://github.com/webhintio/hint/files/3623378/webhint-0.1.0.zip)
* [Firefox webhint-0.1.0.zip](https://github.com/webhintio/hint/files/3623380/webhint-0.1.0.zip)

image

packages/hint-compat-api/src/css.ts
```diff
@@ -223,6 +237,37 @@ const walk = (ast: ContainerBase, context: Context) => {
     reportUnsupported(reportsMap, context);
 };
 
+export const formatAlternates = (language: string, unsupported: UnsupportedBrowsers, formatFeature?: (name: string) => string): string[] => {
+    const groupedAlternates = new Map<string, BrowserAlternateDetails[]>();
```
molant: There's a lot going on in `formatAlternates` 😵
What do you think of splitting it into 2 more functions: 

* `groupAlternates`
* `createAlternateMessages`

And add some documentation with an example?

packages/hint-compat-api/src/css.ts
```diff
@@ -223,6 +237,37 @@ const walk = (ast: ContainerBase, context: Context) => {
...
+            group.push({ browser, ...details.alternate });
+            groupedAlternates.set(name, group);
+        }
+    }
+
+    const alternateMessages = new Set<string>();
+
+    for (const [name, list] of groupedAlternates) {
+        const formattedName = formatFeature ? formatFeature(name) : name;
+        const browsers = list.map(({ browser, versionAdded, versionRemoved }) => {
```
molant:
```suggestion
        const browsers = list.map(formatSupported);
```

packages/hint-compat-api/src/utils/browsers.ts
```diff
@@ -28,6 +28,34 @@ export const filterBrowsers = (browsers: string[]): string[] => {
     });
 };
 
+export const formatSupported = (browser: string, versionAdded?: string, versionRemoved?: string): string => {
+    const name = getFriendlyName(browser);
```
molant: What about renaming `browser` to `browserId` and `name` to `browser`?
I think it will make it easier to know how the message would look like. Same for `formatUnsupported`.

antross:
> How strong do people feel about `alternate` vs `alternative`?

Not strongly, and I agree with your point about the potential confusion with `alternate`. I'll switch this to `alternative`.

packages/hint-compat-api/src/css.ts
antross: That won't work unless I also update the signature for `formatSupported` to take an object (which I'd prefer not to do here).

packages/hint-compat-api/src/utils/browsers.ts
antross: How about keeping `browser` and renaming `name` to `browserName`?

packages/hint-compat-api/src/css.ts
antross: Looking at this, I think just splitting out a separate `groupAlternatives` function for that part of the work helps enough. Splitting both out would make this function largely pass-through which I don't think is needed.

packages/hint-compat-api/src/utils/alternatives.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import { formatSupported } from './browsers';
+
+export type BrowserAlternativeDetails = AlternativeDetails & {
+    browser: string;
+};
+
+/**
+ * Sort alternatives into groups based on those recommending the same
+ * alternate feature.
```
molant: Nit: `alternative` 😛

antross: Some habits are hard to break 😉

Fixed.

molant:
@antross anything else you want to add or can we get this merged after 2 OKs?

antross:
Nope, nothing to add so I went ahead and merged this.

PR #3004, opened 2019-09-18
New: Group hint reports with the same message text

### LINKED ISSUE #2970: Combine hint reports for same issue
Right now we can get the same "hint" multiple times but for different resources, different parts of the code, etc.
This isn't a great experience and can cause the user to miss another hint  after too much repetition. We should combine then (maybe in a nested details/summary?). This applies to the browser extension and the html formatter, although probably the browser has higher priority (IMO).

image

antross:
## Pull request checklist

Make sure you:

Also update layout of related controls to prefer controlling spacing
from parent controls instead of children. This makes child controls
easier to re-use without breaking layout.

- - - - - - - - - -

Fix #2970

### DIFF
--- a/packages/extension-browser/src/_locales/en/messages.json
+++ b/packages/extension-browser/src/_locales/en/messages.json
@@ -83,10 +83,6 @@
         "message": "Help us improve webhint by sending limited usage information (no personal information or URLs will be sent).",
         "description": "Label for telemetry opt-in popup"
     },
-    "hintCountLabel": {
-        "message": "hint $1:",
-        "description": "Label for a test result based on the order in which it is listed, e.g. 'hint 2:'."
-    },
     "inspectElement": {
         "message": "Inspect Element",
         "description": "Label for a button to link to a specific element in the Elements panel."
--- a/packages/extension-browser/src/devtools/views/controls/external-link.css
+++ b/packages/extension-browser/src/devtools/views/controls/external-link.css
@@ -3,14 +3,17 @@
 
     border-bottom: var(--border);
     color: var(--link-color);
+    padding-bottom: 1px;
     text-decoration: none;
 }
 
 .root:focus {
     border-bottom: 2px solid var(--base-color);
     outline: none;
+    padding-bottom: 0;
 }
 
 .root:global(:not(.focus-visible)) {
     border-bottom: var(--border);
+    padding-bottom: 1px;
 }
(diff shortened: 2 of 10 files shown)


antross:
Pre-built bundles for testing:
* [Chromium webhint-0.1.0.zip](https://github.com/webhintio/hint/files/3628498/webhint-0.1.0.zip)
* [Firefox webhint-0.1.0.zip](https://github.com/webhintio/hint/files/3628499/webhint-0.1.0.zip)

Screenshot:
image

antross:
Updated test packages with revised summary behavior:
* [Chromium webhint-0.1.0.zip](https://github.com/webhintio/hint/files/3628644/webhint-0.1.0.zip)
* [Firefox webhint-0.1.0.zip](https://github.com/webhintio/hint/files/3628645/webhint-0.1.0.zip)

Screenshot:

image

sarvaje:
I think it continue being a little bit confusing having a 2 for accessibility, and then there are 3 hints for it. I know that is 2 "rules" that are "failing", but it can be a little bit confusing.

antross:
> I think it continue being a little bit confusing having a 2 for accessibility, and then there are 3 hints for it. I know that is 2 "rules" that are "failing", but it can be a little bit confusing.

Fair point. @hxlnt and @molant what do you think? Should we leave this as-is for now, or consider making the count align with the sum of the "hints: x" summaries for each hint (which could be quite a bit larger)?

molant:
I'd leave as it is, I think it's already an improvement compared to what we have right now. We can maybe brainstorm a bit more until tomorrow noon and if we haven't come up with something better just merge it so we can ship v1.

Some ideas: remove the word "Hints:" from the results and just put the number or think of a better word (Issues?)


molant:
What about something like the following:

image

1. We keep the colored dots (we will have to add the passed one, I just scanned a website with 0 issues in PWA and there wasn't any color)
1. Remove the number of passed hints from there
1. In the title of the category tell how many hints have found issues (current copy is an example, happy to change it to whatever)
1. Replace `Hints:` with `Issues:`. This will be also more consistent with the current `No issues` message in the same place

Thoughts?

antross:
@molant How about we move this to an issue if we want to keep discussing refining the display since we've already agreed on what we're doing for this PR?

molant:
OK. Didn't know we have con sensus for this PR.

PR #3009, opened 2019-09-19
Fix: Close browser if there is an error navigating.

### LINKED ISSUE #3006: [Bug][Puppeteer] Catch exception in `collect` to close the browser.
The navigation to a website can throw an exception if it takes to long. If that happen, some systems will keep the browser process open even when the process is closed (IIRC this is Linux behavior).

We should catch the exception and close the browser manually.

sarvaje:
Fix #3006

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/connector-puppeteer/src/connector.ts
+++ b/packages/connector-puppeteer/src/connector.ts
@@ -402,38 +402,44 @@ export default class PuppeteerConnector implements IConnector {
     public async collect(target: URL): Promise<void> {
         await this.initiate(target);
 
-        await this._engine.emit('scan::start', { resource: target.href });
+        try {
+            await this._engine.emit('scan::start', { resource: target.href });
 
-        // TODO: Figure out how to execute the user tasks in here and when to subscribe to events
-        this.addListeners();
+            // TODO: Figure out how to execute the user tasks in here and when to subscribe to events
+            this.addListeners();
 
-        debug(`Navigating to ${target.href}`);
+            debug(`Navigating to ${target.href}`);
 
-        if (this._auth) {
-            if (typeof this._auth.user === 'string' && typeof this._auth.password === 'string') {
-                await this._page.authenticate({ password: this._auth.password, username: this._auth.user });
-                await this._page.goto(target.href, { waitUntil: this._waitUntil });
+            if (this._auth) {
+                if (typeof this._auth.user === 'string' && typeof this._auth.password === 'string') {
+                    await this._page.authenticate({ password: this._auth.password, username: this._auth.user });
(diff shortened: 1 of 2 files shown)


packages/connector-puppeteer/src/connector.ts
```diff
@@ -402,38 +402,44 @@ export default class PuppeteerConnector implements IConnector {
...
-        // TODO: Figure out how to execute the user tasks in here and when to subscribe to events
-        this.addListeners();
-
-        debug(`Navigating to ${target.href}`);
-
-        if (this._auth) {
-            if (typeof this._auth.user === 'string' && typeof this._auth.password === 'string') {
-                await this._page.authenticate({ password: this._auth.password, username: this._auth.user });
-                await this._page.goto(target.href, { waitUntil: this._waitUntil });
+        try {
```
molant: Why do we do the `try/catch` here? 
I'm looking at the jsdom and it doesn't seem like are are closing any resource automatically (even though the `close` does something). Also [`engine.close()` closes the connector](https://github.com/webhintio/hint/blob/53edf270f84ead765bb981345d5321568ac69142/packages/hint/src/lib/engine.ts#L286) and we are calling that method from [`analyzer.analyze()` in a `finally`](https://github.com/webhintio/hint/blob/53edf270f84ead765bb981345d5321568ac69142/packages/hint/src/lib/analyzer.ts#L236).

Are we doing something wrong somewhere that forces us to clean up after ourselves?

sarvaje: You are right, analyzer will call to `engine.close()`, but you are assuming that people is going to be using the node API. If that is not the case, `this._page.goto` can throw an exception if it timeouts.

If we assume that people is going to be using the node API, this PR can be closed :)

molant: I think we should assume people are using the node API.
`_nameOfFunction` are "private" and people shouldn't be using them.

@antross what do you think?

sarvaje: what do you mean with `_nameOfFunction`? this is inside `collect` which is public.

molant: I thought you were saying people could use `this._page.goto` which is private.

We should push as much as possible for people to use the node API. If we want to keep this change then we should open an issue to do similar changes to jsdom, etc.

sarvaje: > We should push as much as possible for people to use the node API.

I agree.

molant:
Going to close this per the discussion above about devs using the node API only.
@sarvaje if I missed anything let me know!

PR #3013, opened 2019-09-19
New: Add 'close' to Node API.

### LINKED ISSUE #3005: [Feature] Expose a 'close' method in the node API.
Currently, if there is an exception running the scan, the browser will keep alive even if the main process is closed (IIRC this is the Linux behavior).

This close method will close the browser manually before exiting the code.

sarvaje:
Fix #3005

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint/src/lib/analyzer.ts
+++ b/packages/hint/src/lib/analyzer.ts
@@ -60,6 +60,7 @@ const validateConnector = (configuration: Configuration) => {
  */
 export class Analyzer {
     private configuration: Configuration;
+    private engine!: Engine;
     private _resources: HintResources;
     private formatters: IFormatter[];
     private watch: boolean | undefined;
@@ -218,9 +219,9 @@ export class Analyzer {
                 throw new AnalyzerError(`Property 'content' is only supported in formatter local. Webhint will analyze the url ${url.href}`, AnalyzerErrorStatus.AnalyzeError);
             }
 
-            const engine = new Engine(this.configuration, this._resources);
+            this.engine = new Engine(this.configuration, this._resources);
 
-            this.configureEngine(engine, url.href, options);
+            this.configureEngine(this.engine, url.href, options);
 
             let problems: Problem[] | null = null;
 
@@ -228,12 +229,13 @@ export class Analyzer {
                 if (options.targetStartCallback) {
                     await options.targetStartCallback({ url: url.href });
                 }
-                problems = await engine.executeOn(url, { content: target.content });
+                problems = await this.engine.executeOn(url, { content: target.content });
             } catch (e) {
                 throw new AnalyzerError(e, AnalyzerErrorStatus.AnalyzeError);
             } finally {
-                // TODO: Try if this is executed.
-                await engine.close();
+                if (this.engine) {
+                    await this.engine.close();
+                }
             }
 
             if (options.targetEndCallback) {
@@ -269,6 +271,19 @@ export class Analyzer {
         for (const formatter of this.formatters) {
             await formatter.format(problems, options);
         }
+    }
+
+    /**
+     * Close the engine if a scan is still in progress.
+     * To avoid unexpected behavior, use this method
+     * only if `analyze` throws an unhandled exception.
+     */
+    public close(): Promise<void> {
+        if (this.engine) {
+            return this.engine.close();
+        }
+
+        return Promise.resolve();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


packages/hint/src/lib/analyzer.ts
```diff
@@ -60,6 +60,7 @@ const validateConnector = (configuration: Configuration) => {
  */
 export class Analyzer {
     private configuration: Configuration;
+    private engine!: Engine;
```
molant:
```suggestion
    private engine: Engine;
```

You are already checking it exists before using it so this shouldn't be needed I think.

packages/hint/tests/lib/analyzer.ts
```diff
@@ -279,6 +279,24 @@ test('If the target is an URL, it will analyze it', async (t) => {
...
+    const engineExecuteOnStub = sandbox.stub(engine, 'executeOn').resolves([]);
+    const engineCloseStub = sandbox.stub(engine, 'close').resolves();
+    const webhint = new Analyzer({}, {}, []);
+
+    await webhint.analyze(new URL('http://example.com/'));
+
+    await webhint.close();
+
+    t.true(engineExecuteOnStub.calledOnce);
+    t.true(engineCloseStub.calledTwice);
```
molant: Explain why it will be called twice because otherwise we will forget in a couple days :)

packages/hint/tests/lib/analyzer.ts
```diff
@@ -279,6 +279,24 @@ test('If the target is an URL, it will analyze it', async (t) => {
...
+test('.close() will close the engine', async (t) => {
+    const sandbox = t.context.sandbox;
+
+    const { Analyzer, engine } = loadScript(t.context);
+
+    const engineExecuteOnStub = sandbox.stub(engine, 'executeOn').resolves([]);
+    const engineCloseStub = sandbox.stub(engine, 'close').resolves();
+    const webhint = new Analyzer({}, {}, []);
+
+    await webhint.analyze(new URL('http://example.com/'));
```
molant:
```suggestion
    await webhint.analyze(new URL('https://example.com/'));
```

Just because everything should be https 😋

packages/hint/src/lib/analyzer.ts
sarvaje: `engine` is not initialized in the constructor, that's why we need it.

antross: If that's the case then we definitely shouldn't have this and should be verifying `this.engine` actually exists before using it. Use `private engine?: Engine` instead. You can clean up repeated access by temporarily assigning to a local `engine` variable and testing that (which will cause TypeScript to narrow the type).

packages/hint/tests/lib/analyzer.ts
```diff
@@ -272,11 +272,36 @@ test('If the target is an URL, it will analyze it', async (t) => {
...
+
+    await webhint.analyze(new URL('https://example.com/'));
+
+    await webhint.close();
+
+    t.true(engineExecuteOnStub.calledOnce);
+    /*
+     * Engine close will be called twice, one when the analysis finish
+     * and one more time when we call close.
+     * `close` is meant to be call when the is an unhandled exception
```
molant:
```suggestion
     * `close` is meant to be called when there is an unhandled exception
```

packages/hint/src/lib/analyzer.ts
sarvaje: That is what I usually do, but I so that code somewhere in the project and I thought it was the "right" way to do it :s

sarvaje: Fixed!

antross: hehe, we definitely have a few uses of `!` floating around I'd love to get rid of when we get the chance 😁

packages/hint/tests/lib/analyzer.ts
sarvaje: done!

packages/hint/src/lib/analyzer.ts
```diff
@@ -218,22 +219,23 @@ export class Analyzer {
...
                     await options.targetStartCallback({ url: url.href });
                 }
-                problems = await engine.executeOn(url, { content: target.content });
+                problems = await this.engine.executeOn(url, { content: target.content });
             } catch (e) {
                 throw new AnalyzerError(e, AnalyzerErrorStatus.AnalyzeError);
             } finally {
-                // TODO: Try if this is executed.
-                await engine.close();
+                if (this.engine) {
```
antross: I don't think this `if` is needed since `this.engine` is being directly assigned earlier in this function.

sarvaje: done!

Issue #3014, opened 2019-09-20
[Bug] Improve CSP error reporting experience on Firefox


wellington1993:
# 🐛 Bug report

## Description

Can't scan Github

## Environment

* __webhint version:__ 1.0.0 
* __Browser version:__ Mozilla/5.0 (X11; Linux x86_64; rv:71.0) Gecko/20100101 Firefox/71.0 
* __URL for which webhint failed:__ 
https://github.com/forketyfork/mtom-soap

## `webhint` configuration

<details>
<summary>webhint’s configuration</summary>

Categories:
* [x] Accessibility
* [x] Compatibility
* [x] Performance
* [x] Pitfalls
* [x] PWA
* [x] Security

Your target browsers:
* [x] Recommended settings
* [ ] Custom:  

Ignored resources:
* * [x] None
* * [ ] Different origin
* * [ ] Custom:  

</details>

## Debug output

Please include the content of the `Error details` section if an error message was displayed.

<details>
<summary>webhint’s Error details</summary>

```text
Cannot run analysis scripts. Page may have CSP restrictions.
evaluateInPage/</a<@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1608137
setTimeout handler*evaluateInPage/<@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1608129
evaluateInPage@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1608054
d/r/<@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1606469
setTimeout handler*r@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1606443
EventListener.handleEvent*d@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1606969
@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1089013
_/<@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1064922
_@moz-extension://3515716b-a6dd-47be-9b5a-c6e3389e3699/content-script/webhint.js:1:1065065
```

</details>

molant:
Unfortunately this is a known Firefox limitation.

[...]

PR #3020, opened 2019-09-23
Fix: Work around CSP restrictions in Firefox

antross:
## Pull request checklist

Make sure you:

This makes #3014 unnecessary by employing a workaround allowing webhint to analyze pages with CSP restrictions in Firefox. I still need to finish cleaning up the code and updating tests, but it works if anyone wants to give it a try. 

Accomplished by switching to use `browser.devtools.inspectedWindow.eval` for scripts which need to be run in the page (for all browsers). Testing shows this is not subject to CSP restrictions for inline scripts or eval, unlike the current implementation. Unfortunately this approach does not support `Promise`s directly, so we have to manage asynchronous results within the page and periodically poll to see if they're ready.

Test packages:
* [Chromium webhint-1.0.0.zip](https://github.com/webhintio/hint/files/3640622/webhint-1.0.0.zip)
* [Firefox webhint-1.0.0.zip](https://github.com/webhintio/hint/files/3640624/webhint-1.0.0.zip)


antross:
Turns out we don't have any scripts we run directly in the page which return `Promise`s. I'm going to remove support for now to simplify the code, but if we decide to re-add it here's the implementation: 
[inject.zip](https://github.com/webhintio/hint/files/3643355/inject.zip)

PR #3042, opened 2019-09-27
New: Add browser action

### LINKED ISSUE #3041: [Bug] Need to add instructions when clicking on the extension to let users know they should go to the dev tools
Clicking in the button that appears close to the address bar should tell the user to go to the dev tools and click on the hint tab. This is how it looks like right now:

image

molant:
## Pull request checklist

Make sure you:

Fix #3041


molant:
This is how it looks like

image

molant:
And the bundles for testing

[webhint-firefox-1.0.2.zip](https://github.com/webhintio/hint/files/3664722/webhint-firefox-1.0.2.zip)
[webhint-chromium-1.0.2.zip](https://github.com/webhintio/hint/files/3664723/webhint-chromium-1.0.2.zip)

@sarvaje can you please take a quick look at this? 🙏
@antross is already in a ✈

I'll be merging it as soon as I get home with the hopes of doing a release soon after and having it ready by Monday.

packages/extension-browser/src/manifest.json
```diff
@@ -9,8 +9,12 @@
   "icons": {
     "180": "icon.png"
   },
+  "browser_action": {
+      "default_title": "webhint",
+      "default_popup": "browser-action/browser-action.html"
+  },
   "default_locale": "en",
-  "description": "Run webhint in your browser.",
+  "description": "Webhint is a tool for developers that can help you find issues in different areas. Quickly test your website to see if there are any issues with accessibility, browser compatibility, security, performance and more.",
```
molant: I changed this so it doesn't look like:

image

image

packages/extension-browser/src/browser-action/browser-action.html
```diff
@@ -0,0 +1,30 @@
...
+        }
+        .logo {
+            width: 400px;
+        }
+    </style>
+</head>
+
+<body>
+    <img class="logo" src="logo.svg" alt="webhint">
+    <p>Open DevTools and navigate to the Hints tab to start using webhint.</p>
```
hxlnt: Can we add F12 just to be extra clear about how to open DevTools? e.g., "Open DevTools (F12)"

molant: From conversations I had the F12 seems to come from the IE days and not everybody knows what it is. E.g.: I'm used to open the tools via Ctrl+Shift+I

sarvaje:
This is what I see:
image

Looks like the logo is broken

packages/extension-browser/src/browser-action/browser-action.html
hxlnt: I'm OK with replacing (F12) with (Ctrl+Shift+I), as listed in Chrome/Edge context menu. I just think DevTools isn't going to be descriptive enough, as I see a lot of people call them things like "inspector," "elements tool," "source tool," "dev console," etc. This just makes it crystal clear to the user what we mean by DevTools and how to open them.

molant: But Ctrl+Shift+I is only if I'm on Windows... Other people might just do "Inspect element". There's no definitive name and having the image I think helps clarify where they should be looking at.
If they don't know what that is, they have probably download the wrong extension.

Some previous art from other extensions:

axe:
image

react:
image

molant:
Nice catch @sarvaje. svg weren't copy to the dist bundle.

New bundle here and code updated.
[webhint-1.0.2.zip](https://github.com/webhintio/hint/files/3664936/webhint-1.0.2.zip)

packages/extension-browser/package.json
```diff
@@ -132,7 +132,7 @@
     "lint:js": "eslint . --cache --ext js --ext md --ext ts --ext tsx --ignore-path ../../.eslintignore --report-unused-disable-directives",
     "lint:dependencies": "node ../../scripts/lint-dependencies.js",
     "lint:md": "node ../../scripts/lint-markdown.js",
-    "prebuild:assets": "cpx \"./src/**/*.{html,json,png}\" dist/bundle && cpx \"./src/*.svg\" dist/bundle",
+    "prebuild:assets": "cpx \"./src/**/*.{html,json,png,svg}\" dist/bundle && cpx \"./src/*.svg\" dist/bundle",
```
antross: This results in redundant SVG files in our bundle as we inline most of them in the CSS so we don't typically want to copy the files themselves (This is likely why you had to increase the bundle size so much). The one exception are `*.svg` files directly under `src` (as seen later in this line of code).

I'd suggest moving `logo.svg` directly under `src` and reverting this particular change.

packages/extension-browser/package.json
```diff
@@ -7,7 +7,7 @@
     "timeout": "1m"
   },
   "browserslist": "last 2 chrome versions, last 2 firefox versions",
-  "bundleSize": 645000,
+  "bundleSize": 660000,
```
antross: This should be able to be a bit smaller of an increase once we aren't copying redundant SVG files.

packages/extension-browser/package.json
molant: After doing this, the size is 652KB.

packages/extension-browser/package.json
molant: Will set it to 655KB

molant:
Last round of changes pushed. I'll be merging this shortly so I can send it to the store asap. If there are any further changes we can take care of them next week.

Thank you all!

PR #3063, opened 2019-10-02
Build: Improve build times by storing artifacts

### LINKED ISSUE #3062: Store artifacts online and download them automatically on build
Based on the research in #2982, the tasks that takes the most time and where contributors could benefit from is to store the content of the multiple `dist` folders and download them on demand.

The idea is to:

1. Use the assets generated in Azure DevOps when merging into master
   1. Zip them and upload them to a GitHub release "dist"
   1. The filename will be the hash of the commit
   1. This will only be run in CI for the latest node version
1. When running `yarn build` the script will get the current hash and check if the repo doesn't have any pending changes
1. Will try to download the file for that given hash
   1. If it is available then it will be unzipped
   1. If it isn't, then it will fallback to build from the sources

The above is only done when running `yarn build`, not `yarn test`. The reason is that we want to modify also how the testing is done to be a bit more "intelligent" and reduce times but that should be part of another PR.

molant:
## Pull request checklist

Make sure you:

Still making sure it works on [Azure Pipelines](https://dev.azure.com/webhint/webhint/_build/results?buildId=6513) so putting it as a draft

Fix #3062


molant:
Looks like the file got uploaded in the right place (https://github.com/webhintio/hint/releases/tag/dist).

This should be ready for feedback.

molant:
Also before merging we will have to add a GITHUB_TOKEN to the pipeline that has repo access

molant:
To test this PR I think you could clone it locally and run something like:

```
yarn clean
yarn build
cd packages/hint
yarn hint https://example.com
```

image

azure-pipelines/standard.yml
```diff
@@ -18,3 +18,14 @@ steps:
     displayName: 'Install dependencies'
   - script: yarn test
     displayName: 'Test'
+  - script: ./scripts/dist-management/pack-and-upload.sh
+    condition: >
```
antross: Showing my yaml ignorance: Is the `>` here just so the subsequent checks can be spread across multiple lines?

antross: Also probably worth a comment explaining why we have this `condition`. My interpretation is:
1. Ensure we only copy the built files once (we pick Linux on Node 12)
2. Ensure we only copy for changes to `master` (we pick `Schedule` runs for our `master` branch)

scripts/dist-management/download-dist.js
```diff
@@ -0,0 +1,98 @@
...
+    const { stdout: hash } = await execa('git', ['rev-parse', '--verify', 'HEAD']);
+
+    return hash;
+};
+
+const repoClean = async () => {
+    const { stdout } = await execa('git', ['status', '--short']);
+
+    if (stdout !== '') {
+        throw new Error('Repository is not clean');
```
antross: Should we allow this even if the repo is not clean? I could easily see someone making some edits before trying to build for the first time. Of course then we'd need some other way to verify we don't run this on subsequent builds...

scripts/test-all.js
```diff
@@ -107,11 +109,21 @@ const main = async () => {
     const references = loadReferences('tsconfig.json');
     const start = Date.now();
 
-    console.log(chalk.green.bold(`Testing all packages`));
+    console.log(chalk.green.bold(`${action} all packages`));
```
antross: Good change, should we rename this file to something like `build-or-test-all.js` or `run-all.js` while we're at it?

scripts/dist-management/upload-dist.js
```diff
@@ -0,0 +1,60 @@
...
+            'content-type': 'application/zip'
+        },
+        name,
+        url: 'https://uploads.github.com/repos/webhintio/hint/releases/20378720/assets{?name,label}'
+    });
+
+    console.log('File uploaded');
+};
+
+console.log('uploading');
```
antross: Nit: mind moving this inside the start of the `upload` method?

scripts/dist-management/upload-dist.js
```diff
@@ -0,0 +1,60 @@
...
+const path = require('path');
+const fs = require('fs');
+
+const Octokit = require('@octokit/rest');
+const throttling = require('@octokit/plugin-throttling');
+
+const Client = Octokit.plugin(throttling);
+
+const octokitOptions = {
+    auth: `token ${process.env.GITHUB_TOKEN}`, // eslint-disable-line
```
antross: Nit: I'd prefer to only disable the specific `eslint` rule we want to suppress here (same goes for other `eslint-disable-line` instances).

scripts/dist-management/download-dist.js
```diff
@@ -0,0 +1,98 @@
...
+
+        throw e;
+    }
+
+    try {
+        console.log(`Unzipping`);
+
+        await unzip(fileName);
+    } catch (e) {
+        console.error(`Couldn't unzip the contents. Please run "yarn build"`);
```
antross: Is `Please run "yarn build"` helpful here? That's the command that will ultimately have run this script in the first place so I think having this in the error message both here and on line 72 would just be confusing. The calling script will carry on and build from source anyway.

azure-pipelines/standard.yml
molant: > : Is the `>` here just so the subsequent checks can be spread across multiple lines?

Yup

scripts/dist-management/download-dist.js
molant: I debated doing this because we also have the `tsconfig.tsbuildinfo` files but decided against including it on this PR because we will have to do something a bit more clever such as:

* download the assets
* identify the packages that have been modified
* build those packages

We want to improve how we detect the packages that need to be built so I was planning on doing this (or similar) at the same time we take care of that.

scripts/test-all.js
molant: sgtm

scripts/dist-management/download-dist.js
antross: Fair enough, I'm okay tackling this separately - let's just make sure we track it. In the meantime even supporting this only on a clean build is a vast improvement compared the current state of the world.

molant: https://app.zenhub.com/workspaces/webhint-5bc4f59f4b5806bc2bed23d0/issues/webhintio/hint/2095#issuecomment-538697074

azure-pipelines/standard.yml
molant: Added a comment

molant:
Ready for another round

scripts/dist-management/download-dist.js
```diff
@@ -0,0 +1,98 @@
...
+
+        throw e;
+    }
+
+    try {
+        console.log(`Downloading revision "${hash}"`);
+
+        await download(fileName);
+    } catch (e) {
+        console.error(`Couldn't download revision. Maybe sources aren't available? Please run "yarn build"`);
```
antross:
```suggestion
        console.error(`Couldn't download revision. Maybe sources aren't available?`);
```

PR #3071, opened 2019-10-04
severityThreshold + respect user's configuration

molant:
## Pull request checklist

Make sure you:

It turns out that the new proposed format is kind of already supported but not documented.

We can use the shorthand version to declare hints:

```json
{
  "hints": [
    "hint-a",
    "hint-b",
    ["hint-a", { "config": "a" }]
  ]
}
```

On top of that we also have the option to prefix the hint with `-` to disable it (e.g. `-hint-a`), as well as `hint-a:off`. 

The option of `on|off` is still not implemented and right now I'm not sure we should with all the other options we have already supported. I'm thinking we should just remove some of the options (such as `hint-a:warn`, but keep `hint-a:off`) when we are ready for a breaking change.

This PR also: 

* adds the `information` and `hint` severity levels that match the ones in VS Code. 
* adds `severityThreshold` to the schema to let the user decide when the exit code should be different than 0
* makes sure that the user severity configuration per hint is respected. The reason to do this is so we can start adding the new severity to the reports without being a breaking change. If we follow this the only breaking change should be the configuration.

I haven't added any documentation yet because I wanted your feedback about this approach so please let me know :)

### DIFF
--- a/packages/hint/src/lib/config/config-schema.json
+++ b/packages/hint/src/lib/config/config-schema.json
@@ -3,7 +3,68 @@
     "title": "Configuration",
     "description": "A configuration file for hint",
     "type": "object",
-    "properties": {
+    "definitions": {
+        "errorlevelstring": {
+            "type": "string",
+            "enum": [
+                "off",
+                "hint",
+                "information",
+                "warning",
+                "error"
+            ]
+        },
+        "errorlevelnumber": {
+            "type": "number",
+            "enum": [
+                0,
+                1,
+                2,
+                3,
+                4
+            ]
+        },
+        "hintconfig": {
+            "type": "array",
+            "items": [
+                {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/errorlevelstring"
+                        },
+                        {
+                            "$ref": "#/definitions/errorlevelnumber"
+                        }
+                    ]
+                },
+                {
+                    "type": "object"
+                }
+            ],
+            "maxItems": 2,
+            "minItems": 1,
+            "uniqueItems": true
+        },
+        "hintobject": {
+            "type": "object",
+            "patternProperties": {
+                "^.+$": {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/errorlevelstring"
+                        },
+                        {
+                            "$ref": "#/definitions/errorlevelnumber"
+                        },
+                        {
+                            "$ref": "#/definitions/hintconfig"
+                        }
+                    ]
+                }
+            },
+            "minItems": 1,
+            "uniqueItems": true
+        },
         "connector": {
             "description": "The connector to use to gather information",
             "anyOf": [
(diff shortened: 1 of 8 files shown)


packages/hint/src/lib/hint-context.ts
```diff
@@ -125,7 +125,7 @@ export class HintContext<E extends Events = Events> {
 
     /** Reports a problem with the resource. */
     public report(resource: string, message: string, options: ReportOptions = {}) {
-        const { codeSnippet, element, severity } = options;
+        const { codeSnippet, element, severity = Severity.error } = options;
```
antross: Given our discussions thus far, I think we should use `Severity.warning` here since it will be the more common value.
```suggestion
        const { codeSnippet, element, severity = Severity.warning } = options;
```

packages/hint/src/lib/hint-context.ts
```diff
@@ -147,7 +147,7 @@ export class HintContext<E extends Events = Events> {
             location: position || { column: -1, line: -1 },
             message,
             resource,
-            severity: severity || this.severity,
+            severity: this.severity || severity,
```
antross: Am I misunderstanding or does this mean if a hint has a severity configured in `.hintrc` that it trumps any individual reports severity? If that's the case we at least need a way to specify `"default"` as a severity in the `.hintrc` file, though I think I'd prefer to put this back the way it was.

packages/hint/src/lib/config/config-schema.json
```diff
@@ -3,7 +3,66 @@
...
+                "warning",
+                "error"
+            ]
+        },
+        "errorlevelnumber": {
+            "type": "number",
+            "enum": [
+                0,
+                1,
+                2
```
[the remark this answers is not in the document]
molant: Correct

packages/hint/src/lib/hint-context.ts
molant: SGTM

packages/hint/src/lib/hint-context.ts
molant: I believe the way it is right now is not the proper way. The user has explicitly stated a severity error for a given hint and we should honor that. If they decide something is a warning and they get an error they will start scratching their heads (this has happened to me).

By really honoring the user's preference we can start adding default hint severity levels without breaking changes to those particular hints or the main package.
On top of that, the old and new schema can live together (with the expected user's behavior) while we transition everybody to the new schema. People will be able to upgrade smoothly without having to modify their hintrc file if they have one, get the alert message, and know what is coming down the road in case they want to start using the new format.

antross: Got it, but then don't we need to remove severity settings from our built-in configurations as part of this change or we'll always override any per-report severity?

molant: We will remove the severity from the config once we are ready and all the hints are updated.

(looks like GitHub doesn't answer in the thread when using email).

molant:
Taking this out of draft.
I believe we should add the new documentation once we do the breaking changes, in the meantime things will continue working as it is.
What do you think @antross, @sarvaje ?

molant:
FYI, need to work on updating formatters. Most of them only look for error/warning.

molant:
I'm thinking about doing the formatters in another PR and probably when we are going to break. We might want/need to rethink the output (e.g.: `html`, limit the "total XX found" messages, etc.).

Because the defaults are not being changed and we aren't documenting anything at the moment I think it's fine to do this.

What do you think?

molant:
@antross I've rebased this PR and created the new servicing branch.

Can you take a look at this and if everything is OK get it merged?

antross:
@molant - looks like there are some coverage failures in CI 😞

molant:
Yes, I started updating the formatters. I'll remove those changes and push asap.


molant:
I've dropped the latest 2 commits related to formatters so coverage should pass now 🤞

PR #3073, opened 2019-10-04
Fix: Improve performance of getElementByUrl

sorin-davidoi:
## Pull request checklist

Make sure you:

I've notice that this function is hot while scanning https://www.theguardian.com/international.

See the attached profile (search for `getElementByUrl`
in the Stack Chart tab)).

https://profiler.firefox.com/public/0630a907f132fea1773f16909ca659fb1172b3ae/stack-chart/?globalTrackOrder=9-10-0-1-2-3-4-5-6-7-8&hiddenGlobalTracks=1-2-3-5-0-4-10-9-6-7&implementation=js&localTrackOrderByPid=15967-1-2-0~16415-0~19009-0~16503-0~16439-1-0~18110-0~16624-0-1~16559-0-1~18576-0-1~&search=getElementByUrl&thread=10&v=4

The attached image shows the profile before this change (captured from the Firefox Addon). ~I can't provide a profile taken after this change since I wasn't able to find information about running the extension locally.~
image

Profile with all the optimizations applied:

image

https://profiler.firefox.com/public/5a7cbd52513d82324676ade79a27fb7493934e34/stack-chart/?globalTrackOrder=8-0-1-2-3-4-5-6-7&hiddenGlobalTracks=8-0-1-2-3-4-5-6&hiddenLocalTracksByPid=9871-0-1&implementation=js&localTrackOrderByPid=15967-1-2-0~16415-0~19938-0~29276-0~1657-0~16439-1-0~16559-0-1~9871-0-1~&search=getElementByUrl&thread=9&v=4

### DIFF
--- a/packages/utils/src/dom/get-element-by-url.ts
+++ b/packages/utils/src/dom/get-element-by-url.ts
@@ -63,7 +63,6 @@ export const getElementByUrl = (dom: HTMLDocument, url: string): HTMLElement | n
             if (relativeUrl === null) {
                 continue;
             }
-            // TODO: Cache the absolute URL, so we don't run new URL() for the same URL.
 
             /**
              * If `elementUrls` has an element with an invalid value
--- a/packages/utils/src/dom/get-element-by-url.ts
+++ b/packages/utils/src/dom/get-element-by-url.ts
@@ -6,7 +6,7 @@ import { HTMLDocument, HTMLElement } from './html';
  * Two level cache: doc.base / url
  * TODO: Use quick-lru so that it doesn't grow without bounds
  */
-const CACHED_URLS: { [key: string]: { [key: string]: URL } } = { };
+const CACHED_URLS: Map<string, Map<string, URL>> = new Map();
 
 const getSrcsetUrls = (srcset: string | null): string[] => {
     if (!srcset) {
@@ -35,13 +35,13 @@ const getSrcsetUrls = (srcset: string | null): string[] => {
  * @param {string} url - URL that the element has to contain.
  */
 export const getElementByUrl = (dom: HTMLDocument, url: string): HTMLElement | null => {
-    const elements = dom.querySelectorAll('a[href],link[href],audio[src],iframe[src],script[src],source[src],track[src],frame[src],img[src],video[poster],img[srcset],source[srcset]');
+    const elements = dom.querySelectorAll('a[href],audio[src],frame[src],img[src],img[srcset],iframe[src],link[href],script[src],source[src],source[srcset],track[src],video[poster]');
 
-    if (!(dom.base in CACHED_URLS)) {
-        CACHED_URLS[dom.base] = { };
+    if (!CACHED_URLS.has(dom.base)) {
+        CACHED_URLS.set(dom.base, new Map());
     }
 
-    const urlCache = CACHED_URLS[dom.base];
+    const urlCache = CACHED_URLS.get(dom.base) as Map<string, URL>;
 
     /*
      * Even if there are multiple elements with the same URL,
@@ -71,11 +71,11 @@ export const getElementByUrl = (dom: HTMLDocument, url: string): HTMLElement | n
              */
             try {
 
-                if (!(relativeUrl in urlCache)) {
-                    urlCache[relativeUrl] = new URL(relativeUrl, dom.base);
+                if (!urlCache.has(relativeUrl)) {
+                    urlCache.set(relativeUrl, new URL(relativeUrl, dom.base));
                 }
 
-                const { href } = urlCache[relativeUrl];
+                const { href } = urlCache.get(relativeUrl) as URL;
 
                 if (href === url) {
                     return element;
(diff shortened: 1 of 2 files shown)


antross:
Thanks for the PR @sorin-davidoi!

Looks like our linter is complaining about a couple of things (which is why CI is failing). You should be able to see these locally by running `yarn lint` from `packages/extension-browser`.

Apologies for the difficulty running the extension locally. Once you've built the packages according to the instructions in [packages/extension-browser/CONTRIBUTING.md](https://github.com/webhintio/hint/blob/master/packages/extension-browser/CONTRIBUTING.md), you can side-load them in Firefox using the following steps:
1. Go to `about:debugging` via the address bar
2. Click "This Firefox"
3. Click "Load Temporary Add-on..."
4. If you did a debug build (via `yarn build`)
    1. Choose `packages/extension-browser/dist/bundle/manifest.json`
5. If you did a release build (via `yarn build-release`)
    1. Choose `packages/extension-browser/dist/firefox/webhint-1.0.3.zip`

I went ahead and built a release package for Firefox using your change if you want to just try that:
* [webhint-1.0.3.zip](https://github.com/webhintio/hint/files/3691531/webhint-1.0.3.zip)

sorin-davidoi:
@antross Thanks for the tips, fixed the linting errors! The profile shows an improvement, though quite small - that's because I've failed to notice that most of the time is spent in the `css-select` package. Nevertheless, I still think these changes are worth merging.

packages/utils/src/dom/get-element-by-url.ts
```diff
@@ -30,15 +30,28 @@ const getSrcsetUrls = (srcset: string): string[] => {
  */
 export const getElementByUrl = (dom: HTMLDocument, url: string): HTMLElement | null => {
     // TODO: Cache dom.querySelectorAll?.
-    const elements = dom.querySelectorAll('[href],[src],[poster],[srcset]').filter((element: any) => {
+    const elements = dom.querySelectorAll('[href],[src],[poster],[srcset]');
```
sorin-davidoi: Could we maybe restrict the selector? Something like `a[href],link[href],img[src],video[poster],img[srcset],source[srcset]` This should substantially reduce the number of calls to `hasAttrib` (which seems to be the bottleneck).

antross: Yes, I like the notion of scoping the selector to include element names - that should definitely help a bit. The only catch is to make sure we don't miss any. 

Looks like we'd need to include `audio[src]`, `iframe[src]`, `script[src]`, `source[src]`, and `track[src]` as well (and possibly `frame[src]` if we really want to be inclusive, though that's deprecated). I think we'd be okay omitting `applet`, `embed`, `frame`, and `object`, though.

sorin-davidoi: Applied this change and some further optimizations (attached new profile and screenshot to the pull request description). Looks like all the changes combined lead to a ~4x improvement.

molant: > Looks like all the changes combined lead to a ~4x improvement.

😍

packages/utils/src/dom/get-element-by-url.ts
```diff
@@ -2,7 +2,13 @@ import { URL } from 'url';
 
 import { HTMLDocument, HTMLElement } from './html';
 
-const getSrcsetUrls = (srcset: string): string[] => {
+/**
+ * Two level cache: doc.base / url
+ * TODO: Use quick-lru so that it doesn't grow without bounds
+ */
+const CACHED_URLS: { [key: string]: { [key: string]: URL } } = { };
```
antross: This is great! One request: I'd prefer to use nested `Map`s here unless there's a particular reason you didn't. [Browser support for it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility) is plenty good enough for what we're targeting. I think that should give us a slight additional performance edge too (though probably not a significant one).

packages/utils/src/dom/html.ts
```diff
@@ -10,6 +10,9 @@ import { INamedNodeMap } from '../types/html';
 
 import { DocumentData, ElementData } from '../types/snapshot';
 
+// TODO: Use quick-lru so that it doesn't grow without bounds
+const CACHED_CSS_SELECTORS: { [key: string]: cssSelect.CompiledQuery } = { };
```
antross: Let's also use a `Map` here.

packages/utils/src/dom/get-element-by-url.ts
```diff
@@ -29,45 +35,55 @@ const getSrcsetUrls = (srcset: string): string[] => {
  * @param {string} url - URL that the element has to contain.
  */
 export const getElementByUrl = (dom: HTMLDocument, url: string): HTMLElement | null => {
-    // TODO: Cache dom.querySelectorAll?.
-    const elements = dom.querySelectorAll('[href],[src],[poster],[srcset]').filter((element: any) => {
+    const elements = dom.querySelectorAll('a[href],link[href],audio[src],iframe[src],script[src],source[src],track[src],frame[src],img[src],video[poster],img[srcset],source[srcset]');
```
antross: Minor request: would you mind sorting these alphabetically? It will make it easier to see anything we might have missed in the future. Especially having items like `source[src]` and `source[srcset]` next to each other. Thanks!

molant:
This PR was failing for an unrelated reason that has already been addressed in master. I'm re-running it.

PR #3075, opened 2019-10-05
 New: Report active days in a rolling 28-day period

### LINKED ISSUE #3056: [Feature] Include retention data as part of telemetry
We'd like to be able to calculate a rolling 28-day retention metric as part of understanding the feature health of the browser extension. Specifically the goal is to understand how many users were active in the last 28 days (used the extension at least once), how many users were more engaged in the last 28 days (used the extension at least two different days), and how those metrics change over time.

Normally this would be done by counting unique user IDs over a given time period, but we've chosen not to have user IDs in our telemetry pipeline. This is because we only want to measure aggregate data, not individual user data. 

This calculation might be possible without user IDs if we include a client-maintained 28-day retention map in telemetry submissions, but the resulting queries could get complex. Probably worth further discussion to see if we want to go down that route or consider user IDs scoped to this scenario.

antross:
## Pull request checklist

Make sure you:

Fix #3056

### DIFF
--- a/packages/extension-browser/src/devtools/utils/activity.ts
+++ b/packages/extension-browser/src/devtools/utils/activity.ts
@@ -1,0 +1,69 @@
+export type Activity = {
+    [key in keyof {
+
+        /**
+         * 28-entry list of 1 (active) and 0 (inactive) days.
+         * (e.g. `"1001100110011001100110011001"`)
+         */
+        last28Days: string;
+
+        /**
+         * Date in ISO 8601 format (toISOString) with time at zero.
+         * (e.g. `"2019-10-04T00:00:00.000Z"`)
+         */
+        lastUpdated: string;
+
+    }]: string;
+};
+
+const getISODateString = () => {
+    const date = new Date(Date.now());
+
+    date.setUTCHours(0);
+    date.setUTCMinutes(0);
+    date.setUTCSeconds(0);
+    date.setUTCMilliseconds(0);
+
+    return date.toISOString();
+};
+
+const getDaysBetweenUpdates = (currentUpdate: string, lastUpdated?: string) => {
+    if (!lastUpdated) {
+        return 1;
+    }
+
+    const deltaMS = new Date(currentUpdate).getTime() - new Date(lastUpdated).getTime();
+
+    return deltaMS / 1000 / 60 / 60 / 24;
+};
+
+const initializeActivity = (): Activity => {
+    const lastUpdated = '';
+    const last28Days = ''.padEnd(28, '0');
+
+    return { last28Days, lastUpdated };
+};
+
+/**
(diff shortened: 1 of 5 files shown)


packages/hint/docs/user-guide/telemetry/summary.md
```diff
@@ -12,6 +12,7 @@ following information:
 
 * The configuration used.
 * Scan status of individual hints: `"passed"|"failed"|"fixing"|"fixed"`.
+* A log of days used over the past 28 days.
```
molant: What do you think about linking to https://github.com/webhintio/hint/issues/3056#issuecomment-538142964 here so people can look and confirm that we aren't using any type of id?

packages/extension-browser/src/devtools/utils/activity.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    const deltaMS = new Date(currentUpdate).getTime() - new Date(lastUpdated).getTime();
+
+    return deltaMS / 1000 / 60 / 60 / 24;
+};
+
+const initializeActivity = (): Activity => {
+    const lastUpdated = '';
+    const last28Days = new Array(28)
```
sarvaje: what about `''.padEnd(28, 0)` instead of the `new Array` thing?

packages/extension-browser/src/devtools/utils/activity.ts
```diff
@@ -0,0 +1,79 @@
...
+        return null;
+    }
+
+    activity.last28Days = [
+
+        // Today's activity.
+        1,
+
+        // Inactivity for past `delta - 1` days (maxing out at 28 - 1).
+        ...new Array(Math.min(delta - 1, 28 - 1)).fill(0),
```
sarvaje: you can use the 'padEnd' method here too.

packages/extension-browser/src/devtools/utils/activity.ts
antross: 😍

packages/hint/docs/user-guide/telemetry/summary.md
antross: Good idea - I'll add that.

molant:
@antross looks like the size of the bundle has been increased by [1KB in the build](https://dev.azure.com/webhint/webhint/_build/results?buildId=6636&view=logs&jobId=729d4de1-20bd-5332-288f-14ad5932d244&taskId=00522b24-a919-5b70-7116-ad06d8a33c92&lineStart=1690&lineEnd=1691&colStart=1&colEnd=1).
Maybe a dependency is getting bigger?

antross:
Heh, I'm guessing we were right on the edge before. Probably a few more characters is all it took.

Issue #3057, opened 2019-10-01
Review extension browser `devDependencies`


molant:
We might have more things that needed (`jsdom`?)

antross:
Looks like it. I definitely don't think `jsdom` should be there...

molant:
This was done a few days ago via #3104

PR #3104, opened 2019-10-09
Fix/dependency bloat

### LINKED ISSUE #3103: Create script to detect unused dependencies
We have a script that detects if we have a missing dependency on our `package.json` but we don't have anything that goes in the opposite sense (check `package.json` and make sure everything is used).

molant:
## Pull request checklist

Make sure you:

Fix #3103 and also improves the detection of missing dependencies in the previous code.

And fix the reports from the linting :)


molant:
No idea how CI is passing on Windows and not macOS or Linux 🤔

molant:
Because `src/shared/hint.import.ts` is generated on build this is failing.

We have 2 possibilities:

* Add this file to the repo and commit it every time there's a change (shouldn't be too often but it is yet another file to track)
* Ignore all of those packages. This could blind us sometimes although it seems like the packages that we usually forget are the `parser`s

@sarvaje @antross thoughts about this?

molant:
This should be ready for review. It also takes care of #3057

packages/connector-jsdom/package.json
```diff
@@ -9,9 +9,9 @@
   "dependencies": {
     "@hint/utils": "^5.0.2",
     "@hint/utils-connector-tools": "^4.0.12",
+    "@types/tough-cookie": "^2.3.5",
```
sarvaje: I think this should be a devDependency

molant: We export it via our types so we need it as a `dependency` to make it work correctly.

scripts/prebuild.js
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Searches all the `package.json` under `packages` and returns
+ * an array with the path and the content of each one.
+ * @returns { {packagePath: string, content: string}[] } The packages information
+ */
+const getPackages = async () => {
+    const packagesPaths = await globby([
+        `packages/*/package.json`,
+        `!packages/**/dist/**/*`,
+        `!packages/**/node_modules/**/*` // needed when we are inside a package like extension-vscode
```
antross: Now that we use `webpack` with `extension-vscode` we can probably remove `nohoist` from our root `package.json` and avoid this particular exception.

scripts/prebuild.js
```diff
@@ -0,0 +1,98 @@
...
+    const packages = packagesPaths.map((packagePath) => {
+        try {
+            const content = require(packagePath);
+
+            return {
+                content,
+                packagePath
+            };
+        } catch (e) {
+            // This happens if we are creating a new package and switching branches. We can ignore
```
antross: Is there a specific error message we could check for here? E.g. `Cannot find module '${packagePath}'`? I could see this potentially getting triggered due to a file permissions issue as well in which case we wouldn't want to ignore.

packages/hint-no-vulnerable-javascript-libraries/scripts/pack-snyk.js
```diff
@@ -62,7 +63,7 @@ fs.readFile(require.resolve('js-library-detector'), 'utf8', (err, data) => {
 
     filterSnykData(getDetectablePackages(data), snyk.npm);
 
-    const filename = path.resolve('./dist/src/snyk-snapshot.json');
+    const filename = path.resolve(snykSnapshotPath);
```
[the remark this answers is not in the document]
molant: > Is that your intent?

Yup, also I was having issues with non existing paths and such when running the script.

>  I bet this fixes the random issue we have where `extension-browser` bundles the non-compressed version of the snapshot...

🤞🤞🤞

scripts/prebuild.js
molant: In my case I wanted to ignore the error and continue the execution, that's why I didn't say anything. If we don't want to hide it, then let's remove the `try/catch`. The exception was clear enough to know what was going on.

scripts/prebuild.js
antross: Because of a change in another PR where we're no longer bundling the VS Code extension with webpack during test runs we'll actually still need this.

PR #3109, opened 2019-10-11
New: Add support for JSX

### LINKED ISSUE #2889: Support JSX files
We would like to run the HTML hints, not only the JS ones, when parsing this type of files.

antross:
## Pull request checklist

Make sure you:

Fix #2889 

* Allows JS hints such as `hint-create-element-svg` to run against `*.jsx` files.
* Also allows HTML hints such as `hint-button-type` to run against `JSX` content.
* Introduces the concept of fragment documents and expressions to represent template content.
* Strips JSX libraries when bundling `extension-browser`

### DIFF
--- a/packages/parser-javascript/src/parser.ts
+++ b/packages/parser-javascript/src/parser.ts
@@ -1,229 +1,71 @@
-import { Options, parse, tokenizer } from 'acorn';
-import * as acornWalk from 'acorn-walk';
-import * as ESTree from 'estree';
+import { Options, Parser } from 'acorn';
+import { Node } from 'estree-jsx';
 
+import { debug as d } from '@hint/utils/dist/src/debug';
 import { HTMLElement } from '@hint/utils/dist/src/dom';
-import * as logger from '@hint/utils/dist/src/logging';
 import { determineMediaTypeForScript } from '@hint/utils/dist/src/content-type';
 import { ElementFound, FetchEnd, Parser as WebhintParser } from 'hint/dist/src/lib/types';
 import { Engine } from 'hint/dist/src/lib/engine';
 
-import { ScriptEvents, Walk, NodeVisitor } from './types';
+import { ScriptEvents } from './types';
+import { combineWalk } from './walk';
 
 export * from './types';
 
-type Key = {
-    node: ESTree.Node;
-    base?: NodeVisitor;
-    state?: any;
-};
-
-/**
- * A WalkArray is a pair of Key and a Map with
- * all the callbacks (for methods full and fullAncestor)
- * or all the NodeVisitor (for methods simple and ancestor) for that Key.
- *
- * The key is the same if the root node, the base and the state
- * you call walk.(simple|ancestor|full|fullAncestor) is the same.
- *
- * In case of method used is `full` or `fullAncestor`, the map will
- * have an only key `callbacks`.
- */
-type WalkArray = Array<[Key, Map<keyof NodeVisitor | 'callbacks', Function[]>]>;
+const debug = d(__filename);
+const jsx = require('acorn-jsx');
+const jsParser = Parser.extend();
+const jsxParser = jsx ? Parser.extend(jsx()) : Parser.extend();
 
 export default class JavascriptParser extends WebhintParser<ScriptEvents> {
     public constructor(engine: Engine<ScriptEvents>) {
         super(engine, 'javascript');
 
         engine.on('fetch::end::script', this.parseJavascript.bind(this));
+        engine.on('fetch::end::unknown', this.onFetchUnknown.bind(this));
         engine.on('element::script', this.parseJavascriptTag.bind(this));
     }
 
-    private getCurrentVisitorsOrCallback(walkArray: WalkArray, node: ESTree.Node, base?: NodeVisitor, state?: any) {
-        const item = walkArray.find(([key]) => {
-            return key.node === node && key.base === base && key.state === state;
-        });
-
-        return item ? item[1] : null;
-    }
-
-    private async emitScript(sourceCode: string, resource: string, element: HTMLElement | null) {
+    private async emitScript(parser: typeof Parser, sourceCode: string, resource: string, element: HTMLElement | null) {
         try {
             await this.engine.emitAsync(`parse::start::javascript`, { resource });
 
-            const options: Options = { locations: true };
-            const ast = parse(sourceCode, options) as ESTree.Node;
-            const tokens = [...tokenizer(sourceCode, options)];
-            const defaultCallbacksProperty = 'callbacks';
+            const options: Options = { locations: true, ranges: true };
+            const ast = parser.parse(sourceCode, options) as Node;
+            const tokens = [...parser.tokenizer(sourceCode, options)];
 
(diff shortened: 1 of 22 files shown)


packages/parser-javascript/tests/fixtures/example.jsx
```diff
@@ -0,0 +1,14 @@
+let circle = document.createElement('circle');
```
antross: Note to self: Remove this file. I was using it to manually verify the end-to-end experience in the VS Code extension. Separate tests capturing individual conditions here are now part of `parser-jsx/tests`.

packages/parser-jsx/tests/tests.ts
```diff
@@ -0,0 +1,157 @@
...
+    return { engine, jsParser, jsxParser };
+};
+
+const parseJSX = (content: string) => {
+    const { engine } = mockContext();
+
+    return emitFetchEndJSX(engine, content);
+};
+
+test.serial('It translates basic JSX to HTML', async (t) => {
```
antross:
```suggestion
test('It translates basic JSX to HTML', async (t) => {
```

packages/utils/src/content-type.ts
```diff
@@ -452,6 +452,7 @@ const getType = (mediaType: string) => {
     switch (mediaType) {
         case 'application/javascript':
         case 'text/javascript':
+        case 'text/jsx':
             return 'script';
         case 'text/css':
```
antross: I added `text/jsx` to map to `script` above so we get `fetch::end::script` events for `*.jsx` files. However we don't currently do something similar for `*.sass` and `*.less` files to map them to `fetch::end::css` (probably would have if it had been `fetch::end::style`) so they map to `fetch::end::unknown`.

I feel we probably should make these consistent one way or the other. Thoughts?

packages/parser-javascript/src/estree-jsx.ts
```diff
@@ -0,0 +1,120 @@
...
+    Literal,
+    Node as ESTreeNode,
+    SpreadElement
+} from 'estree';
+
+export * from 'estree';
+
+type Omit<Props, T> = Pick<T, Exclude<keyof T, Props>>;
+
+export type Node =
```
antross: If we're happy with this file I can try contributing it to DefinitelyTyped as it should be complete. I'll likely re-work the existing `@types/estree` package slightly to export a few private types like `BaseNode` so they can be re-used instead of copied like I had to do here.

packages/utils/src/content-type.ts
molant: I'm torn apart with this.
On one side I think that mapping to `unknown` is a bit obscure for developers. On the other side we will have to add a bunch of things to `utils` and thus release everything again.

If we decide to add more mappings I think we should try add as many in one go as possible, maybe even some that don't have a parser yet.

packages/parser-javascript/src/estree-jsx.ts
molant: We should contribute upstream. Let's hope they don't take too much time to approve the PR.

packages/utils/src/content-type.ts
antross: Yeah, I'm having the same internal conflict. 

Personally I'd almost like to make the events themselves flexible so we automatically have `fetch::end::jsx`, etc. based on content type, but I wouldn't be able to type it correctly. I think I'm okay keeping just JSX special as we intend to parse it with our JS parser to extract normal JS content contained within (whereas LESS and SASS require fully dedicated parsers).

Another thought is having a generic `fetch::end::text` event for all `text/*` content types that don't map to something more specific (slightly better than `unknown`).

antross: Given we don't want to make a habit of this, I decided to revert `text/jsx` to `unknown` for now. Plus that was the only reason left this PR was touching the `utils` package and it didn't seem worth it.

antross:
@molant @sarvaje This is ready for a final round of review. CI failures are infrastructure related at the moment. Tests should be passing once service issues are resolved.

molant:
Looks like everything is passing! There's only a duplicate of the macos run for whatever reason 🎉

packages/parser-javascript/src/parser.ts
```diff
@@ -1,229 +1,70 @@
...
+                    resource,
+                    sourceCode,
+                    tokens,
+                    walk
                 });
             });
+
         } catch (err) {
-            logger.error(`Error parsing JS code: ${sourceCode}`);
+            logger.error(`Error parsing JS code (${err}): ${sourceCode}`);
```
molant: Do we want to do this via `logger` or `debug`? It's not like the execution is being interrupted so I'm more inclined to use `debug`.

packages/parser-javascript/src/walk.ts
```diff
@@ -0,0 +1,222 @@
...
+
+            if (typeof visitorsOrCallback === 'function') {
+                // `full` and `fullAncestor` only track an array of callbacks.
+                const name = defaultCallbacksProperty;
+                const visitorCallbacks = currentVisitors.get(name) || [];
+
+                visitorCallbacks.push(visitorsOrCallback);
+                currentVisitors.set(name, visitorCallbacks);
+
+                return Promise.resolve();
```
molant: Can't we just do `return`? The code doesn't seem to return any "real"  `Promise`  🤔
Probably worth a comment if there's a particular reason.

antross: Yep, this can just be a `return` (as can the similar `return` at the end of this function). This was left over from a different experiment with registering post-walk callbacks.

packages/parser-javascript/src/walk.ts
```diff
@@ -0,0 +1,222 @@
...
+            for (const [name, callback] of Object.entries(visitorsOrCallback)) {
+                // `ancestor` and `simple` track an array of NodeVisitors.
+                const mapName = name as keyof NodeVisitor;
+                const visitorCallbacks = currentVisitors.get(mapName) || [];
+
+                visitorCallbacks.push(callback!);
+                currentVisitors.set(mapName, visitorCallbacks);
+            }
+
+            return Promise.resolve();
```
antross: Note to self: remove this.
```suggestion
```

packages/parser-javascript/src/parser.ts
antross: I can change this to `debug`, though I'll admit I'm not clear on when we prefer using one over the other.

molant: The undocumented feedback in my head is:

1. Use `logger` if you want the user to see it 
1. Use `debug` if we (devs of webhint) want to see it when running `--debug`

Unless it's an error that's going to break the execution or that can help the user figure out why something isn't working (generally) I tend to use `debug`.

One thing we could do in the future is to generate a `.log` file automatically (without `--debug`) somewhere. If something fails then we point the user there to look into it or upload it into an issue.

molant:
@antross this has been approved. Do you need to do some rebasing here?

antross:
Yeah, I need to splice in the last commit. Once that's up-to-date and CI passes I'll go ahead and merge.

PR #3123, opened 2019-10-15
Fix: Location for json errors

### LINKED ISSUE #2037: [Bug] Bad location for compilerOptions reports in tsconfig rules
image

### LINKED ISSUE #2222: [Feature] Point to `compilerOptions` in tsconfig.json for missing options
Currently missing options like `compilerOptions.importHelpers` just point to the start of the file. If `compilerOptions` is present, pointing to it would provide more local and useful feedback.

This is most noticeable in the vscode extension.

sarvaje:
Fix #2222
Fix #2037

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/utils/src/json-parser.ts
+++ b/packages/utils/src/json-parser.ts
@@ -1,6 +1,6 @@
 import { findNodeAtLocation, Node, parse, parseTree, Segment } from 'jsonc-parser';
 import { IJSONLocationOptions, IJSONResult } from './types/json-parser';
-import { ProblemLocation} from './types/problem-location';
+import { ProblemLocation } from './types/problem-location';
 
 const rxIsNumber = /^[0-9]+$/;
 
@@ -27,7 +27,13 @@ class JSONResult implements IJSONResult {
 
     public getLocation(path: string, options?: IJSONLocationOptions): ProblemLocation | null {
         const segments = this.pathToSegments(path);
-        const node = findNodeAtLocation(this._root, segments) || null;
+        let node = null;
+
+        while (!node && segments.length > 0) {
+            node = findNodeAtLocation(this._root, segments) || null;
+
+            segments.pop();
+        }
 
         /**
          * The node isn't in the current file. Use alternative path if provided. This happens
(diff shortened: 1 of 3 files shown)


molant:
@sarvaje tests related to `extends` seem to be failing

packages/utils/src/json-parser.ts
```diff
@@ -27,7 +27,13 @@ class JSONResult implements IJSONResult {
...
     public getLocation(path: string, options?: IJSONLocationOptions): ProblemLocation | null {
         const segments = this.pathToSegments(path);
-        const node = findNodeAtLocation(this._root, segments) || null;
+        let node = null;
+
+        while (!node && segments.length > 0) {
+            node = findNodeAtLocation(this._root, segments) || null;
+
+            segments.pop();
+        }
```
molant: I'm working on something similar (making sure the typescript errors point to the wrong values and not the property names)  and with your change and my change we are pointing to the opening `{` instead of the property name when it's missing:

image

I'll look into this but I'll probably open a PR that supersedes this one.

molant: Got it working. This is my code:

```js
         while (!node && segments.length > 0) {
            node = findNodeAtLocation(this._root, segments) || null;

            if (!node) {
                missing = true;
            }

            segments.pop();
        }

        /**
         * If the property is missing but its ancestor is present we want to report
         * the property name and not the value (which will be an opening `{` ).
         */
        if (node && missing && options && options.at === 'value') {
            return this.offsetToLocation(this.getAdjustedOffset(node, path));
        }

        /**
         * The node isn't in the current file. Use alternative path if provided. This happens
         * when extending configurations.
         */
        if (!node && this._alternatePath && path !== this._alternatePath) {
            return this.getLocation(this._alternatePath, options);
        }
```

If we are OK with your PR I'm thinking about cherry-picking to preserve the history and the author. @antross @sarvaje what do you think?

sarvaje: Do I need to touch something?

molant: Don't think so. Depends on how we want to merge all of this. My code depends on yours so we could have just one PR with all the changes or a couple. Either way, can we wait until I release the new version of `create-hint` before merging anything else?

antross: New version of `create-hint` has been released. Discussed with @molant and we're okay merging this PR as-is then he can rebase his work on top of it.

PR #3128, opened 2019-10-17
Fix: Unhandled promise rejections when self-signed in jsdom

sarvaje:
## Pull request checklist

Make sure you:

### DIFF
--- a/packages/connector-jsdom/src/evaluate-runner.ts
+++ b/packages/connector-jsdom/src/evaluate-runner.ts
@@ -4,13 +4,25 @@ import { JSDOM, VirtualConsole } from 'jsdom';
 import * as jsdomutils from 'jsdom/lib/jsdom/living/generated/utils';
 
 import { debug as d } from '@hint/utils';
+import { Requester } from '@hint/utils-connector-tools';
 
 import { beforeParse } from './before-parse';
 
 const debug: debug.IDebugger = d(__filename);
 
-const run = async (data: { source: string }) => {
-    const { source } = data;
+const run = async (data: { options: any; source: string }) => {
+    const { options = {}, source } = data;
+    const requesterOptions = Object.assign(
+        {},
+        {
+            rejectUnauthorized: !options.ignoreHTTPSErrors,
+            strictSSL: !options.ignoreHTTPSErrors
+        },
+        options.requestOptions || {}
+    );
+
+    const requester = new Requester(requesterOptions);
+
     const result = {
         error: null as Error | null,
         evaluate: 'result'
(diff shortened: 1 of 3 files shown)


packages/connector-jsdom/src/evaluate-runner.ts
```diff
@@ -4,13 +4,25 @@ import { JSDOM, VirtualConsole } from 'jsdom';
...
+const run = async (data: { options: any; source: string }) => {
+    const { options = {}, source } = data;
+    const requesterOptions = Object.assign(
+        {},
+        {
+            rejectUnauthorized: !options.ignoreHTTPSErrors,
+            strictSSL: !options.ignoreHTTPSErrors
+        },
+        options.requestOptions || {}
+    );
```
antross: How about using object-spread here?
```ts
const requesterOptions = {
    rejectUnauthorized: !options.ignoreHTTPSErrors,
    strictSSL: !options.ignoreHTTPSErrors,
    ...options.requestOptions
};
```

packages/connector-jsdom/src/evaluate-runner.ts
```diff
@@ -28,7 +40,19 @@ const run = async (data: { source: string }) => {
         debug(err);
     });
 
-    const jsdom = await JSDOM.fromURL(url, {
+    let html = '';
+
+    try {
+        const networkData = await requester.get(url);
```
antross: Probably worth a comment here or before the `run` function explaining we need to request the HTML separately instead of just using `JSDOM.fromURL` in order to respect user options to ignore HTTPS errors.

packages/connector-jsdom/src/evaluate-runner.ts
```diff
@@ -28,7 +40,19 @@ const run = async (data: { source: string }) => {
...
+    } catch (error) {
+        process.send!({ error });
+
+        return;
+    }
+
+    const jsdom = new JSDOM(html, {
         beforeParse: beforeParse(url),
         pretendToBeVisual: true,
         resources: 'usable',
```
molant: Shouldn't we use the resource loader you've created to make sure all the assets are downloaded? Otherwise CSS files and other scripts might not be downloaded. Not sure that will impact a lot the output of axe but just for being extra cautious.

sarvaje: sure, we can do that.

sarvaje: I add a custom resource loader, but that is can't be used for JSDOM.fromURL, because JSDOM.fromURL is expecting that the promise has a `response` property.

PR #3138, opened 2019-10-18
Fix: Allow ranged location in css.

### LINKED ISSUE #3130: Dashed CSS property names are only partially underlined in the VS Code extension
E.g. if a report is generated for `background-color`, only `background` ends up underlined.

Fixing this probably requires a few steps:
* Introduce support for ranges on `Problem` (e.g. adding an optional `endColumn` and `endLine`)
* Update affected hints to provide ranges with reports
* Update the VS Code extension to map them to VS Code ranges

sarvaje:
Fix #3130

## Pull request checklist

Make sure you:

- [x] Introduce support for ranges on Problem (e.g. adding an optional endColumn and endLine)
- [x] Update affected hints to provide ranges with reports
- [x] Update the VS Code extension to map them to VS Code ranges

### DIFF
--- a/packages/hint-compat-api/tests/css.ts
+++ b/packages/hint-compat-api/tests/css.ts
@@ -86,7 +86,7 @@ testHint(hintPath,
             reports: [
                 {
                     message: `'display: grid' is not supported by Edge < 16. Add 'display: -ms-grid' to support Edge 12+.`,
-                    position: { match: 'display: grid; /* Report */' }
+                    position: { match: 'grid; /* Report */' }
                 }
             ],
             serverConfig: generateCSSConfig('supports')
@@ -96,19 +96,19 @@ testHint(hintPath,
             reports: [
                 {
                     message: `'display: grid' is not supported by Internet Explorer.`,
-                    position: { match: 'display: grid; /* Report 1 */' }
+                    position: { match: 'grid; /* Report 1 */' }
                 },
                 {
                     message: `'display: grid' is not supported by Internet Explorer.`,
-                    position: { match: 'display: grid; /* Report 2 */' }
+                    position: { match: 'grid; /* Report 2 */' }
                 },
                 {
                     message: `'display: -ms-grid' is not supported by Chrome, Firefox, Internet Explorer < 10. Add 'display: grid' to support Chrome 57+, Firefox 52+.`,
-                    position: { match: 'display: -ms-grid; /* Report 3 */' }
+                    position: { match: '-ms-grid; /* Report 3 */' }
                 },
                 {
                     message: `'display: grid' is not supported by Edge < 16, Internet Explorer. Add 'display: -ms-grid' to support Edge 12+, Internet Explorer 10+.`,
-                    position: { match: 'display: grid; /* Report 4 */' }
+                    position: { match: 'grid; /* Report 4 */' }
                 }
             ],
             serverConfig: generateCSSConfig('values')
--- a/packages/hint-compat-api/tests/css.ts
+++ b/packages/hint-compat-api/tests/css.ts
@@ -42,27 +42,27 @@ testHint(hintPath,
             reports: [
                 {
                     message: `'appearance' is not supported by Internet Explorer.`,
-                    position: { match: 'appearance: button; /* Report 1 */' }
+                    position: { match: 'appearance: button; /* Report 1 */', range: 'appearance' }
                 },
                 {
                     message: `'appearance' is not supported by Internet Explorer.`,
-                    position: { match: 'appearance: button; /* Report 2 */' }
+                    position: { match: 'appearance: button; /* Report 2 */', range: 'appearance' }
                 },
                 {
                     message: `'-webkit-appearance' is not supported by Internet Explorer.`,
-                    position: { match: '-webkit-appearance: button; /* Report 3 */' }
+                    position: { match: '-webkit-appearance: button; /* Report 3 */', range: '-webkit-appearance' }
                 },
                 {
                     message: `'-moz-appearance' is not supported by Internet Explorer.`,
-                    position: { match: '-moz-appearance: button; /* Report 4 */' }
+                    position: { match: '-moz-appearance: button; /* Report 4 */', range: '-moz-appearance' }
                 },
                 {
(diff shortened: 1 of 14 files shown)


sarvaje:
@molant @antross any idea why this PR is failing? I can't reproduce the error in my machine :(

sarvaje:
Ok, I'm able to reproduce the error.

I think we already discuss this before, but `yarn build` should remove dist before build. Today is the second or third time that I get confuse with the tests because of this change we did long time ago.

antross:
> I think we already discussed this before, but `yarn build` should remove dist before build.

That's what `yarn clean` is for. If we remove `dist` on every build we eliminate files which help rebuilds run faster, notably `tsconfig.tsbuildinfo`.

sarvaje:
Yes, but you need to remember run `yarn clean`...

antross:
@sarvaje Can you add a screenshot of a CSS file in the VS Code extension with these changes? I'm curious to see exactly what gets underlined as part of the new `ChildNode` mapping.

sarvaje:
image

molant:
@sarvaje can you add one with a value underlined?

sarvaje:
do you have an example?

sarvaje:
when a property is not supported I get underlined the property :s

sarvaje:
here is a screenshot.

image

antross:
@sarvaje to ensure you're using your local copy of `hint` and not a shared install you'll need to test this by opening a fixture file that's inside the repo folder. Otherwise the VS Code extension is going to be loading `hint` from `npm` which won't have your changes.

sarvaje:
@molant here is an updated screenshot with all the packages compiled with this version of the code hehehe

image

As you can see, the range includes the property and the value.
I have been talking with @antross and we think we have enough information to take the ranged position of each one (property and value) with some extra work. The only question is if we want to do that or it is ok underlining everything.

molant:
How much extra work would that be? The error is in the value so I think that we should underline only the value but I think I could convince otherwise.

antross:
> The error is in the value

@molant It depends on which error you’re talking about too. The display error would be on the value, but the appearance error would be on the name.

molant:
I meant for the screenshot that @sarvaje shared.

sarvaje:
@molant @antross what do we want to do with this? point only to the property or value, depending on the test, or keep it as is right no?

antross:
If it's not too much work, I think we should point to the property name or value depending on the specific case. 

Go ahead and try to make this work for `hint-css-prefix-order` first. Once that's working I'm happy to pair up to work out plumbing the required data through `hint-compat-api/css`.

sarvaje:
ok!

antross:
And for `display: grid` vs `display: -ms-grid`?

sarvaje:
for now it is only working for properties hehehe

sarvaje:
image

antross:
Looks good - probably worth throwing in a few test cases with weird amounts of white space and content between the name and the value to ensure we've got this right. Some ideas:
```css
.example1 {
    display:     grid;
    display: -ms-grid;
}
.example2 {
    display : grid;
    display : -ms-grid; 
}
.example3 {
    display:
        grid;
    display:
        -ms-grid;
}
.example4 {
    display /* comment */ : /* comment */ grid;
    display: -ms-grid;
}
```

sarvaje:
Those are already covered, now I'm working in a more compliated case:
```
.example4 {
    display /* comment */ :
    /* comment */ grid
    ;
    display: -ms-grid;
}
```

sarvaje:
What about this?

image

molant:
> What about this?

🧐

sarvaje:
Right now is working just for `hint-css-prefix-order` where only the pair property/value can "fail". Should I continue with the other hints?

antross:
Great! Hope it wasn't too crazy to pull off. 😉

molant:
> Should I continue with the other hints?

Yes, please!

packages/hint-compat-api/tests/css.ts
```diff
@@ -86,7 +86,7 @@ testHint(hintPath,
             reports: [
                 {
                     message: `'display: grid' is not supported by Edge < 16. Add 'display: -ms-grid' to support Edge 12+.`,
-                    position: { match: 'display: grid; /* Report */' }
+                    position: { match: 'grid; /* Report */' }
```
antross: I think we should be able to update the `match` logic to allow validating the new ranges as well. 

We can keep the current `string` value for matching the context, then add an optional `range` item to specify the sub-section we want to include. When `range` is present we'll match `endColumn` and `endLine` in addition to `column` and `line`. The necessary updates would be in `findPosition` of `@hint/utils-tests-helpers/src/hint-runner.ts`. 

For example here we could use:
```ts
{ match: 'display: grid; /* Report */', range: 'grid' }
```

What do you think?

sarvaje: Sure, I like the idea

packages/utils/src/report/get-css-location-from-node.ts
```diff
@@ -0,0 +1,64 @@
...
+    start.line = end.line;
+    start.column = end.column - node.value.length;
+};
+
+/**
+ * Returns the location of a CSS node.
+ * @param node postcss Node.
+ * @param isValue Indicates if we need the location of the property value.
+ */
+export const getLocationFromNode = (node: ChildNode, isValue?: boolean): ProblemLocation | undefined => {
```
molant: I think the name should be the same as the file to prevent any confusion and keep things as they are already:
```suggestion
export const getCSSLocationFromNode = (node: ChildNode, isValue?: boolean): ProblemLocation | undefined => {
```

molant: Also I think we should be consistent with what we have right now in [`json-parser`](https://github.com/webhintio/hint/blob/94dfef55140f6aa9b3dc3d071f9558a4afecb47d/packages/utils/src/json-parser.ts#L28) and accept an `options` object ([`IJSONLocationOptions`](https://github.com/webhintio/hint/blob/94dfef55140f6aa9b3dc3d071f9558a4afecb47d/packages/utils/src/types/json-parser.ts#L3)) in case we need to add more things in the future.

sarvaje: That type shouldn't be an `interface` :P. I will do it

PR #3145, opened 2019-10-18
New: Add isFragment and isAttributeAnExpression

antross:
New helpers for working with HTML template content.

Co-authored-by: Antón Molleda <amolleda@gmail.com>

## Pull request checklist

Make sure you:

- [ ] Signed the [Contributor License Agreement](https://cla.js.foundation/webhintio/hint)


molant:
I talked with @antross and I'm going to work on moving some of the work from #3109 here.

molant:
This should be ready for review.

Also 😍:

 image

packages/hint-axe/src/util/axe.ts
```diff
@@ -24,6 +24,26 @@ type Registration = {
 
 type RegistrationMap = Map<EngineKey, Map<string, Registration[]>>;
 
+let html: HTMLElement;
```
antross: @molant we'll need to reset `html` on `scan::end` or this won't work correctly on subsequent runs. 

That said, what do you think about enhancing the `can-evaluate::script` instead to include the relevant `HTMLDocument` instance to make this simpler?

molant: Need to find out if that's possible in jsdom or if we will have to emit the event at another time but definitely better than what is there right now.

molant: Your registration thing in axe is giving me a headache @antross 😅

molant: ready for another review

packages/utils/src/dom/html.ts
```diff
@@ -211,17 +229,21 @@ export class HTMLDocument {
...
     public get base(): string {
         return this._base;
     }
 
+    /**
+     * Check if this represents a template fragment as opposed to a full document.
+     */
+    public get isFragment(): boolean {
+        // Document is a fragment if `<html>` wasn't part of the original source.
+        return !this.originalDocument && !this._documentElement.sourceCodeLocation;
```
sarvaje: Is it possible to not have `originalDocument` and not to have `_documentElement`?

molant: Yes, in fact most of the times we create the `HTMLDocument` without `originalDocument`. `originalDocument` is used in `jsdom` and `puppeteer` because we parsed twice:

1. original HTML down the wire
2. another time when we do the traversing to make sure we visit all possible new elements

It's the second time we send the `originalDocument` so we can find the position of the ones that were already present initially.

I also see why the name `originalDocument` could confuse anyone.

sarvaje: Yep, maybe the name was confusing me.

Issue #2741, opened 2019-07-26
[Hint] Base font-size


ststimac:
# 🚀 Hint request
Check the base font size of a website and whether it's using relative units.

`The best practice is to set the font-size on the html using % (or any other relative unit), and then set all the other elements to either em or rem which will be relative to the body size.`

## Description

 What do you want webhint to test?

The hint should check to see if font-size is set on `html` in the CSS file and if it is, what unit of measurement is being used.

 Are you willing to work on this?
I can write the documentation. Will need help writing the hint.

## What should be tested exactly?

 Explain what and how things should be tested. If you don't know,
don't worry and leave this empty.

The Hint should pass if a relative unit is being used to set the font-size: ex, ch, em, rem and %.
The Hint should fail if anything else is being used, like pixels or viewport units.*

*Important note about viewport units:

> viewport units aren’t relative units. I know they’re listed under relative units and described as such, but they don’t always act like relative units (resizing font sizes set with viewport units worked for me in Safari but not in Chrome and Firefox).

## Relevant links

 Any links to documentation, discussions, etc. relevant to the
proposed hint that motivated you to open this issue.

> most browsers agreed on the default size of 16 pixels, but the user can change that. Let’s say someone has an impaired vision and sets their browser’s default font size to something larger. Setting the base font size in our CSS to pixels will overwrite their browser default font size setting. Instead of seeing most of the text at a larger font size, we now force them to read it at 16 pixels size. That’s because pixels are an absolute unit, not a relative one.

This Hint is important because it ensures that if users need to zoom in on a webpapge, the text on the page will scale accordingly and maintain readability. 

All quotes in this are from https://betterwebtype.com/articles/2019/06/16/5-keys-to-accessible-web-typography/

PR #3167, opened 2019-10-22
New: Add accessibility-related hint for relative font units

smhigley:
Ref #2741

## Pull request checklist

Make sure you:

Adds an accessibility-focused check related to the [WCAG Text Spacing criterion](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html). Addresses #2741, but rather than only check the `html` tag, it more robustly tests all `font-size`, `line-height`, and `letter-spacing` styles.


packages/hint-font-relative-units/README.md
```diff
@@ -0,0 +1,66 @@
+# font-relative-units (`@hint/hint-font-relative-units`)
+
+Ensure font styles use relative units instead of pixels.
+
+## Why is this important?
+
+If website stylesheets define certain styles in pixels, they cannot be easily overridden by custom stylesheets and browser plugins. Some people benefit from altering font size, line height, and letter spacing to make text easier to read.
```
molant: Our linters are going to complain if the length of a line in markdown is more than 80 characters.

If you run `yarn test` from your new hint to run all linters.

packages/hint-font-relative-units/src/_locales/en/messages.json
```diff
@@ -0,0 +1,14 @@
+{
+    "description": {
+        "description": "Metadata description",
+        "message": "Ensure font styles use relative units to enable user stylesheet overrides"
+    },
+    "name": {
+        "description": "Metadata name",
+        "message": "font-relative-units"
```
molant: I believe this is the friendly name of the hint, no need for `-`

packages/hint-font-relative-units/src/hint.ts
```diff
@@ -0,0 +1,91 @@
...
+                if (!validRule) {
+                    invalidDeclarations.push(decl);
+                }
+            });
+
+            return invalidDeclarations;
+        };
+
+        context.on('parse::end::css', ({ ast, element, resource }: StyleParse) => {
+            debug('Validating hint css-prefix-order');
```
molant:
```suggestion
            debug('Validating hint font-relative-units');
```

packages/hint-font-relative-units/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## What does the hint check?
+
+This hint examines `css` and `scss` files and checks that the following properties are not defined in pixels:
```
molant: I believe `less` files will be checked as well. @antross can you confirm?

packages/hint-font-relative-units/README.md
```diff
@@ -0,0 +1,66 @@
...
+{
+    "connector": {...},
+    "formatters": [...],
+    "parsers": [...],
+    "hints": {
+        "font-relative-units": "error"
+    },
+    ...
+}
+```
```
molant: **Note to @webhintio/core:**  If we decide to enable this hint in any of the built-in configurations we should update this section and use some of the wording we have in others. Also we should check this text is not in the templates of `create-hint`.

packages/hint-font-relative-units/package.json
```diff
@@ -0,0 +1,68 @@
...
+    ],
+    "timeout": "1m"
+  },
+  "description": "Ensure font styles use relative units to enable user stylesheet overrides",
+  "devDependencies": {
+    "@hint/utils-tests-helpers": "^5.0.11",
+    "@types/node": "^12.7.5",
+    "@typescript-eslint/eslint-plugin": "^1.13.0",
+    "@typescript-eslint/parser": "^1.12.0",
+    "ava": "^1.4.1",
```
molant: **Note to @webhintio/core:** Make sure all dependencies are up to date before merging. We've had a few issues in the past with this.

packages/hint-font-relative-units/src/meta.ts
```diff
@@ -0,0 +1,32 @@
...
+    getName(language: string) {
+        return getMessage('name', language);
+    },
+    id: 'font-relative-units',
+    schema: [
+        /*
+         * If you want to allow the user to configure your hint
+         * you should use a valid JSON schema. More info in:
+         * https://webhint.io/docs/contributor-guide/hints/#themetaproperty
+         */
```
molant: Remove the comment inside `schema`:

`schema: [],`

packages/hint-font-relative-units/tests/tests.ts
```diff
@@ -0,0 +1,132 @@
...
+            },
+            {
+                message: 'letter-spacing should be defined with relative units instead of pixels',
+                position: { match: 'letter-spacing: 0.5px;' }
+            },
+            {
+                message: 'font-size should be defined with relative units instead of pixels',
+                position: { match: 'font-size: 17px;' }
+            }
+            ,
```
molant: Linter is complaining about this `,` here and not in the line above.

packages/hint-font-relative-units/src/_locales/en/messages.json
```diff
@@ -0,0 +1,14 @@
...
+    },
+    "pixelError": {
+        "description": "Metadata pixelError",
+        "message": "$1 should be defined with relative units instead of pixels"
```
molant: For consistency with other messages can you please update this (and the tests then) to:

```suggestion
        "message": "$1 should be defined with relative units instead of pixels."
```

packages/hint-font-relative-units/src/hint.ts
```diff
@@ -0,0 +1,91 @@
+/**
+ * @fileoverview Ensure font styles use relative units to enable user stylesheet overrides
+ */
+
+import { Declaration, Rule } from 'postcss';
+import { HintContext } from 'hint/dist/src/lib/hint-context';
+// The list of types depends on the events you want to capture.
```
molant: You can remove this line
```suggestion
```

packages/hint-font-relative-units/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Further Reading
+
+<!-- Link labels: -->
+
+[WCAG Text Spacing requirement]: https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html
```
molant: I don't see this link used in the markdown and otherwise it will not be visible. I believe it should be under `## Further Reading`: 

```md
## Further Reading

* [WCAG Text Spacing requirement][]
```

packages/hint-font-relative-units/tests/tests.ts
```diff
@@ -0,0 +1,132 @@
...
+        }
+    };
+};
+
+/*
+ * You should test for cases where the hint passes and doesn't.
+ * More information about how `hintRunner` can be configured is
+ * available in:
+ * https://webhint.io/docs/contributor-guide/how-to/test-hints/
+ */
```
molant: You can remove this comment too.

packages/hint-font-relative-units/README.md
ststimac: I think we should reword the "why is this important" section. 
Pixels are not a relative unit and do not scale, but that doesn't necessarily mean pixels are hard to override. Because pixels aren't in a scalable unit, when users who need to adjust a web page for readability, they don't actually scale to increase or decrease with page zoom, disadvantaging users who need to alter those settings.

packages/hint-font-relative-units/src/hint.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+export default class FontA11yFontSizeUnitsHint implements IHint {
+
+    public static readonly meta = meta;
+    public readonly FONT_PROPERTIES = ['font-size', 'line-height', 'letter-spacing'];
+
+    public constructor(context: HintContext<StyleEvents>) {
+
+        const getLocation = (decl: Declaration): ProblemLocation => {
```
sarvaje: This should be removed in favor of `getLocationFromNode` once https://github.com/webhintio/hint/pull/3138 is merged.

packages/hint-font-relative-units/src/hint.ts
```diff
@@ -0,0 +1,91 @@
...
+            debug('Validating hint css-prefix-order');
+
+            const getPixelErrorMessage = function(decl: Declaration): string {
+                return getMessage('pixelError', context.language, [decl.prop]);
+            };
+
+            ast.walkRules((rule) => {
+                for (const invalidDeclaration of validateRule(rule)) {
+                    const message = getPixelErrorMessage(invalidDeclaration);
+                    const location = getLocation(invalidDeclaration);
```
sarvaje: `const location = getLocationFromNode(invalidDeclaration);` once https://github.com/webhintio/hint/pull/3138 is merged.

packages/hint-font-relative-units/README.md
antross: Yes, as will `sass` files (in addition to `scss`) and potentially more in the future as we add new parsers (e.g. CSS-in-JS). Plus inline style blocks in `html` files are also analyzed. 

Perhaps sticking with just "`css` files" is enough to get the point across. Or we could try something more generic like "files containing style information".

packages/hint-font-relative-units/README.md
smhigley: @ststimac it’s less for zoom (most browsers will actually zoom pixel values as well as relative units) and more for changing the default browser font size or user stylesheets (or plugins that work by adding custom css) that set top-level font size, line height, etc.

How would you feel about a change to the wording along the lines of “If website stylesheets define certain styles in pixels, they do not reflect changes to root-level font size”?

ststimac: That sounds good!

packages/hint-font-relative-units/src/hint.ts
```diff
@@ -0,0 +1,90 @@
...
+
+            return {
+                column: 0,
+                line: 0
+            };
+        };
+
+        const validateFontValue = (prop: string, value: string): boolean => {
+            if (this.FONT_PROPERTIES.includes(prop)) {
+                return value.indexOf('px') === -1;
```
sarvaje: I think you need to check also the viewport units per this comment in the hint description:

>viewport units aren’t relative units. I know they’re listed under relative units and described as such, but they don’t always act like relative units (resizing font sizes set with viewport units worked for me in Safari but not in Chrome and Firefox).

Issue #2076, opened 2019-03-17
[Bug] Consistency with error locations


molant:
While we've been a lot of progress improving the location of the errors, I don't think we are fully consistent. For example, when reporting errors in `html` we point to the first letter of the element:

![image](https://user-images.githubusercontent.com/606594/54485327-e9874880-4833-11e9-88fa-37b14cfc2a5a.png)

But when we do it on JSON we point to `"`: 

![image](https://user-images.githubusercontent.com/606594/54485330-fb68eb80-4833-11e9-9aad-95c56bfa3b3e.png)

Visually I think I prefer pointing to the letter rather than `<` or `"` but I care more about being consistent.

antross:
Also worth noting that we point to the first letter of tag names because VS Code will only underline the `<` if you point to that (unless you also provide an end position, which we don’t currently but could consider as well).

antross:
Plus I think we do point to the first letter in the JSON case as well judging from my current PR to match text for test cases. VS Code may be choosing to extend the underline to include the quotes in this case. Once again something we can probably override if we provide an end position.

molant:
I've updated the initial comment with the different cases I've identified so far and what I think should be the expected output (haven't checked if it is possible in all scenarios).

Let me know what you think!

molant:
Looks like most of the json issues are going to be easy to fix, we were missing `{ at: 'value' }` in a bunch of places.

While we are changing the position of the output I'd consider that a `fix` rather and a `new`.

molant:
I'm changing my PoV and I think we should point to `"` at least on JSON:

![image](https://user-images.githubusercontent.com/606594/66973411-cbddc400-f04c-11e9-9336-9ace8285cc68.png)

Otherwise we will have an special case and I'm not sure it's worth it.

molant:
@sarvaje, @antross I have a few questions about where we want to report issues with `extends`.

In the case of TypeScript and after @sarvaje's changes, the error should be pointing to `compilerOptions` in most cases. E.g.: If an `extends` has `removeComments = false` the error will point to the `compilerOptions` of the file using the `extends`:

![image](https://user-images.githubusercontent.com/606594/67064079-fa779f80-f11d-11e9-997b-774547d7fbf0.png)

I'm not sure this is entirely correct as the file having the issue is the extended one and not the current. Anyways, now I'm changing to point to the values so we are underlining the `extends`:

[...]

PR #3168, opened 2019-10-22
Fix/error location

molant:
## Pull request checklist

Make sure you:

Point to the values where applicable instead of the property names on `json` based tests (manifest, typescript, babel).

Also fix an issue not returning the right file when "extending".

This is part of #2076 but a bit more of work is needed for other file formats so not going to link to it (we might need to think about changing it into an epic tbh).


molant:
@antross @sarvaje this should be ready to review. I've separated everything in 5 commits so hopefully that should make things easier to review.

packages/hint-babel-config/tests/is-valid.ts
```diff
@@ -62,7 +62,7 @@ const tests: HintLocalTest[] = [
         reports: [
             {
                 message: `Circular reference found in file ${path.join(__dirname, 'fixtures', 'circular-2', '.babelrc')}`,
-                position: { match: 'extends' }
+                position: { match: '"../circular-2/.babelrc"' }
```
antross: Note: Once @sarvaje adds range support to tests per https://github.com/webhintio/hint/pull/3138/files#r338139101 we should be able to use that here. Though we may want to make `match` optional for cases like this where the `range` and `match` are going to be identical.

molant:
Missing property on top level:
image

Missing property in an item in an array:
image

Incorrect property value:
image

Missing property in an array:
image

Invalid array property:
image

Invalid property:
image

Missing properties:
image

PR #3170, opened 2019-10-22
New: Support Vue single file components #2890

### LINKED ISSUE #2890: Support Vue single file components

arturmiz:
## Pull request checklist

Make sure you:

Closes #2890

### DIFF
--- a/packages/parser-vue/README.md
+++ b/packages/parser-vue/README.md
@@ -1,0 +1,40 @@
+# Vue (`@hint/parser-vue`)
+
+The `Vue` parser is built so hints can analyze `Vue` files.
+
+This package is installed automatically by webhint:
+
+```bash
+npm install hint --save-dev
+```
+
+To use it, activate it via the [`.hintrc`][hintrc] configuration file:
+
+```json
+{
+    "connector": {...},
+    "formatters": [...],
+    "hints": {
+        ...
+    },
+    "parsers": ["vue"],
+    ...
+}
+```
+
+**Note**: The recommended way of running webhint is as a `devDependency` of
+your project.
+
+## Events emitted
+
+This `parser` emits the following events:
+
+* `parse::start::vue` of type `Event` which contains the following information:
+
+  * `resource`: the resource we are going to parse.
+
+* `parse::end::vue` of type `Event` which has the following information:
+
+  * `resource`: the parsed resource.
+
+[hintrc]: https://webhint.io/docs/user-guide/configuring-webhint/summary/
--- a/packages/configuration-development/index.json
+++ b/packages/configuration-development/index.json
@@ -46,6 +46,7 @@
         "less",
         "sass",
         "typescript-config",
+        "vue",
         "webpack-config"
     ]
 }
--- a/packages/configuration-development/package.json
+++ b/packages/configuration-development/package.json
@@ -26,6 +26,7 @@
     "@hint/parser-less": "^1.0.5",
     "@hint/parser-sass": "^1.0.5",
     "@hint/parser-typescript-config": "^2.3.10",
+    "@hint/parser-vue": "^1.0.0",
     "@hint/parser-webpack-config": "^2.1.13"
   },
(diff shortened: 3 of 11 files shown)


antross:
@arturmiz Awesome to see this progressing! 

As always, feel free to ask if you have any questions - we're happy to help.

arturmiz:
@antross I'll ask for tips for sure ;) I had some issues to build the extension locally, that's why I decided to try to build it via a pipeline to verify how it goes. I'll do my best to implement this one :)

antross:
> I had some issues to build the extension locally

Sorry about that. I'd love to help you get that working locally too if you can share more about what you ran into. Was this the `extension-vscode` package that was giving you trouble or `extension-browser` (or something else entirely)?

packages/parser-vue/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+**Note**: The recommended way of running webhint is as a `devDependency` of
+your project.
+
+## Events emitted
+
+This `parser` emits the following events:
+
+* `parse::start::vue` of type `Event` which contains the following information:
```
antross: I think we'll want to have this emit `parse::start::html` and `parse::end::html` events similar to what we did for `parser-jsx`. This will allow existing hints written for HTML content to pick this up automatically. From what I've seen of single-file components, this shouldn't require much (if any) translation to be suitable-enough for hint consumption.

On a related note, I should mention we now have built-in support for recognizing when documents are fragments to avoid applying guidance for things like `meta` tags, etc. I suspect this should "just work" if you use a similar approach to `parser-html` to parse the content of the component file.

arturmiz: Hi again! 😉 
I got back to this and I think I'm missing some place to reference the `parser-vue` package as when I build it and try to run the `extension-code`, I open `.vue` file inside new VSC window but nothing gets logged. I checked the `dist` folder inside `extension-code` package, but the only thing matching `vue` is `onLanguage:vue` I put in `package.json`. 
How to debug specific parser code then? 🤔

antross: I think this may be because `.vue` is not recognized by VS Code out-of-the-box so the file never gets passed to our extension.

Here's the list of built-in supported languages (and instructions for adding new ones):
https://code.visualstudio.com/docs/languages/identifiers

Because of this, registering for `onLanguage:vue` won't do anything unless we define the language as well, or until another extension that declares the `vue` language is installed. 

We probably don't want to be in the business of declaring the language configuration, but at least from the samples it doesn't look like that's a required field. So in this case we'd probably need to either need to declare the language using something like (a):

```json
    "languages": [{
        "id": "vue",
        "extensions": [ ".vue" ],
        "aliases": [ "Vue", "vue" ]
    }]
```
Or (b) require vetur to also be installed before we start analyzing `.vue` files: https://marketplace.visualstudio.com/items?itemName=octref.vetur&WT.mc_id=marketplace-pack-sdras

The former seems to be a potentially better user experience, but I don't know how multiple definitions are resolved by VS Code. If we go that route, we'd need to ensure it still works properly if a user already has the vetur extension installed.

arturmiz: I did add it to `package.json` under `contributes` but still doesn't work. But the weird part is that when I run Debugger, the extension is not loaded on the extensions list... I guess it should, right? Very weird...

antross: That is weird. Apart from the debugger you can check for signs of life from the extension by going to "View > Output" which opens a panel in the same area as the Terminal. In the "Output" panel there's a drop-down to the right which you should be able to switch to "webhint" to see debugging logs. 

I'd say try this first with something like a `.html` file we know should get activated. If that doesn't work then something's not getting attached/started...

StEvUgnIn:
The author has defintely some issue with merge.

Can you help him to merge his last commit into master:
https://github.com/webhintio/hint/pull/3170/commits/2f6ef06d94b2feb1acb086a7b41ef471ffe75513

PR #3184, opened 2019-10-24
Testing/same thread server

### LINKED ISSUE #1780: [Feature] Run web test server in same thread when in CI
# 🚀 Feature request

## Description

When running in CI we should run the web test servers in the same thread. 

From a quick test there are about ~5min difference in average per platform when running the tests:
* https://dev.azure.com/webhint/webhint/_build/results?buildId=1078
* https://dev.azure.com/webhint/webhint/_build/results?buildId=1072

Keep in mind that in the fastest one there are 3 packages that are not build nor tested (48 vs 45) so the real time difference is probably smaller but still significant.

molant:
## Pull request checklist

Make sure you:

Fix #1780

Quick run from my machine looks promising. Results are from running `hint-sri` which has a few tests.

With new thread:
image

Same thread:
image

### DIFF
--- a/packages/utils-create-server/src/index.ts
+++ b/packages/utils-create-server/src/index.ts
@@ -1,14 +1,12 @@
-import { spawn, ChildProcess, SpawnOptions } from 'child_process';
-
 import map = require('lodash/map');
 import reduce = require('lodash/reduce');
 
 import * as isCI from 'is-ci';
 
-import { IServer, Message, ServerConfiguration } from './types';
-import { replacer, reviver } from './buffer-serialization';
+import { IServer, ServerConfiguration } from './types';
 
-import { Server as ServerProcess } from './server';
+import { SameThreadServer } from './same-thread-server';
+import { IndependentServer } from './independent-thread-server';
 
 export { ServerConfiguration };
 export { IServer };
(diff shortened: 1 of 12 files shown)


molant:
This is ready for review. I had to make the tests in `strict-transport-security` serial because I couldn't find out how to run them in parallel, even when using sinon, mocks, etc. I think we can open another PR/issue for that because the hint has some "special" code to be mockable.

Also results seems promising. Everything finished under 60 minutes (linux in under 45min) and this touches almost every single package (I believe the missing one is `utils`?).

To compare, [the commit that enables TypeScript by default](https://dev.azure.com/webhint/webhint/_build/results?buildId=7043) takes about 10min more.
Need to double check if we are running the same number of tests (we are probably running more in this PR).

packages/utils-create-server/src/index.ts
```diff
@@ -3,19 +3,25 @@ import { spawn, ChildProcess, SpawnOptions } from 'child_process';
...
-import { Message, ServerConfiguration } from './types';
+import * as isCI from 'is-ci';
+
+import { IServer, Message, ServerConfiguration } from './types';
 import { replacer, reviver } from './buffer-serialization';
 
+import { Server as ServerProcess } from './server';
+
 export { ServerConfiguration };
+export { IServer };
```
sarvaje: why do you export `IServer` here?

packages/utils-create-server/src/index.ts
```diff
@@ -3,19 +3,25 @@ import { spawn, ChildProcess, SpawnOptions } from 'child_process';
...
 // Do all magic wit process here and maybe have a wrapper
 
 type ServerOptions = {
     configuration?: ServerConfiguration;
     isHTTPS?: boolean;
+    sameThread?: boolean;
 }
 
-export class Server {
+class IndependentServer implements IServer {
```
sarvaje: What's the difference between `IndependentServer` and `Server`? Can you add some comments?

packages/utils-create-server/src/index.ts
```diff
@@ -112,20 +118,26 @@ export class Server {
...
     public get port() {
         return this._port;
     }
 
     public configure(configuration: ServerConfiguration) {
         return this.actionWrapper('configure', configuration);
     }
+}
 
+export class Server {
```
sarvaje: IMO, the naming is a little bit confusing, you are importing a Server renamed to ServerProcess, and then you are exporting here another Server that looks to be a factory for `Server` and `IndependentServer`, plus other functionality.

packages/utils-create-server/src/index.ts
sarvaje: Also, maybe it would be good to move this class in its own file.

molant: `Server` is a static class with a factory that get exported so everyone has to go through here.

packages/utils-create-server/src/index.ts
molant: `Server` and `IServer` are the only pieces seen from the outside. As you've seen `Server` is a factory of `IServer` that will decide what to use depending on the environment or the user's configuration (`Independent` if not in ci).

molant:
@sarvaje give it another look please

packages/utils-create-server/src/index.ts
sarvaje: Ahh, ok, this is because this is the `index.ts` file hehehe.

packages/utils-create-server/src/index.ts
```diff
@@ -156,16 +56,41 @@ export class Server {
...
 
         if (options && options.configuration) {
-            await server.configure(Server.updateLocalhost(options.configuration, server._port));
+            await server.configure(Server.updateLocalhost(options.configuration, server.port));
         }
 
         return server;
     }
+
+    private static async createServer(options: ServerOptions) {
```
sarvaje: Rename `createServer` to `createSameThreadServer`.

PR #3196, opened 2019-10-27
Fix: Edge error locations when using `extends`

molant:
## Pull request checklist

Make sure you:

Improvements in a couple of edge cases:

- `target` was pointing to `{` because it did not do the right search
- when a configuration `extends` but none of the files had a required
  property, `extends` got underlined because it was set as the
  `alternatePath`. With the improvements in the node search this is no
  longer needed.

### DIFF
--- a/packages/hint-typescript-config/src/helpers/config-checker.ts
+++ b/packages/hint-typescript-config/src/helpers/config-checker.ts
@@ -1,5 +1,5 @@
 import { TypeScriptConfigParse, TypeScriptConfig } from '@hint/parser-typescript-config';
-import { HintContext } from 'hint';
+import { HintContext, IJSONLocationFunction } from 'hint';
 
 import { getMessage, MessageName } from '../i18n.import';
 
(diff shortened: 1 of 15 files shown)


sarvaje:
@molant can you upload some screenshots?

molant:
Wrong `target` points to `target` value:

image


Missing `target` points to `compilerOptions`:

image

Wrong `target` in `extends`:

image

Missing `target` in `extends` and current file points to `extends`. This is because the default value is `es3` so we need to point to somewhere and right now the behavior is `extends` (and both files are technically incorrect): 

image

If we don't have `compilerOptions` is not a valid tsconfig.json so it doesn't pass.

molant:
I've also added position to the tests to make sure we don't regress.

molant:
> Do we have tests for the extends property and value?

Yes, they were already in `is-valid` and `no-comments` .
I can add a new one for `target` if you want.

sarvaje:
And those tests are testing also the position (for property and value)? If so, no need to add tests for `target` too.

molant:
Yes, they test for the position but I'm adding the `target` ones as well as they use a different logic.

sarvaje:
Ok, perfect!

molant:
Done

packages/hint-typescript-config/src/helpers/config-checker.ts
```diff
@@ -15,6 +15,38 @@ const findValue = (property: string, config: TypeScriptConfig) => {
...
+
+    const ancestors = propertyPath.split('.');
+
+    ancestors.pop();
+
+    if (typeof valueInOriginal !== 'undefined') {
+
+        location = getLocation(propertyPath, { at: 'value' });
+
+    } else if (targetValue !== valueInOriginal &&
```
antross: Given `valueInOriginal` is `undefined` by this point do we even need to compare?
```suggestion
    } else if (
```

molant: Done

molant:
@antross do you have the right timestamp or is my machine the broken one?
image

antross:
@molant must be your machine - your commit to fix linting appears even before my review comments from earlier in the day...

image

antross:
I just combined related commits. Will merge once CI passes.

antross:
@sarvaje Did you want to give this one final look? We're hoping to merge before releasing today if possible.

PR #3297, opened 2019-11-07
New: Improve error messages when running tests

molant:
## Pull request checklist

Make sure you:

When validating a report `hint-runner` checked all the properties at the
same time. This made difficult to debug what property was not matching,
and what was the expected value.

Now `hint-runner` provides more information for each of the properties
that validates, and comparing to the closest matched report.

Old behavior:

image

New behavior:

image

### DIFF
--- a/packages/hint-apple-touch-icons/tests/tests.ts
+++ b/packages/hint-apple-touch-icons/tests/tests.ts
@@ -205,10 +205,12 @@ const tests: HintTest[] = [
             },
             {
                 message: elementAlreadySpecifiedErrorMessage,
+                position: { column: 17, line: 4 },
                 severity: Severity.warning
             },
             {
                 message: elementAlreadySpecifiedErrorMessage,
+                position: { column: 17, line: 6 },
                 severity: Severity.warning
             }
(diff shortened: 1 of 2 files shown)


molant:
The current error is because in apple-touch-icons we have 2 reports that are exactly the same (they only have the message). I think we should add a location to both of them and in general try to avoid that scenario.

@antross what do you think?

antross:
> The current error is because in apple-touch-icons we have 2 reports that are exactly the same (they only have the message). I think we should add a location to both of them and in general try to avoid that scenario.
> 
> @antross what do you think?

SGTM

molant:
`no-broken-links` fails because it has duplicate reports.
We haven't done any progress on pointing to attributes so not sure how to fix this. Maybe have some code at the end in `hint-test-runner` that checks that all the filtered messages are the same and validate everything (but we will have to remove them from the list similarly to what was done before).

molant:
@sarvaje this should be ready for another review

sarvaje:
LGTM if the tests pass we can merge it!

PR #3303, opened 2019-11-07
New: Default severity for `button-type`

### LINKED ISSUE #3212: [Severity Updates] hint-button-type
| Report | Severity |
| --------| ----------|
| Message to show when a button doesn't have a type attribute.: Button type attribute has not been set | "Warning if the button is in a form. Hint if we don't know" |
| Message to show when a button type attribute is not valid.: Invalid button type: $1 | Error |

molant:
## Pull request checklist

Make sure you:

Fix #3212


packages/hint-button-type/src/hint.ts
```diff
@@ -23,6 +23,20 @@ export default class ButtonTypeHint implements IHint {
...
+        let inAForm = false;
+
+        const traverseDown = (evt: TraverseDown) => {
+            if (evt.element.nodeName === 'form') {
+                inAForm = true;
+            }
+        };
+
+        const traverseUp = (evt: TraverseUp) => {
+            if (evt.element.nodeName === 'form') {
```
antross: This seems a bit indirect. 

How about walking the `parentElement` chain on the `HTMLElement` returned by the `ElementFound` event instead?

molant: Not a big fun of that tbh.
Is it just because this is the first time we use this event and you wanted to get rid of them? 😋

antross: > Not a big fun of that tbh.

Why?

> Is it just because this is the first time we use this event and you wanted to get rid of them?

No 😋 But seriously, I'm concerned about storing contextual state in a "global" context and taking a dependency on the order in which events will fire.

For example, as-is this code will break if encountering a nested form. Not saying a nested form is a good idea, but I'm sure we'll hit it on some site somewhere...

molant: Hope you like this version better 😋

PR #3307, opened 2019-11-08
New: Default severity for `hint-compat-api`

### LINKED ISSUE #3213: [Severity Updates] Hint-compat-api
| Report | Severity |
| --------| ----------|
| Message showing a feature which can be added to enhance support.: Add '$1' to support $2. | Error |
| Message to show when a feature is not supported: '$1' is not supported by $2. | Warning |

sarvaje:
Fix #3213

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint-compat-api/src/css.ts
+++ b/packages/hint-compat-api/src/css.ts
@@ -6,7 +6,7 @@ import intersection = require('lodash/intersection');
 import { vendor, AtRule, Rule, Declaration, ChildNode, ContainerBase } from 'postcss';
 
 import { HintContext } from 'hint/dist/src/lib/hint-context';
-import { IHint } from 'hint/dist/src/lib/types';
+import { IHint, Severity } from 'hint/dist/src/lib/types';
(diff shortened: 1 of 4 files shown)


packages/hint-compat-api/src/css.ts
```diff
@@ -24,6 +24,7 @@ type ReportData = {
     formatFeature?: (name: string) => string;
     isValue?: boolean;
     node: ChildNode;
+    severity?: Severity;
```
molant: I don't think I've seen any `ReportData` with a custom `severity`. If we aren't using it maybe we should just delete this?

antross: Agreed. This is going to require a different approach anyway as the current implementation always ends up with a severity of `warning`. That didn't get caught in CI due to an issue in the test runner which I just posted a fix for in #3310.

@sarvaje I think the easiest way to flip between `warning` and `error` is to check in the `report` function if `formatAlternatives` returns a non-empty array. If so, that should be an `error`, otherwise a `warning`.

PR #3332, opened 2019-11-11
Build: Ensure connectors are built before running tests

antross:
Resolves CI failures affecting multiple PRs.

## Pull request checklist

Make sure you:

- [x] Signed the [Contributor License Agreement](https://cla.js.foundation/webhintio/hint)
- [x] Followed the [commit message guidelines](https://webhint.io/docs/contributor-guide/getting-started/pull-requests/#commit-messages)

For non-trivial changes, please make sure you also:

- [ ] Added/Updated related documentation.
- [ ] Added/Updated related tests.

PR #3334, opened 2019-11-12
Build: Ensure connectors are built before running tests (v2)

molant:
Built on top of #3332 rebasing with master and not following `peerDependencies`


molant:
After talking with @sarvaje what he's going to do is move `generate-html-page` to `@hint/utils-create-server` and update the references. That should break the cycle we have between connectors <--> utils-test-helpers and pass the build.
Closing this PR as it will be obsolete hopefully soon.

PR #3339, opened 2019-11-12
New: Default severity for `http-cache`

### LINKED ISSUE #3222: [Severity Updates] hint-http-cache
| Report | Severity |
| --------| ----------|
| Report message when the directives are invalid: The $1 $2 $3 invalid | Error |
| Report message when the directives have invalid value: The following $1 an invalid value: $2 | Error |
| Report message when a directive is not recomended: The directive "$1" is not recommended | Warning | 
| Report message when a cache busting pattern is not configured: No configured patterns for cache busting match $1. See docs to add a custom one. | Warning |
| Report message when the header "cache-control" doesn't exists or is empty: No "cache-control" header or empty value found. It should have a value | Error |
| Report message when an static resource is not using the directive 'immutable': Static resources should use the "immutable" directive:Directives used: $1 | Warning if targeting a browser that isn't immutable by default, Hint otherwise. Take also into account cache busting (warning if yes, hint if not) |
| Report message when a static resource has an short cache value: Static resources should have a long cache value ($1):Directives used: $2 | Warning if file has cache busting, Hint otherwise |
| Report message when the target shouldn't be cached: The target should not be cached, or have a small "max-age" value ($1): $2 | Warning | 
| Report message when the Cache-Control header is using a wrong combination of directives: The following Cache-Control header is using a wrong combination of directives: $1 | Error |

Anton: We could probably make the hint smarter and if no cache-busting pattern is identified for a given asset, then we do not require a long cache. This should solve flagging google-analytics for example
Tony: Should we consider values that are "close" to our long/short recommendations?

molant:
## Pull request checklist

Make sure you:

Fix #3222

### DIFF
--- a/packages/hint-http-cache/tests/tests.ts
+++ b/packages/hint-http-cache/tests/tests.ts
@@ -1,9 +1,10 @@
 import { generateHTMLPage } from '@hint/utils-create-server';
-import { getHintPath, HintTest, testHint } from '@hint/utils-tests-helpers';
+import { getHintPath, testHint } from '@hint/utils-tests-helpers';
+import { Severity } from '@hint/utils-types';
 
 const hintPath = getHintPath(__filename);
 
-const defaultTests: HintTest[] = [
+const defaultTests = [
     {
         name: `Target with "Cache-Control: no-cache" passes`,
         serverConfig: {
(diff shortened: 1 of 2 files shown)


packages/hint-http-cache/tests/tests.ts
```diff
@@ -1,9 +1,9 @@
-import { generateHTMLPage } from '@hint/utils-create-server';
-import { getHintPath, HintTest, testHint } from '@hint/utils-tests-helpers';
+import { generateHTMLPage, getHintPath, testHint } from '@hint/utils-tests-helpers';
```
sarvaje: You need to undo this change. `generateHTMLPage` is in `@hint/utils-create-server` now.

molant: done

PR #3348, opened 2019-11-15
Fix: Schema and Types for hints

sarvaje:
## Pull request checklist

Make sure you:

### DIFF
--- a/packages/utils/src/types/config.ts
+++ b/packages/utils/src/types/config.ts
@@ -29,7 +29,7 @@ export type HintConfig = HintSeverity | [HintSeverity, any];
  * ```
  */
 export type HintsConfigObject = {
-    [key: string]: HintConfig | HintConfig[];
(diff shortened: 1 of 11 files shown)


packages/utils/src/types/config.ts
```diff
@@ -47,12 +47,16 @@ export type IgnoredUrl = {
     hints: string[];
 };
 
+export type HintConfigArray = [string, { [key: string]: any }];
```
molant: Do we need to export these types? While we are here maybe we can do some cleaning?

sarvaje: It is not needed right now, we are not referencing it anywhere. Do you want me to remove the export?

molant: Yes, and any other not used elsewhere if you can.
Thanks!

PR #3371, opened 2019-11-20
Breaking: Remove shortcuts

sarvaje:
## Pull request checklist

Make sure you:

### DIFF
--- a/packages/extension-browser/src/content-script/connector.ts
+++ b/packages/extension-browser/src/content-script/connector.ts
@@ -10,13 +10,12 @@ import {
     restoreReferences,
     traverse
 } from '@hint/utils-dom';
-
 import {
-    ConnectorOptionsConfig,
     IConnector,
     FetchEnd,
     NetworkData
 } from 'hint/dist/src/lib/types';
+import { ConnectorOptionsConfig } from '@hint/utils';
 
 import { browser, document, eval, location, window } from '../shared/globals';
 import { Events } from '../shared/types';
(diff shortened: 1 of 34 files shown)


packages/extension-browser/src/content-script/connector.ts
```diff
@@ -10,13 +10,12 @@ import {
...
     traverse
 } from '@hint/utils-dom';
-
 import {
-    ConnectorOptionsConfig,
     IConnector,
     FetchEnd,
     NetworkData
 } from 'hint/dist/src/lib/types';
+import { ConnectorOptionsConfig } from '@hint/utils/dist/src/types/config';
```
molant: `/dist/src/types/...`  😞

Can we make the types a top level of the packages? Is this part of one of the pending `utils` issues we have opened?

sarvaje: The are at the top level of the packages, but this is the extension, and the utils package, so we need to use the full path.

molant: I believe that in talking with @antross we were OK to link to the root path now that everything was split even if that was a small bump on the extension size.

Can you give it a try and see how much bigger it becomes?

sarvaje: I didn't try lately, but the error we use to have using `@hint/utils`  wasn't only the size, it was that the build of the extension fails. But I will try again

sarvaje: we can use `@hint/utils` only if we are importing types, but if we are importing a method, then we need to use the path to the file.

sarvaje:
@molant, can we merge this?

molant:
Yes, can you take care of it? I won't have access to my laptop until later today

PR #3377, opened 2019-11-20
Fix: Stability issues with VS Code extension

molant:
## Pull request checklist

Make sure you:

The VS Code extension could end up breaking `node_modules` in the
workspace or the global shared folder. This was caused because we were
installing `hint@latest` in the workspace and because that process
took too long and the user might closed the editor before it was
ended.
To avoid that we now let the user manage their own `hint` installation
in the workspace, no more prompting about installation. If no version
can be loaded from the workspace, the shared one (`hint@latest`) will
be used.
Also, the shared version is updated after 2 minutes instead of
immediately and only if the shared version has been successfully
loaded.
If the user already has a workspace version there is no point in
updating the shared one as it will not be used.

### DIFF
--- a/packages/extension-vscode/src/utils/analytics.ts
+++ b/packages/extension-vscode/src/utils/analytics.ts
@@ -103,6 +103,7 @@ export const trackSave = (uri: string, languageId: string) => {
     }
 
     // Throttle tracking saves to reduce redundant reports when autosave is on.
+    /* istanbul ignore if */
     if (lastSave && now - lastSave < twoMinutes) {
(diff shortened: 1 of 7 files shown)


molant:
vsix for testing (you'll have to remove the .zip extension and side load it)
[vscode-webhint-1.4.1.vsix.zip](https://github.com/webhintio/hint/files/3872028/vscode-webhint-1.4.1.vsix.zip)

sarvaje:
@molant there is a coverage error :(

PR #3529, opened 2020-02-07
Fix: Check for localStorage availability

### LINKED ISSUE #3494: [Bug] Chrome extension fails to read `localStorage` when 3p cookies are disabled.
# 🐛 Bug report

## Description

Extension does not work [when third-party cookies are blocked](https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document) in Chrome (79.0.3945.130):

`Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

webhint-not-working-if-3p-cookies-disallowed

## Environment

* __webhint version:__ 2.0.1

<hr>

This issue should ultimately be solved on Chrome's side, see: https://bugs.chromium.org/p/chromium/issues/detail?id=971848

Hirse:
## Pull request checklist

Make sure you:

Fall back to non-persistent in-memory cache if `localStorage` is not available.
Fixes #3494

chromium: [webhint-2.0.3.zip](https://github.com/webhintio/hint/files/4172602/webhint-2.0.3.zip)

### DIFF
--- a/packages/extension-browser/src/devtools/utils/storage.ts
+++ b/packages/extension-browser/src/devtools/utils/storage.ts
@@ -1,6 +1,30 @@
+let storage: {
+    getItem(key: string): string | null;
+    setItem(key: string, value: string): void;
+};
+
+try {
+    storage = {
+        getItem: localStorage.getItem.bind(localStorage),
+        setItem: localStorage.setItem.bind(localStorage)
+    };
+} catch (error) {
+    console.warn(`LocalStorage not available, falling back to in-memory storage: ${error}`);
+    const cache = new Map<string, string>();
+
+    storage = {
+        getItem (key) {
+            return cache.get(key) || null;
+        },
+        setItem (key, value) {
+            cache.set(key, value);
+        }
+    };
+}
+
 /* istanbul ignore next*/
 export const getItem = (key: string): any => {
-    const value = localStorage.getItem(key);
+    const value = storage.getItem(key);
 
     try {
         return value ? JSON.parse(value) : undefined;
@@ -13,5 +37,5 @@ export const getItem = (key: string): any => {
 
 /* istanbul ignore next*/
 export const setItem = (key: string, value: any): void => {
-    localStorage.setItem(key, JSON.stringify(value));
+    storage.setItem(key, JSON.stringify(value));
 };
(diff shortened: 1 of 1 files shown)


packages/extension-browser/src/devtools/utils/storage.ts
```diff
@@ -1,6 +1,30 @@
+let storage: {
+    getItem(key: string): string | null;
+    setItem(key: string, value: string): void;
+};
```
antross: Let's pull the signatures from the built-in `Storage` type, but still scoped to just the methods we want.
```suggestion
let storage: Pick<Storage, 'getItem' | 'setItem'>;
```

packages/extension-browser/src/devtools/utils/storage.ts
```diff
@@ -1,6 +1,30 @@
...
+
+try {
+    storage = {
+        getItem: localStorage.getItem.bind(localStorage),
+        setItem: localStorage.setItem.bind(localStorage)
+    };
+} catch (error) {
+    console.warn(`LocalStorage not available, falling back to in-memory storage: ${error}`);
+    const cache = new Map<string, string>();
```
antross: We'll know for sure once CI finishes, but I think our linter is configured to require a new line before `const`.
```suggestion

    const cache = new Map<string, string>();
```

packages/extension-browser/src/devtools/utils/storage.ts
Hirse: Didn't know `Pick` before. This is awesome.

packages/extension-browser/src/devtools/utils/storage.ts
Hirse: ESLint doesn't complain and it doesn't seem to need one here:
https://github.com/webhintio/hint/blob/98861bac7336d535ae100d680497a01e9ad3113a/packages/extension-browser/src/devtools/utils/storage.ts#L2-L3

antross: CI passed so I must have been thinking of a different rule 🤷‍♂️

sarvaje: Maybe you are thinking on `return`, the linter is configured you require a new line after `const`

Hirse:
@antross Attached the zip and used `Pick`

PR #3530, opened 2020-02-07
Chore: Allow running tests with only puppeteer installed

antross:
## Pull request checklist

Make sure you:

### DIFF
--- a/packages/connector-puppeteer/src/connector.ts
+++ b/packages/connector-puppeteer/src/connector.ts
@@ -103,9 +103,13 @@ export default class PuppeteerConnector implements IConnector {
             options.headless :
             isCI || getPlatform() === 'wsl';
 
-        const executablePath = 'browser' in options ?
-            getInstallationPath({ browser: options.browser }) :
-            getInstallationPath();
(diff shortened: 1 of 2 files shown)


packages/connector-puppeteer/src/connector.ts
```diff
@@ -103,9 +103,13 @@ export default class PuppeteerConnector implements IConnector {
             options.headless :
             isCI || getPlatform() === 'wsl';
 
-        const executablePath = 'browser' in options ?
-            getInstallationPath({ browser: options.browser }) :
-            getInstallationPath();
+        let executablePath: string | undefined;
+
+        if (!options.puppeteerOptions || !('executablePath' in options.puppeteerOptions)) {
```
molant: Can we use null coalescing here?

antross: I don't think we can do so cleanly since this is an `in` type guard. The closest would be something like:
```ts
if (!('executablePath' in (options.puppeteerOptions ?? {}))) {
```
But I don't think that reads very well.

PR #3533, opened 2020-02-07
Fix: Report all cookie errors

### LINKED ISSUE #2642: Update hint for Set-Cookie header to report all issues up front
From @JamoCA in #2641

> This hint wasn't reported until after manually fixing both "httponly" and "secure" hints. Shouldn't this hint always be displayed if the date format is invalid regardless of the "httponly" & "secure" status?

Good feedback. Currently the hint checks a few different categories of problems and exits early if any particular category fails, but that's not necessarily the optimal behavior as you point out.

Hirse:
## Pull request checklist

Make sure you:

Report all cookie errors instead of exiting early when a category reports errors.

Fixes #2642

### DIFF
--- a/packages/hint-validate-set-cookie-header/src/hint.ts
+++ b/packages/hint-validate-set-cookie-header/src/hint.ts
@@ -340,17 +340,11 @@ export default class ValidateSetCookieHeaderHint implements IHint {
 
                 parsedSetCookie.resource = resource;
 
-                defaultValidators.every((defaultValidator) => {
-                    const messages: ValidationMessages = defaultValidator(parsedSetCookie);
-
-                    if (messages.length) {
-                        reportBatch(messages, codeLanguage, codeSnippet);
-
-                        return false;
-                    }
-
-                    return true;
+                const messages = defaultValidators.flatMap((defaultValidator) => {
+                    return defaultValidator(parsedSetCookie);
                 });
+
+                reportBatch(messages, codeLanguage, codeSnippet);
             }
         };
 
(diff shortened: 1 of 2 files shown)


packages/hint-validate-set-cookie-header/src/hint.ts
```diff
@@ -340,17 +340,11 @@ export default class ValidateSetCookieHeaderHint implements IHint {
...
-                    const messages: ValidationMessages = defaultValidator(parsedSetCookie);
-
-                    if (messages.length) {
-                        reportBatch(messages, codeLanguage, codeSnippet);
-
-                        return false;
-                    }
-
-                    return true;
+                const messages = defaultValidators.flatMap((defaultValidator) => {
```
antross: The `flatMap` usage threw me at first because we only have a single-depth array of validators (I missed they each return an array of messages). Not sure if a comment might help others here.

Unfortunately I think we're going to have to use something else because [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap), `flatMap` is only supported in Node 11+ (and we currently support Node 10+ in webhint). CI didn't catch this because we only run Node 12 on PRs currently.

You can probably fairly easily switch to using `flatMap` from Lodash if you want as we use Lodash in a number of other places. Just be sure to use the `import flatMap = require('lodash/flatMap');` syntax as that helps us avoid bundling all of Lodash with webhint's browser extension (we'd like to switch to ESM bundling with tree shaking in the future, but haven't gotten there yet).

Hirse: @antross I changed it to use `reduce`.

PR #3568, opened 2020-02-24
New: Try puppeteer's Chromium if no browser is specified

antross:
## Pull request checklist

Make sure you:

Expand the cases where webhint will "just work" by allowing it to
use the version of Chromium bundled with a local puppeteer
installation if one is available.

This will also fix current CI issues on Mac which are failing due to
an inability to find a Chrome installation (puppeteer is present).

### DIFF
--- a/packages/connector-puppeteer/src/lib/chromium-finder.ts
+++ b/packages/connector-puppeteer/src/lib/chromium-finder.ts
@@ -337,7 +337,8 @@ const resolveChromiumPath = () => {
 
 /**
  * Searchs for a valid Chromium browser from the ones supported. The current priority list is:
- * `Chrome Canary`, `Chrome`, `Chromium`, `Edge Canary`, `Edge Dev` (`Edge` only on `win32 platforms).
+ * `Puppeteer`, `Chrome Canary`, `Chrome`, `Chromium`, `Edge Canary`, `Edge Dev` (`Edge` only
+ * on `win32 platforms).
  *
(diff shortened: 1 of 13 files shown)


packages/connector-puppeteer/src/lib/chromium-finder.ts
```diff
@@ -337,7 +337,8 @@ const resolveChromiumPath = () => {
 
 /**
  * Searchs for a valid Chromium browser from the ones supported. The current priority list is:
- * `Chrome Canary`, `Chrome`, `Chromium`, `Edge Canary`, `Edge Dev` (`Edge` only on `win32 platforms).
+ * `Puppeteer`, `Chrome Canary`, `Chrome`, `Chromium`, `Edge Canary`, `Edge Dev` (`Edge` only
+ * on `win32 platforms).
```
molant: Now that it is also in mac we should fix this...

antross:
> Are you going to remove your previous workaround for using puppeteer in CI?

Yeah, I'll do a follow-up commit once CI passes on Mac so don't merge yet. I just wanted to make sure this worked first.

packages/connector-puppeteer/src/lib/chromium-finder.ts
antross: Good point - didn't even notice that part of the comment. I assume there's work to do here to find the right path so we'll probably want to log a bug.

PR #3593, opened 2020-03-06
Breaking: Update APIs to support axe-core

antross:
## Pull request checklist

Make sure you:

Add and update APIs to better align with DOM standards.
Updates aim to provide sufficient support to run axe-core.

### DIFF
--- a/packages/utils-dom/src/create-element.ts
+++ b/packages/utils-dom/src/create-element.ts
@@ -1,0 +1,40 @@
+import { HTMLAnchorElement } from './elements/htmlanchorelement';
+import { HTMLBodyElement } from './elements/htmlbodyelement';
+import { HTMLHtmlElement } from './elements/htmlhtmlelement';
+import { HTMLElement } from './htmlelement';
+import { HTMLDocument } from './htmldocument';
+import { ElementData } from './types';
+
+const getElementType = (name: string): typeof HTMLElement => {
+    switch (name) {
+        case 'a':
+            return HTMLAnchorElement;
+        case 'body':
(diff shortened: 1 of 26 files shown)


packages/utils-dom/src/create-element.ts
```diff
@@ -0,0 +1,36 @@
...
+import { ElementData } from './types';
+
+const getElementType = (name: string): typeof HTMLElement => {
+    switch (name) {
+        case 'a':
+            return HTMLAnchorElement;
+        case 'body':
+            return HTMLBodyElement;
+        case 'html':
+            return HTMLHtmlElement;
```
molant: 🤯

packages/utils-dom/src/node.ts
```diff
@@ -0,0 +1,121 @@
...
+                return this._node.name.toUpperCase();
+            case 'text':
+                return '#text';
+            /* istanbul ignore next */
+            default:
+                throw new Error(`Unrecognized node type ${(this._node as any).type}`);
+        }
+    }
+
+    public get nodeType(): number {
```
molant: Will be good to point to the spec just for reference

antross: I'd prefer to reference MDN (which itself links to the related spec or specs). I find it to be a better place to start to understand the basics of what an API does. What do you think?

E.g. for this case the link would be
https://developer.mozilla.org/docs/Web/API/Node/nodeType

molant: SGTM!

PR #3711, opened 2020-04-29
Breaking: Remove redundant data and simplify reported messages

### LINKED ISSUE #3693: Remove URLs and code snippets from messages
We should review and simplify messages reported from hints to clean up the UX across entry points.

Why?
1. Redundant with URLs and code snippets provided in report metadata
2. Causes some messages to be VERY long
3. Prevents collapsing similar messages

antross:
## Pull request checklist

Make sure you:

The goal of simplifying reported messages is two-fold:
1. Make messages easier to read by eliminating long blocks of markup and URLs (which are repeated in the expanded details of a message).
2. Increase the ability for similar messages to be collapsed, giving users an improved overview of the number of unique issue types reported.

I've also made editorial changes to messages for clarity and consistency as I see opportunities. Feel free to provide additional feedback to this end. Since we're changing the messages anyway, now is a great time to ensure we're happy with the precise wording.

Fixes #3693

### DIFF
--- a/packages/hint-http-cache/src/_locales/en/messages.json
+++ b/packages/hint-http-cache/src/_locales/en/messages.json
@@ -33,7 +33,7 @@
     },
     "staticResourceCacheValue": {
         "description": "Report message when a static resource has an short cache value",
-        "message": "Static resources should use a 'cache-control' header with a long cache value (max-age=$1)."
+        "message": "Static resources should use a 'cache-control' header with 'max-age=$1' or more."
     },
     "targetShouldNotBeCached": {
         "description": "Report message when the target shouldn't be cached",
--- a/packages/hint-http-cache/src/_locales/en/messages.json
+++ b/packages/hint-http-cache/src/_locales/en/messages.json
@@ -5,42 +5,42 @@
     },
     "directiveInvalid": {
         "description": "Report message when the directives are invalid",
-        "message": "The $1 $2 $3 invalid"
+        "message": "A 'cache-control' header contains invalid directives: $1"
     },
     "directiveInvalidValue": {
         "description": "Report message when the directives have invalid value",
-        "message": "The following $1 an invalid value:\n$2"
+        "message": "A 'cache-control' header contains directives with invalid values: $1"
     },
     "directiveNotRecomended": {
         "description": "Report message when a directive is not recomended",
-        "message": "The directive \"$1\" is not recommended"
+        "message": "A 'cache-control' header contains directives which are not recommended: $1"
     },
     "name": {
         "description": "Metadata name",
         "message": "HTTP cache"
     },
     "noCacheBustingPattern": {
-        "description": "Report message when a cache busting pattern is not configured",
-        "message": "No configured patterns for cache busting match $1. See docs to add a custom one."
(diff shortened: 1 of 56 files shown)


molant:
Haven't reviewed yet, but are there any other breaking changes we want to do while we are at it?

antross:
> Haven't reviewed yet, but are there any other breaking changes we want to do while we are at it?

None that are top-of-mind. Happy to debate to what degree we want to consider this breaking too. By default this will only be breaking at the individual hint level. 

Also, I personally don't feel the exact text of a message is a contract at the semver level (especially once you throw i18n in the mix), but I erred on the conservative side with marking these as breaking for now given our prior conversations. I figure we can debate this a bit before actually merging the changes.

That said, we currently don't expose the concept of an error code, so there's nothing more granular than a hint ID to use to recognize an individual error unless you look at the message. Perhaps that's something we should consider.

antross:
@hxlnt, @johnemau, @molant: This is ready for review when you get the chance.

If you want you can self-host the changes by side-loading this updated version of the browser extension:
* [webhint-2.1.1.zip](https://github.com/webhintio/hint/files/4600799/webhint-2.1.1.zip)

packages/hint-http-cache/src/_locales/en/messages.json
```diff
@@ -5,42 +5,42 @@
     },
     "directiveInvalid": {
         "description": "Report message when the directives are invalid",
-        "message": "The $1 $2 $3 invalid"
+        "message": "A 'cache-control' header contains invalid directives: $1"
```
molant, line 8: Should we wrap the directives with `'`  or `"` for consistency and make it even clearer?

packages/hint-no-http-redirects/src/_locales/en/messages.json
```diff
@@ -7,12 +7,8 @@
...
         "message": "Avoid HTTP redirects"
     },
-    "redirectDectected": {
-        "description": "Report message when a redirect are detected",
-        "message": "$1 redirect detected for '$2' (max is $3)."
-    },
     "redirectsDectected": {
         "description": "Report message when a redirect are detected",
-        "message": "$1 redirects detected for '$2' (max is $3)."
+        "message": "Too many HTTP redirects detected (max is $1)."
```
molant, line 12: 👏

packages/hint-http-cache/src/_locales/en/messages.json
antross: Were you thinking of quoting the directives as a group or individually (e.g. when there's more than one)?

molant: I think group should be fine as they are part of the same header value (and it should be easier to implement)? Don't have a strong opinion though.

antross: Ok, I went with quoting the individual values.

antross:
@hxlnt do you have any feedback or does this look good to merge?

@molant do you have a preference whether this is squashed into a single breaking commit (since it's effectively the same intent across packages) or remains a series of per-package commits?

molant:
Squashing is fine 👍

PR #3776, opened 2020-05-25
Add a few special disallowed headers

### LINKED ISSUE #1096: Improve `hint-no-disallowed-headers`
Look over: https://www.fastly.com/blog/headers-we-dont-want, and see if we can add more unneeded headers to the list.

The code for this hint is under `packages/hint-no-disallowed-headers/`

captainbrosset:
This PR should fix #1096.

It adds warnings for the following HTTP response headers:

* Expires
* Host
* P3P
* Pragma
* Via
* X-Frame-Options
* X-UA-Compatible

These were added based on https://www.fastly.com/blog/headers-we-dont-want.

Instead of adding these headers to the existing list of disallowed headers, I added specific checks for these ones. The idea is, this way, we can have specific messages for each of them.

I created a couple of preliminary commits to prepare the code for this, then added the logic for this in a third commit, and finished adding all warnings in a 4th commit.
Finally, I made sure the ignored headers, as configured by users, would also ignore these special headers.

### DIFF
--- a/packages/hint-no-disallowed-headers/src/_locales/en/messages.json
+++ b/packages/hint-no-disallowed-headers/src/_locales/en/messages.json
@@ -11,6 +11,30 @@
         "description": "Report message when the 'server' header contains more than the server name",
         "message": "The 'server' header should only contain the server name."
     },
+    "disallowedExpiresHeader": {
+        "description": "Report message when the response includes the Expires header",
+        "message": "The 'Expires' header should not be used, 'Cache-Control' should be preferred."
+    },
+    "disallowedHostHeader": {
+        "description": "Report message when the response includes the Host header",
+        "message": "The 'Host' header should not be used, it is a request header only."
+    },
+    "disallowedP3PHeader": {
+        "description": "Report message when the response includes the P3P header",
+        "message": "The 'P3P' header should not be used, it is a non-standard header only implemented in Internet Explorer."
+    },
+    "disallowedPragmaHeader": {
+        "description": "Report message when the response includes the Pragma header",
+        "message": "The 'Pragma' header should not be used, it is deprecated and is a request header only."
+    },
+    "disallowedViaHeader": {
+        "description": "Report message when the response includes the Via header",
+        "message": "The 'Via' header should not be used, it is a request header only."
+    },
+    "disallowedXFrameOptionsHeader": {
+        "description": "Report message when the response includes the X-Frame-Options header",
+        "message": "The 'X-Frame-Options' header should not be used. A similar effect, with more consistent support and stronger checks, can be achieved with the 'Content-Security-Policy' header and 'frame-ancestors' directive."
+    },
     "name": {
         "description": "Metadata name",
         "message": "Disallowed HTTP headers"
--- a/packages/hint-no-disallowed-headers/src/hint.ts
+++ b/packages/hint-no-disallowed-headers/src/hint.ts
@@ -17,9 +17,61 @@ import { FetchEnd, IHint } from 'hint/dist/src/lib/types';
 import { Severity } from '@hint/utils-types';
 
 import meta from './meta';
-import { getMessage } from './i18n.import';
+import { getMessage, MessageName } from './i18n.import';
 
 const debug = d(__filename);
+
+/*
+ * ------------------------------------------------------------------------------
+ * Private
+ * ------------------------------------------------------------------------------
+ */
+
+const serverHeaderContainsTooMuchInformation = (serverHeaderValue: string): boolean => {
+    const regex = [
+        /*
+         * Version numbers.
+         *
+         * e.g.:
+         *
+         *   apache/2.2.24 (unix) mod_ssl/2.2.24 openssl/1.0.1e-fips mod_fastcgi/2.4.6
+         *   marrakesh 1.9.9
+         *   omniture dc/2.0.0
+         *   microsoft-iis/8.5
+         *   pingmatch/v2.0.30-165-g51bed16#rel-ec2-master i-077d449239c04b184@us-west-2b@dxedge-app_us-west-2_prod_asg
+         */
+
+        /\/?v?\d\.(\d+\.?)*/,
+
+        /*
+         * OS/platforms names (usually enclose between parentheses).
+         *
+         * e.g.:
+         *
+         *  apache/2.2.24 (unix) mod_ssl/2.2.24 openssl/1.0.1e-fips mod_fastcgi/2.4.6
+         *  apache/2.2.34 (amazon)
+         *  nginx/1.4.6 (ubuntu)
+         */
+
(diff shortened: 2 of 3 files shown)


captainbrosset:
Note that I did sign the CLA, but with my microsoft email address. I just added it to my github account now. So maybe the licence/cla bot needs to re-run or something.

antross:
I'm not sure why the CLA bot didn't pick up the change here, but I can see it worked correctly for your second PR. We can just overrule the bot for this one rather than jump through hoops to figure out how to get it to reset.

packages/hint-no-disallowed-headers/src/_locales/en/messages.json
```diff
@@ -11,6 +11,34 @@
...
+    },
+    "disallowedViaHeader": {
+        "description": "Report message when the response includes the Via header",
+        "message": "The 'Via' header should not be used, it is a request header only."
+    },
+    "disallowedXFrameOptionsHeader": {
+        "description": "Report message when the response includes the X-Frame-Options header",
+        "message": "The 'X-Frame-Options' header should not be used, the same effect can be achieved with more consistent support using the 'Content-Security-Policy' header."
+    },
+    "disallowedXUACompatibleHeader": {
```
antross, line 38: Apologies for missing this in our earlier discussion, but `X-UA-Compatible` overlaps with our recommendations from `hint-highest-available-document-mode` (https://webhint.io/docs/user-guide/hints/hint-highest-available-document-mode/). 

That hint is a touch more nuanced: it recommends removing the meta tag or header if targeting IE11+, but recommends including it if for some crazy reason you're still targeting IE10 or below to ensure you're running in the latest standards mode.

We should probably not include `X-UA-Compatible` as part of this hint to avoid redundant reports.

captainbrosset: Good to know!
I changed the code to not include `X-UA-Compatible`.

packages/hint-no-disallowed-headers/src/_locales/en/messages.json
```diff
@@ -11,6 +11,30 @@
         "description": "Report message when the 'server' header contains more than the server name",
         "message": "The 'server' header should only contain the server name."
     },
+    "disallowedExpiresHeader": {
+        "description": "Report message when the response includes the Expires header",
+        "message": "The 'Expires' header should not be used, 'Cache-Control' should be preferred."
```
antross, line 16: Thanks again for breaking these out! I really appreciate having actionable messages, especially when we're suggesting developers use an alternative (as opposed to simply removing a header).

packages/hint-no-disallowed-headers/src/hint.ts
```diff
@@ -43,75 +94,53 @@ export default class NoDisallowedHeadersHint implements IHint {
...
-
-                /(mod_|openssl|php)/
-            ];
-
-            return regex.some((r) => {
-                return r.test(serverHeaderValue);
-            });
-        };
+        const validateGeneralHeaders = ({ response, resource }: FetchEnd) => {
+            const headers: string[] = includedHeaders(response.headers, generalDisallowedHeaders);
```
antross, line 122: You don't actually need to change this since it was pre-existing code you moved, but I wanted to call out we've been slowly moving toward dropping redundant types in our existing code when we make updates. Keeps things a bit easier to read and closer to JS. Probably worth us finding/creating a linting rule at some point to track down the rest of these...
```suggestion
            const headers = includedHeaders(response.headers, generalDisallowedHeaders);
```

packages/hint-no-disallowed-headers/src/hint.ts
```diff
@@ -27,13 +27,64 @@ const debug = d(__filename);
...
 
+interface ISpecialHeadersWithMessage {
+    [key: string]: MessageName;
+}
+
 export default class NoDisallowedHeadersHint implements IHint {
 
     public static readonly meta = meta;
 
+    private serverHeaderContainsTooMuchInformation(serverHeaderValue: string): boolean {
```
antross, line 38: Nit: I'm glad to see this moved out of the constructor, but since it doesn't use any instance state can we either make it `static` or move it outside of the class entirely as a helper function (the latter being my personal preference)?

packages/hint-no-disallowed-headers/src/_locales/en/messages.json
```diff
@@ -11,6 +11,30 @@
...
+        "description": "Report message when the response includes the Pragma header",
+        "message": "The 'Pragma' header should not be used, it is deprecated and is a request header only."
+    },
+    "disallowedViaHeader": {
+        "description": "Report message when the response includes the Via header",
+        "message": "The 'Via' header should not be used, it is a request header only."
+    },
+    "disallowedXFrameOptionsHeader": {
+        "description": "Report message when the response includes the X-Frame-Options header",
+        "message": "The 'X-Frame-Options' header should not be used, the same effect can be achieved with more consistent support using the 'Content-Security-Policy' header."
```
Malvoz, line 36: I suggest explicitly mentioning the [`frame-ancestors`](https://www.w3.org/TR/CSP/#directive-frame-ancestors) navigation directive here as it might save developers some time if they're unfamiliar with it.

(Also, while I'm not sure there's much benefit in describing it here, `frame-ancestors` and `X-Frame-Options` doesn't have the _exact_ same effect, `frame-ancestors` is stronger in that it checks against _any_ ancestors where `X-Frame-Options` only checks the top-level document's location as described here: https://www.w3.org/TR/CSP/#frame-ancestors-and-frame-options)

packages/hint-no-disallowed-headers/src/hint.ts
captainbrosset: Thanks for letting me know. I'll do this change since it's so simple. And I'll keep that in mind for the future.

packages/hint-no-disallowed-headers/src/hint.ts
captainbrosset: Ah, thanks, I wasn't sure whether to do this or not. But I agree with you, I'd prefer for this to be outside of the class entirely.
I'll address this in this PR soon.

packages/hint-no-disallowed-headers/src/_locales/en/messages.json
captainbrosset: Thanks, it makes sense to me to mention the `frame-ancestors` directive. I'll update the message.

captainbrosset:
Update: I've addressed all remaining review comments, and squashed the whole thing down to 1 commit.
I guess it's ready for a last round of reviews and/or merge whenever CI has completed.

antross:
Thanks @captainbrosset, all the updates look good to me!

@molant, @johnemau, @utsavized - do any of you have time to take a look?

PR #3777, opened 2020-05-25
Fix: Better wrapping logic for config page

### LINKED ISSUE #3528: [Bug] Items in configuration page wrap poorly at smaller widths in browser extension
At certain sizes just before the layout changes from horizontal to vertical, items in the configuration page wrap poorly:

image

This is probably most easily fixed by adjusting the media queries being used to snap to the vertical layout slightly sooner.

captainbrosset:
This should fix #3528.

Using CSS grid minmax, we can make the various columns
in the config page be at least some predefined width that
avoids poor wrapping of form fields, and at most take all
the size they need.

Now, to avoid them stretching too wide and becoming hard to
read, I preserved the max-width that was already there, but
moved it to the parent element to avoid conflicting with
the auto-fit logic (see below), as this was causing overflow.

We're using auto-fit to make the number of columns adapt to
the available space.

### DIFF
--- a/packages/extension-browser/src/devtools/views/controls/valid-input.css
+++ b/packages/extension-browser/src/devtools/views/controls/valid-input.css
@@ -5,6 +5,7 @@
     color: var(--base-color);
     font-size: var(--font-size-sm);
     padding: 0.1875rem 0.75rem 0.3125rem; /* 3px 12px 5px */
+    width: 8.25rem; /* 132px */
 }
 
 .root:focus {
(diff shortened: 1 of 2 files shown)


captainbrosset:
For info, this is how things look with the fix applied:

config-wrap

antross:
> I'd be okay fixing this either by explicitly shortening the input width or by making our min-width for the columns a touch larger to accommodate.

On second thought we'd better just go ahead and set an explicit width on the input elements. Who knows what we might get on other machines...

captainbrosset:
Good catch @antross . Indeed for some reason I'm getting a different width on Firefox that avoids the problem. I tested on Chrome canary and did reproduce the issue.
So this updated commit introduces a fixed width for the ValidInput component. I wasn't sure if I better set the width on that component directly or only override it at the Config screen level, but it looks like it's only ever used there anyway.
Let me know what you think.

PR #3782, opened 2020-05-28
Fix: Don't delay FetchEnd for DOM snapshot

### LINKED ISSUE #3771: [Bug] Network related hints are delayed until page load completes
[utils-worker] We need to send the `fetchEnd` event without waiting for the document.

See:
https://github.com/webhintio/hint/blob/master/packages/utils-worker/src/connector.ts#L112

Most network hints do not need the document, the hints use the document to link to elements in the results, so passing null will lose that but provide results faster.

[feat request] Long term we should lazy-load the element or re-report the hint result when the document is ready.

johnemau:
## Pull request checklist

Make sure you:

[util-worker]

Remove FetchEnd event queuing which was previously implemented to ensure the document element was ready (via a DOM snapshot). The document element is optional so this created an unnecessary delay especially for pages with large DOMs.

In the future we should introduce a mechanism to send an updated FetchEnd event when the DOM snapshot is ready.

Fixes #3771

### DIFF
--- a/packages/utils-worker/src/connector.ts
+++ b/packages/utils-worker/src/connector.ts
@@ -23,7 +23,6 @@ export default class WebWorkerConnector implements IConnector {
     private _document: HTMLDocument | undefined;
     private _originalDocument: HTMLDocument | undefined;
     private _engine: Engine;
-    private _fetchEndQueue: FetchEnd[] = [];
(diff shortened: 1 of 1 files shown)


johnemau:
Tracking follow-up work with #3784

Issue #3479, opened 2020-01-05
[Bug] Supported CSS features not flagging 'display: subgrid'


ststimac:
# 🐛 Bug report
I was running the browser extension against a subgrid demo, https://stephaniestimac.com/css-grid-experiments/project5/good-omens-subgrid, and "display: subgrid" isn't being flagged as not supported in browsers (except firefox, it's supported there).

molant:
IIRC we treat display values as progressive enhancement to generate less noise but maybe we have to review this.
How broken is the experience in other browsers?
Are you using @supports or similar?


________________________________

antross:
Double-checking the demo code it looks like `grid-template-columns: subgrid;` is the value that isn't getting flagged. 

Checking on MDN, it appears [`grid-template-rows`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows) has browser support listed for `subgrid`, but [`grid-template-columns`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns) does not.

So we'll need to file an issue (or PR) on browser-compat-data to add a `subgrid` entry for `grid-template-columns`: https://github.com/mdn/browser-compat-data/blob/master/css/properties/grid-template-columns.json, probably matching the one from `grid-template-rows`: https://github.com/mdn/browser-compat-data/blob/master/css/properties/grid-template-rows.json#L465

molant:
> So we'll need to file an issue (or PR) on browser-compat-data to add

@ststimac can you please open the issue? You can probably quote this issue when opening and tag @antross in there too.

captainbrosset:
The `subgrid` browser support  information has, since then, been updated on MDN.
So webhint now correctly reports a warning for that `grid-template-columns: subgrid;` usage.

However, I think there is still kind of a problem: the warning I'm seeing is: `'grid-template-columns' is not supported by Internet Explorer`. While that is true, it is supported on other browsers, except for the `subgrid` value which is only available in Firefox.

Should a different warning be emitted for this special case? Does webhint only warn on property names, or can it also check values?

antross:
Nevermind, I was using an old build. Yes, we do flag this correctly when excluding IE, but not when IE is included:
![image](https://user-images.githubusercontent.com/785009/83273588-47325700-a192-11ea-8eb9-61cbb7691fdc.png)

Digging in, indeed if anything is flagged as unsupported at the declaration level, we don't currently drill down into the value:
```ts
[...]

PR #3806, opened 2020-06-09
Fix: Correctly drill down to CSS values when needed

captainbrosset:
This is a fix proposal for #3479. There are multiple ways for doing this, and the code for it is deviating a bit from the simple AST node visiting logic, so let me know if you have a better idea.

---

Until now, if a CSS property was not supported on at least one browser
the `compat-api/css` hint would not go any further and would report this.

With this change, the hint now checks if at least some browsers did
support the property name, and if so then checks the value on those browsers
too.

This means that something like `grid-template-rows:subgrid` would lead
to 2 reports:

* one for `grid-template-rows` on browsers that don't support it like IE,
* and one for `grid-template-rows:subgrid` on browsers that do support the property
but not this particular value (like Edge 16).

### DIFF
--- a/packages/hint-compat-api/src/css.ts
+++ b/packages/hint-compat-api/src/css.ts
@@ -67,8 +67,12 @@ const validateAtRule = (node: AtRule, context: Context): ReportData | null => {
     return null;
 };
 
-const validateDeclValue = (node: Declaration, context: Context): ReportData | null => {
-    const unsupported = getUnsupportedDetails({ property: node.prop, value: node.value }, context.browsers);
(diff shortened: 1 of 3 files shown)


captainbrosset:
Thanks @molant!
I know @antross is going to be away for some time. Are you comfortable merging this before he reviews? Or perhaps somebody else can take a look too, maybe @sarvaje?
I'm not in a rush though!

molant:
Let's just merged. I've pointed to the issue and this PR in the commit comment in case we want to revisit this later if this causes problems.

Thanks a lot!

PR #3830, opened 2020-06-16
New: Add endpoint/data contract to log telemetry to internal api

### LINKED ISSUE #3829: [Feature] Update utils-telemetry package with new endpoint and data contract
Update utils-telemetry package with updated telemetry endpoint and also simplify data model.

utsavized:
Closes #3829

### DIFF
--- a/packages/utils-telemetry/package.json
+++ b/packages/utils-telemetry/package.json
@@ -65,5 +65,5 @@
     "watch": "npm run build && npm-run-all --parallel -c watch:*",
     "watch:ts": "npm run build:ts -- --watch"
   },
-  "version": "1.1.0"
+  "version": "1.0.3"
 }
--- a/packages/utils-telemetry/package.json
+++ b/packages/utils-telemetry/package.json
@@ -65,5 +65,5 @@
     "watch": "npm run build && npm-run-all --parallel -c watch:*",
     "watch:ts": "npm run build:ts -- --watch"
   },
-  "version": "1.0.3"
+  "version": "1.1.0"
 }
--- a/packages/utils-telemetry/src/telemetry-api.ts
+++ b/packages/utils-telemetry/src/telemetry-api.ts
@@ -3,7 +3,7 @@ type Properties = { [key: string]: string };
 
 type TelemetryItem = {
     data: {
-        measurements: Measurements,
+        measurements: Measurements;
         name: string;
         properties: Properties;
     };
@@ -73,8 +73,9 @@ const track = async (type: string, data: TelemetryItem['data']) => {
             name: data.name,
             properties: { ...options.defaultProperties, ...data.properties }
         },
-        type: `${type}Data`,
-        time: new Date().toISOString()
+        time: new Date().toISOString(),
+        type: `${type}Data`
+
     });
 
     if (!options.batchDelay) {
--- a/packages/utils-telemetry/src/telemetry-api.ts
+++ b/packages/utils-telemetry/src/telemetry-api.ts
@@ -3,22 +3,16 @@ type Properties = { [key: string]: string };
 
 type TelemetryItem = {
     data: {
-        baseData: {
-            measurements: Measurements;
-            name: string;
-            properties: Properties;
-            ver?: number;
-        };
-        baseType: string;
+        measurements: Measurements,
+        name: string;
+        properties: Properties;
     };
-    iKey: string;
-    name: string;
+    type: string;
     time: string;
 };
 
-const appInsightsEndpoint = 'https://dc.services.visualstudio.com/v2/track';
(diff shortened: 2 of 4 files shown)


utsavized:
Bump. @sarvaje @molant

packages/utils-telemetry/package.json
```diff
@@ -65,5 +65,5 @@
     "watch": "npm run build && npm-run-all --parallel -c watch:*",
     "watch:ts": "npm run build:ts -- --watch"
   },
-  "version": "1.0.3"
+  "version": "1.1.0"
```
molant, line 68: You don't need to manually bump it. Packages in this repo get automatically bumped based on the commit history during the release 😊

packages/utils-telemetry/src/telemetry-api.ts
```diff
@@ -3,30 +3,23 @@ type Properties = { [key: string]: string };
...
+        properties: Properties;
     };
-    iKey: string;
-    name: string;
+    type: string;
     time: string;
 };
 
-const appInsightsEndpoint = 'https://dc.services.visualstudio.com/v2/track';
+const telemetryEndPoint = 'https://webhint-telemetry.azurewebsites.net/api/log';
```
molant, line 14: @hxlnt just want confirmation this is what we want: to use our Azure Function to report the telemetry to app insights so we can clean up any extra information that we do not want to store.
Is that correct?

packages/utils-telemetry/src/telemetry-api.ts
```diff
@@ -3,30 +3,23 @@ type Properties = { [key: string]: string };
...
 
-let nameKey = '';
 let sendTimeout: NodeJS.Timeout | null = null;
 let telemetryQueue: TelemetryItem[] = [];
 
 let options = {
     batchDelay: 15000,
     defaultProperties: {} as Properties,
     enabled: false,
-    instrumentationKey: '8ef2b55b-2ce9-4c33-a09a-2c3ef605c97d',
```
molant, line 29: Should we do any type of validation to make sure we are not getting spammed?

packages/utils-telemetry/package.json
utsavized: Ah, okay. I'll reset.

packages/utils-telemetry/src/telemetry-api.ts
utsavized: I also had this discussion with Tony, and the gist of it was that even AI endpoint was public and we could have been spammed before. So, we thought we wouldn't need it, but if we do feel like we want to track some spamming, we should do it as part a different task.

molant: we had an ID but agree that doesn't seem worth it :)

utsavized: We will notice a deliberate spamming attempt in the chart for sure, but it will be more of a reactive feature then, rather than a proactive feature if we do it right now. We can definitely keep an eye on it to see if it is necessary.

utsavized:
> @utsavized does your other telemetry PR ([webhintio/telemetry#190](https://github.com/webhintio/telemetry/pull/190)) need to be deployed first?

Yes. Telemetry, then hint, then webhint.io (that PR is coming after these get deployed) :)

packages/utils-telemetry/src/telemetry-api.ts
hxlnt: SGTM.

utsavized:
@molant Is this good to go?

packages/utils-telemetry/src/telemetry-api.ts
```diff
@@ -3,30 +3,23 @@ type Properties = { [key: string]: string };
 
 type TelemetryItem = {
     data: {
-        baseData: {
-            measurements: Measurements;
-            name: string;
-            properties: Properties;
-            ver?: number;
```
molant, line 10: Is the version added automatically when sending the telemetry? I know we have a few metrics around this.

utsavized: Is this tracking webhint version? A bunch of other places (including webhint.io) had this hardcoded to 2, so I thought this was the v2 of the app insights REST API.  If you look at Line 84 of this file (before change), you will see if set to 2.

molant: TBH, I don't know. @antross it he one that took care of the telemetry and now you so probably you know best how it works 😅

Let's get #3837 merged first and then this one, sounds good?

utsavized: Sounds good ... looks like that one is merged now, is this one good to go then?

Issue #3809, opened 2020-06-12
Rename master branches


hxlnt:
Per our team meeting on June 12, the team would like to do an _en masse_ rename of master branches throughout our repos. I think we informally talked about changing it to "main," which works for me.

We may need to do some research on how to unblock impacts on our build pipeline and other dependencies, if any.

molant:
This is done across the org except the archived repos, PRs included.

PR #3837, opened 2020-06-22
Chore: Move from master to main

molant:
Part of #3809 
All the other repos have been renamed but this one is a bit trickier so doing a PR in case someone catches something I missed.

Once this is merged the only remaining task would be to change the default branch and delete master.
If other PRs are merged in master before this PR should be rebased.


molant:
Windows timedout while installing dependencies but the others pass with no pbms.

@sarvaje can you take a look?

azure-pipelines.yml
```diff
@@ -1,18 +1,18 @@
-# Scope the validations to master and PRs to master
+# Scope the validations to main and PRs to main
 trigger:
-  - master
+    - main
```
sarvaje, line 3: any reason for the extra spaces?

molant: an autoformatter when saving 😠
Should be fixed now

sarvaje:
@molant the timeout issue looks like it was a temporary issue, the build is passing now.

PR #3839, opened 2020-06-24
Fix: Icon purpose doesn't support combination of values

### LINKED ISSUE #3791: web manifest does not support combination of values for icon purpose
The commit that added support for `maskable` is https://github.com/webhintio/hint/commit/c5eb40f819f79b6f92b18dcd015bfda5807bb034#diff-07d179b4d8516fb968d006563b37bfe3

Unfortunately, as @amazingcaio it does not allow for combinations:

> I don't think this issue is fixed. `"purpose"` can have multiple valid values, so `"any maskable"` and `"maskable any"`, and (I suppose) even `"any maskable badge"` should also be accepted. I think the real issue here is, because the possible options are hardcoded you'd have to have all 15 different possibilities (cos you have to treat `"any maskable badge"` and `"ask badge maskable"` as different, for instance). I reckon having to type 15 different possible options is fine, but it gets more complicated as more values are added to the web manifest spec. If one more value is added, for instance, there would be 60 different possible combinations. So changing the code so you don't have to hardcode every option and disregard the order of the values (so `"any maskable"` and `"maskable any"` would be considered the same) is the best solution.

Need to find a way to enable this in the schema and our types.

_Note: Original issue text by @tmb-github is available in the history_

sarvaje:
Fix #3791

## Pull request checklist

Make sure you:

Allow to the icons in manifest to have a combination of values.

### DIFF
--- a/packages/hint-manifest-icons/package.json
+++ b/packages/hint-manifest-icons/package.json
@@ -14,7 +14,8 @@
     "@hint/utils-i18n": "^1.0.3",
     "@hint/utils-types": "^1.0.1",
     "image-size": "^0.8.3",
-    "image-type": "^4.1.0"
+    "image-type": "^4.1.0",
+    "lodash": "^4.17.15"
   },
   "devDependencies": {
     "@hint/parser-manifest": "^2.2.18",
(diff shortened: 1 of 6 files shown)


sarvaje:
@webhintio/contributors anyone knows how to update the CLA, I didn't realize the email configured was wrong :s

molant:
@sa change the author in the commit and reopen the pr?

PR #3841, opened 2020-06-24
Fix: Icon purpose doesn't support combination of values

### LINKED ISSUE #3791: web manifest does not support combination of values for icon purpose

sarvaje:
Fix #3791

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint-manifest-icons/package.json
+++ b/packages/hint-manifest-icons/package.json
@@ -14,7 +14,8 @@
     "@hint/utils-i18n": "^1.0.3",
     "@hint/utils-types": "^1.0.1",
     "image-size": "^0.8.3",
-    "image-type": "^4.1.0"
+    "image-type": "^4.1.0",
+    "lodash": "^4.17.15"
   },
   "devDependencies": {
     "@hint/parser-manifest": "^2.2.18",
--- a/packages/hint-manifest-icons/package.json
+++ b/packages/hint-manifest-icons/package.json
@@ -14,8 +14,7 @@
     "@hint/utils-i18n": "^1.0.3",
     "@hint/utils-types": "^1.0.1",
     "image-size": "^0.8.3",
-    "image-type": "^4.1.0",
-    "lodash": "^4.17.15"
+    "image-type": "^4.1.0"
   },
   "devDependencies": {
     "@hint/parser-manifest": "^2.2.18",
--- a/packages/hint-manifest-icons/src/_locales/en/messages.json
+++ b/packages/hint-manifest-icons/src/_locales/en/messages.json
@@ -13,11 +13,11 @@
     },
     "iconPurposeDuplicate": {
         "description": "Report message when an icon has a duplicate value in the property purpose",
-        "message": "Duplicate value(s) '$1' found in icon purpose property."
+        "message": "Duplicate value(s) found in icon purpose porperty: $1."
     },
     "iconPurposeInvalid": {
         "description": "Report message when an icon has an invalid value in the property purpose",
-        "message": "Invalid icon property: $1."
+        "message": "Invalid value for icon property purpose: $1."
     },
     "iconShouldBeValidImageType": {
         "description": "Report message when the image type is not valid",
--- a/packages/hint-manifest-icons/src/_locales/en/messages.json
+++ b/packages/hint-manifest-icons/src/_locales/en/messages.json
@@ -17,7 +17,7 @@
     },
     "iconPurposeInvalid": {
         "description": "Report message when an icon has an invalid value in the property purpose",
-        "message": "Invalid icon property $1."
+        "message": "Invalid icon property: $1."
     },
     "iconShouldBeValidImageType": {
         "description": "Report message when the image type is not valid",
--- a/packages/hint-manifest-icons/src/_locales/en/messages.json
+++ b/packages/hint-manifest-icons/src/_locales/en/messages.json
@@ -13,7 +13,7 @@
     },
     "iconPurposeDuplicate": {
         "description": "Report message when an icon has a duplicate value in the property purpose",
-        "message": "Duplicate value(s) found in icon purpose porperty: $1."
+        "message": "Duplicate value(s) found in icon purpose property: $1."
     },
     "iconPurposeInvalid": {
         "description": "Report message when an icon has an invalid value in the property purpose",
--- a/packages/hint-manifest-icons/src/_locales/en/messages.json
+++ b/packages/hint-manifest-icons/src/_locales/en/messages.json
@@ -11,6 +11,14 @@
         "description": "Report message when an icon request return an status code !== 200",
         "message": "Icon could not be fetched (status code: $1)."
     },
+    "iconPurposeDuplicate": {
+        "description": "Report message when an icon has a duplicate value in the property purpose",
+        "message": "Duplicate value(s) '$1' found in icon purpose property."
+    },
+    "iconPurposeInvalid": {
+        "description": "Report message when an icon has an invalid value in the property purpose",
+        "message": "Invalid icon property $1."
+    },
     "iconShouldBeValidImageType": {
         "description": "Report message when the image type is not valid",
         "message": "Icon should be a valid image type $1"
@@ -44,7 +52,7 @@
         "message": "Sizes not specifed for icon"
     },
     "validIconsNotFound": {
-        "description": "Report message when ",
+        "description": "Report message when a valid icon is not found",
         "message": "Valid icons property was not found in the web app manifest"
     }
 }
--- a/packages/hint-manifest-icons/src/hint.ts
+++ b/packages/hint-manifest-icons/src/hint.ts
@@ -270,6 +270,7 @@ export default class ManifestIconHint implements IHint {
             return validSizes;
         };
 
+        // See https://w3c.github.io/manifest/#purpose-member
         const validateIconsPurpose = (icons: ManifestImageResource[], resource: string, getLocation: JSONLocationFunction) => {
             for (let iconIndex = 0; iconIndex < icons.length; iconIndex++) {
                 const icon = icons[iconIndex];
--- a/packages/hint-manifest-icons/src/hint.ts
+++ b/packages/hint-manifest-icons/src/hint.ts
@@ -282,9 +282,22 @@ export default class ManifestIconHint implements IHint {
                 const purposes = icon.purpose.split(' ').filter((purpose) => {
                     return purpose;
                 });
+                /*
+                 * Creating a set will remove automatically the duplicated items.
+                 */
                 const purposesSet = new Set(purposes);
 
+                /*
+                 * This loop will do a couple of tasks:
+                 *   1. Will check if the `purpose` is a valid one.
+                 *   2. It will remove from `purposes` the purpose it is checking,
+                 *      that will keep in `purposes` the duplicated ones.
+                 *
+                 * We are using the `purposesSet` to iterate because it won't have any
+                 * duplicate value and it will make duplicate detection easier.
+                 */
                 for (const purpose of purposesSet) {
+                    // Check if the purpose is valid.
                     if (!validPurposes.includes(purpose)) {
                         context.report(resource, getMessage('iconPurposeInvalid', context.language, purpose), {
                             location: getLocation(`icons[${iconIndex}].purpose`),
@@ -299,6 +312,10 @@ export default class ManifestIconHint implements IHint {
                     }
                 }
 
+                /*
+                 * Check if purposes contains any purpose, if so, those
+                 * items are duplicated.
+                 */
                 if (purposes.length > 0) {
                     context.report(resource, getMessage('iconPurposeDuplicate', context.language, [...(new Set(purposes))].join(' ')), {
                         location: getLocation(`icons[${iconIndex}].purpose`),
--- a/packages/hint-manifest-icons/src/hint.ts
+++ b/packages/hint-manifest-icons/src/hint.ts
@@ -6,7 +6,7 @@ import { imageSize as getImageData } from 'image-size';
 import imageType from 'image-type';
 import { IHint, NetworkData, HintContext } from 'hint';
 import { JSONLocationFunction } from '@hint/utils-json';
-import { ManifestEvents, ManifestParsed, ManifestImageResource, ManifestInvalidSchema } from '@hint/parser-manifest';
+import { ManifestEvents, ManifestParsed, ManifestImageResource } from '@hint/parser-manifest';
 import { ProblemLocation, Severity } from '@hint/utils-types';
 import { debug as d } from '@hint/utils-debug';
 import { determineMediaTypeBasedOnFileExtension } from '@hint/utils/dist/src/content-type';
@@ -17,12 +17,8 @@ import { extname } from 'path';
 
 const debug: debug.IDebugger = d(__filename);
 
-/*
- * The icon errors will look like:
- *     - .icons[0].purpose
- *     - .icons[1].purpose
- */
-const iconsErrorRegex: RegExp = /\.icons\[\d\]\.(purpose)/i;
+// Valid values for icon purpose property.
+const validPurposes = ['any', 'maskable', 'badge', 'monochrome'];
 
 /*
  * ------------------------------------------------------------------------------
@@ -274,28 +270,39 @@ export default class ManifestIconHint implements IHint {
             return validSizes;
         };
 
-        const validateIconsPurpose = (icons: ManifestImageResource[], resource: string, location: ProblemLocation | null) => {
-            for (const icon of icons) {
+        const validateIconsPurpose = (icons: ManifestImageResource[], resource: string, getLocation: JSONLocationFunction) => {
+            for (let iconIndex = 0; iconIndex < icons.length; iconIndex++) {
+                const icon = icons[iconIndex];
+
                 if (!icon.purpose) {
                     continue;
                 }
 
-                const purposes = icon.purpose.split(' ').filter((i) => {
-                    return i;
+                const purposes = icon.purpose.split(' ').filter((purpose) => {
+                    return purpose;
                 });
                 const purposesSet = new Set(purposes);
 
                 for (const purpose of purposesSet) {
-                    const index = purposes.indexOf(purpose);
+                    if (!validPurposes.includes(purpose)) {
+                        context.report(resource, getMessage('iconPurposeInvalid', context.language, purpose), {
+                            location: getLocation(`icons[${iconIndex}].purpose`),
+                            severity: Severity.warning
+                        });
+                    }
 
-                    if (index >= 0) {
-                        purposes.splice(index, 1);
+                    const purposeIndex = purposes.indexOf(purpose);
+
+                    if (purposeIndex >= 0) {
+                        purposes.splice(purposeIndex, 1);
                     }
                 }
 
                 if (purposes.length > 0) {
-                    // Is warning the right severity?
-                    context.report(resource, getMessage('iconPurposeDuplicate', context.language, [...(new Set(purposes))].join(' ')), { location, severity: Severity.warning });
+                    context.report(resource, getMessage('iconPurposeDuplicate', context.language, [...(new Set(purposes))].join(' ')), {
+                        location: getLocation(`icons[${iconIndex}].purpose`),
+                        severity: Severity.hint
+                    });
                 }
             }
         };
@@ -318,7 +325,7 @@ export default class ManifestIconHint implements IHint {
                         hasRequiredSizes(validSizes, resource, iconlocation);
                     }
 
-                    validateIconsPurpose(icons, resource, iconlocation);
+                    validateIconsPurpose(icons, resource, getLocation);
                 } else {
                     // Empty array in `icons` property (otherwise the schema will not validate)
                     const message = getMessage('validIconsNotFound', context.language);
@@ -337,21 +344,6 @@ export default class ManifestIconHint implements IHint {
             }
         };
 
-        const invalidManifest = ({ errors, resource }: ManifestInvalidSchema) => {
-            for (const error of errors) {
-                const isIconError = error.dataPath.match(iconsErrorRegex);
-
-                if (isIconError) {
-                    // Is error the right severity?
-                    context.report(resource, getMessage('iconPurposeInvalid', context.language, isIconError[1]), {
-                        location: error.location,
-                        severity: Severity.error
-                    });
-                }
-            }
-        };
-
         context.on('parse::end::manifest', validate);
-        context.on('parse::error::manifest::schema', invalidManifest);
     }
 }
--- a/packages/hint-manifest-icons/src/hint.ts
+++ b/packages/hint-manifest-icons/src/hint.ts
@@ -294,7 +294,7 @@ export default class ManifestIconHint implements IHint {
                  *      that will keep in `purposes` the duplicated ones.
                  *
                  * We are using the `purposesSet` to iterate because it won't have any
-                 * duplicate value and it will make duplicate detection easier.
+                 * duplicated value and it will make duplicate detection easier.
                  */
                 for (const purpose of purposesSet) {
                     // Check if the purpose is valid.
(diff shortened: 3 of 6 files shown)


packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +269,32 @@ export default class ManifestIconHint implements IHint {
...
+                for (const purpose of purposesSet) {
+                    const index = purposes.indexOf(purpose);
+
+                    if (index >= 0) {
+                        purposes.splice(index, 1);
+                    }
+                }
+
+                if (purposes.length > 0) {
+                    // Is warning the right severity?
```
johnemau, line 292: @molant

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +269,32 @@ export default class ManifestIconHint implements IHint {
...
 
+        const validateIconsPurpose = (icons: ManifestImageResource[], resource: string, location: ProblemLocation | null) => {
+            for (let index = 0; index < icons.length; index++) {
+                const icon = icons[index];
+
+                if (!icon.purpose) {
+                    continue;
+                }
+
+                const purposes = compact(icon.purpose.split(' '));
```
johnemau, line 280: IMO it would be better to avoid pulling in `lodash`

```js
const purposes = icon.purpose.split(' ').filter(i => i);
```

packages/hint-manifest-icons/package.json
```diff
@@ -14,7 +14,8 @@
     "@hint/utils-i18n": "^1.0.3",
     "@hint/utils-types": "^1.0.1",
     "image-size": "^0.8.3",
-    "image-type": "^4.1.0"
+    "image-type": "^4.1.0",
+    "lodash": "^4.17.15"
```
johnemau, line 18: IMO we should avoid adding dependencies when possible. I think we can get away without lodash for this change.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -302,6 +332,24 @@ export default class ManifestIconHint implements IHint {
...
 
+        const invalidManifest = ({ errors, resource }: ManifestInvalidSchema) => {
+            for (const error of errors) {
+                const isIconError = error.dataPath.match(iconsErrorRegex);
+
+                if (isIconError) {
+                    // Is error the right severity?
+                    const message = getMessage('iconPurposeInvalid', context.language, isIconError[1]);
+
+                    console.log(message);
```
johnemau, line 343: Is this console.log left over from debugging?

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -6,16 +6,18 @@ import { imageSize as getImageData } from 'image-size';
...
 import { debug as d } from '@hint/utils-debug';
 import { determineMediaTypeBasedOnFileExtension } from '@hint/utils/dist/src/content-type';
+import compact = require('lodash/compact');
 
 import meta from './meta';
 import { getMessage } from './i18n.import';
 import { extname } from 'path';
 
 const debug: debug.IDebugger = d(__filename);
+const iconsErrorRegex: RegExp = /\.icons\[\d\]\.(purpose)/i;
```
johnemau, line 20: Please include a comment with an example error message you expect this regex to match.
```js
// example: manifest.icons9.purpose

File: packages/hint-manifest-icons/src/hint.ts [hunk 6]
```diff
@@ -267,6 +269,32 @@ export default class ManifestIconHint implements IHint {
             return validSizes;
         };
 
+        const validateIconsPurpose = (icons: ManifestImageResource[], resource: string, location: ProblemLocation | null) => {
+            for (let index = 0; index < icons.length; index++) {
```
johnemau, line 273: nit: suggest using `for of` to avoid shadowing the below `index` variable.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -302,6 +332,24 @@ export default class ManifestIconHint implements IHint {
             }
         };
 
+        const invalidManifest = ({ errors, resource }: ManifestInvalidSchema) => {
+            for (const error of errors) {
+                const isIconError = error.dataPath.match(iconsErrorRegex);
+
+                if (isIconError) {
+                    // Is error the right severity?
```
johnemau, line 340: @molant

packages/parser-manifest/src/schema.json
```diff
@@ -10,13 +10,7 @@
...
                 "purpose": {
                     "default": "any",
-                    "enum": [
-                        "any",
-                        "any badge",
-                        "badge",
-                        "badge any",
-                        "maskable"
-                    ],
+                    "pattern": "^\\s*(any|maskable|badge)(\\s+(any|maskable|badge)?)*$",
```
johnemau, line 13: https://github.com/webhintio/hint/issues/3791#issuecomment-641935207

We need to add support for `monochrome` as well, I am OK if you think that should be done in a follow-up change.

packages/hint-manifest-icons/src/hint.ts
sarvaje: yes! Good catch

packages/parser-manifest/src/schema.json
sarvaje: I can add it to this PR

packages/hint-manifest-icons/src/hint.ts
sarvaje: > const purposes = icon.purpose.split(' ').filter(i => i);

I like that!

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +269,32 @@ export default class ManifestIconHint implements IHint {
...
+                    const index = purposes.indexOf(purpose);
+
+                    if (index >= 0) {
+                        purposes.splice(index, 1);
+                    }
+                }
+
+                if (purposes.length > 0) {
+                    // Is warning the right severity?
+                    context.report(resource, getMessage('iconPurposeDuplicate', context.language, [...(new Set(purposes))].join(' ')), { location, severity: Severity.warning });
```
molant, line 298: We might even want to lower the severity and mark it as a hint. Having a valid duplicate purpose doesn't break anything.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -302,6 +337,21 @@ export default class ManifestIconHint implements IHint {
             }
         };
 
+        const invalidManifest = ({ errors, resource }: ManifestInvalidSchema) => {
+            for (const error of errors) {
+                const isIconError = error.dataPath.match(iconsErrorRegex);
+
+                if (isIconError) {
+                    // Is error the right severity?
+                    context.report(resource, getMessage('iconPurposeInvalid', context.language, isIconError[1]), {
```
molant, line 346: The [spec says the following](https://w3c.github.io/manifest/#purpose-member):

> If an icon contains multiple purposes, it could be used for any of those purposes. If none of the stated purposes are recognized, the icon is totally ignored. For example, if an icon has purpose "monochrome fizzbuzz", then it could be used as a monochrome icon, as "monochrome" is a valid purpose. However, if an icon just has the purpose "fizzbuzz", then it will be ignored.

That said, I think that we should have an Error if none of the purposes are recognized (the icon is not going to work so that's probably an error) and a warning otherwise.
We could maybe lower the severity one level for each but probably at that moment we will have to look at other icons and what they do, etc. which might complicate things a lot for not a lot of value.

@Malvoz what do you think?

packages/hint-manifest-icons/src/_locales/en/messages.json
```diff
@@ -11,6 +11,14 @@
         "description": "Report message when an icon request return an status code !== 200",
         "message": "Icon could not be fetched (status code: $1)."
     },
+    "iconPurposeDuplicate": {
+        "description": "Report message when an icon has a duplicate value in the property purpose",
+        "message": "Duplicate value(s) '$1' found in icon purpose property."
```
molant, line 16: The convention is to use `$1` at the end so localization is easier.
@johnemau do you have any suggestions?

packages/hint-manifest-icons/src/_locales/en/messages.json
```diff
@@ -11,6 +11,14 @@
...
+    },
+    "iconPurposeInvalid": {
+        "description": "Report message when an icon has an invalid value in the property purpose",
+        "message": "Invalid icon property $1."
```
molant, line 20:
```suggestion
        "message": "Invalid icon property: $1."
```

packages/hint-manifest-icons/src/hint.ts
sarvaje: If having an invalid value "is not a problem" we should modify the schema to support an string, and do all the checks in the hint.

packages/hint-manifest-icons/src/_locales/en/messages.json
sarvaje: what about `Duplicate value(s) found in icon purpose porperty: $1`?

molant: SGTM

packages/hint-manifest-icons/src/hint.ts
molant: That's what the spec says so...

sarvaje: Ok, I will:

1. Remove the regex from the schema.
2. Check if any value is invalid.
3. In case any value is invalid, report a warning.

> We could maybe lower the severity one level for each but probably at that moment we will have to look at other icons and what they do, etc. which might complicate things a lot for not a lot of value.

If you mean the other icons in the list of icons, I think we should report what we found in each icon independently.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +270,43 @@ export default class ManifestIconHint implements IHint {
             return validSizes;
         };
 
+        const validateIconsPurpose = (icons: ManifestImageResource[], resource: string, getLocation: JSONLocationFunction) => {
```
molant, line 273:
```suggestion

        // See https://w3c.github.io/manifest/#purpose-member
        const validateIconsPurpose = (icons: ManifestImageResource[], resource: string, getLocation: JSONLocationFunction) => {
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +270,43 @@ export default class ManifestIconHint implements IHint {
...
+            for (let iconIndex = 0; iconIndex < icons.length; iconIndex++) {
+                const icon = icons[iconIndex];
+
+                if (!icon.purpose) {
+                    continue;
+                }
+
+                const purposes = icon.purpose.split(' ').filter((purpose) => {
+                    return purpose;
+                });
```
molant, line 283: Is this `filter` to remove any possible trailing space? For example, if the purpose is `maskable badge `?
If so I think it will be more readable if you just do this:

```suggestion
                const purposes = icon.purpose.trim().split(' ');
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +270,43 @@ export default class ManifestIconHint implements IHint {
...
+                    }
+
+                    const purposeIndex = purposes.indexOf(purpose);
+
+                    if (purposeIndex >= 0) {
+                        purposes.splice(purposeIndex, 1);
+                    }
+                }
+
+                if (purposes.length > 0) {
```
molant, line 301: It took me a bit to figure out what you were doing in here and why `purposes.length` had to be 0 by line 301.

What do you think of instead doing this:

```suggestion                    
                }

                if (purposes.length !== purposesSet.size) {
```

packages/hint-manifest-icons/tests/tests.ts
```diff
@@ -214,7 +214,123 @@ const tests: HintTest[] = [
...
+                    }`
+            }
+        }
+    },
+    {
+        name: 'Duplicated purpose found',
+        reports: [
+            {
+                message: `Duplicate value(s) found in icon purpose porperty: any.`,
+                position: {
```
molant, line 227: Maybe we could use the position matching so this is easier to maintain in the future?

packages/hint-manifest-icons/src/hint.ts
sarvaje: It is also to remove possible extra spaces in the middle of the string: `any       badge`

packages/hint-manifest-icons/src/hint.ts
sarvaje: That was my first solution, but you can't know the duplicated items with that code. Maybe I can add a few comments?

packages/hint-manifest-icons/tests/tests.ts
sarvaje: That would be awesome, but does it work with this match `icons[0].purpose` because I think I tried and it didn't work. (I am going take another look)

sarvaje: I think I can use `purpose` for the match, but it will only match the first occurrence of `purpose` and there are 3 icons in the tests.

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -17,6 +17,9 @@ import { extname } from 'path';
 
 const debug: debug.IDebugger = d(__filename);
 
+// Valid values for icon purpose property.
+const validPurposes = ['any', 'maskable', 'badge', 'monochrome'];
```
amazingcaio, line 21: as w3c/manifest#833 has been merged, the spec does not mention `badge` any more, so i think it should be removed from `validPurposes`, though maybe there should be a different issue dealing with this? (especially because you might want webhint to have a specific behaviour, like warning `badge` was replaced?)

molant: @sarvaje @hxlnt this hint is still not [published "officially"](https://github.com/webhintio/hint/blob/145ab5a11f7077a63b28a954f9a45bc5efa20e9b/packages/hint-manifest-icons/package.json#L55) although it seems we are using it seems to be active in different places. IIRC the missing piece was to update the documentation. Ideally we should do the `badge/monochrome` before we "release" it.

https://github.com/w3c/manifest/pull/833#issuecomment-645756918

> Confirmed that Chrome does not use purpose: badge anywhere, including on the server-side infrastructure. From our perspective, we are good to rename it and change its semantics. LGTM.

@amazingcaio, thanks for catching this change!

packages/hint-manifest-icons/tests/tests.ts
molant: I thought it was a bit more flexible. Not worth doing it then.

packages/hint-manifest-icons/src/hint.ts
molant: The following should work but at this time I'm not sure what's easier to read

```suggestion
                const purposes = icon.purpose.trim().replace(/\s\s*/g,' ').split(' ').;
```

@johnemau what do you think?

sarvaje: for readability, I think the regex is easy and more clear what we are doing. Performance wise, I'm not sure what is better.

molant: Let's optimize for readability then, I don't think we can really measure the impact of one or the other 😊

packages/hint-manifest-icons/src/hint.ts
molant: I think it might be we are doing things in 2 different places.

What about filtering/finding duplicated in the same place?

```ts
if (purposes.length !== purposesSet.size) {
  // Check if it's duplicated here and report
}
```

That way if we want to refactor that code into another function we can. We might have to do an extra loop but it should be negligible and make it better to read/maintain.

packages/hint-manifest-icons/src/hint.ts
sarvaje: Thanks @amazingcaio!

amazingcaio: no problem. also, there were a couple of places where you misspelt `property` as `porperty`. you may want to fix that :)

sarvaje: Thanks!! It should be fixed now.

packages/hint-manifest-icons/src/_locales/en/messages.json
```diff
@@ -11,6 +11,14 @@
         "description": "Report message when an icon request return an status code !== 200",
         "message": "Icon could not be fetched (status code: $1)."
     },
+    "iconPurposeDuplicate": {
+        "description": "Report message when an icon has a duplicate value in the property purpose",
+        "message": "Duplicate value(s) found in icon purpose porperty: $1."
```
amazingcaio, line 16: missed misspelling
```suggestion
        "message": "Duplicate value(s) found in icon purpose property: $1."
```

packages/hint-manifest-icons/tests/tests.ts
```diff
@@ -214,7 +214,123 @@ const tests: HintTest[] = [
...
                                 "type": "image/png"
                             }
-                            ]
+                        ]
+                    }`
+            }
+        }
+    },
+    {
+        name: 'Duplicated purpose found',
```
amazingcaio, line 223: sorry for being so pedantic
```suggestion
        name: 'Duplicate purpose found',
```

packages/hint-manifest-icons/src/hint.ts
sarvaje: done

packages/hint-manifest-icons/src/hint.ts
sarvaje: done

packages/hint-manifest-icons/tests/tests.ts
sarvaje: 🤦‍♂️I forgot to save the change before pushing... 
image

packages/hint-manifest-icons/src/_locales/en/messages.json
sarvaje: 🤦‍♂️ I forgot to save my changes before pushing...
image

packages/hint-manifest-icons/tests/tests.ts
molant: English isn't @sarvaje's nor mine first language so any grammar, spelling, or style corrections are more than welcome 😊

Thanks a lot!

sarvaje: Thanks for the correction!!, as @molant says, English isn't our first language :)

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +270,78 @@ export default class ManifestIconHint implements IHint {
...
+            for (const purpose of purposes) {
+                // Check if the purpose is valid.
+                if (!validPurposes.includes(purpose)) {
+                    context.report(resource, getMessage('iconPurposeInvalid', context.language, purpose), {
+                        location: getLocation(`icons[${iconIndex}].purpose`),
+                        severity: Severity.warning
+                    });
+                }
+            }
+            console.log(iconIndex);
```
molant, line 283:
```suggestion
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +270,78 @@ export default class ManifestIconHint implements IHint {
...
+            if (purposesCopy.length > 0) {
+                context.report(resource, getMessage('iconPurposeDuplicate', context.language, [...(new Set(purposesCopy))].join(' ')), {
+                    location: getLocation(`icons[${iconIndex}].purpose`),
+                    severity: Severity.hint
+                });
+            }
+
+            console.log(getMessage('iconPurposeDuplicate', context.language, [...(new Set(purposesCopy))].join(' ')));
+            console.log(getLocation(`icons[${iconIndex}].purpose`));
+            console.log(iconIndex);
```
molant, line 320: Are this leftovers or did you intend to use `debug` instead?

sarvaje: Sorry, I forgot to remove the `console.log`s

sarvaje: done!

packages/hint-manifest-icons/tests/tests.ts
amazingcaio: english isn't my first language either lol. the correction was more about consistency, as `duplicate` was used in other cases, not `duplicated`. (plus `duplicate` does sound better in my ears, but english is my second language so i'm not that reliable.)

molant:
@sarvaje is this ready for another review?

sarvaje:
> @sarvaje is this ready for another review?

@molant Yes, everything should be addressed

antross:
@molant any further feedback on this or is it good to merge?

molant:
Don't think I'll have the time to check until next week. If it's OK on your side then go ahead 😊

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +270,73 @@ export default class ManifestIconHint implements IHint {
...
+                    continue;
+                }
+
+                /*
+                 * Remove unnecessary empty spaces before split the array.
+                 */
+                const purposes = icon.purpose
+                    .trim()
+                    .replace(/\s\s*/g, ' ')
+                    .split(' ');
```
antross, line 333: We can just pass the regex to `split` directly.
```suggestion
                    .split(/\s\s*/g);
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -267,6 +270,73 @@ export default class ManifestIconHint implements IHint {
...
+                    purposesCopy.splice(purposeIndex, 1);
+                }
+            }
+
+            /*
+             * Check if purposesCopy contains any purpose, if so, those
+             * items are duplicated.
+             */
+            if (purposesCopy.length > 0) {
+                context.report(resource, getMessage('iconPurposeDuplicate', context.language, [...(new Set(purposesCopy))].join(' ')), {
```
antross, line 311: Probably should comma-separate these:
```suggestion
                context.report(resource, getMessage('iconPurposeDuplicate', context.language, [...(new Set(purposesCopy))].join(', ')), {
```

packages/hint-manifest-icons/src/hint.ts
```diff
@@ -277,13 +347,15 @@ export default class ManifestIconHint implements IHint {
                 if (icons.length > 0) {
 
                     debug(`Validating if manifest-icon file exists`);
-                    const validSizes = await validateIcons(icons, hostnameWithProtocol, resource, getLocation);
+                    const validSizes = await validateIconsSize(icons, hostnameWithProtocol, resource, getLocation);
+                    const iconlocation = getLocation('icons');
 
                     if (validSizes.length > 0) {
-                        const iconlocation = getLocation('icons');
```
antross, line 283: Seems like this isn't used outside this block, so can probably be kept here rather than moved out.

packages/hint-manifest-icons/tests/tests.ts
antross: The position matching should work here, we just need to include enough of the matching line to be unique within the file.

PR #3860, opened 2020-07-10
Fix: Improve perf by using pageDOM.querySelector in hint-axe

johnemau:
## Pull request checklist

Make sure you:

This change improves the performance of hint-axe by using `pageDOM.querySelector` instead of `HintContext.querySelectorAll`. The latter will stop DOM traversal after the first match: best case O(1), the former always does complete DOM traversal: best case O(N).

The typing changed to include `undefined` however this is was the case before because the zero index of an empty array is `undefined`.

### DIFF
--- a/packages/hint-axe/src/util/axe.ts
+++ b/packages/hint-axe/src/util/axe.ts
@@ -33,11 +33,10 @@ type RegistrationMap = Map<EngineKey, Map<string, Registration[]>>;
  */
 const registrationMap: RegistrationMap = new Map();
 
-const getElement = (context: HintContext, node: AxeNodeResult): HTMLElement => {
+const getElement = (context: HintContext, node: AxeNodeResult): HTMLElement | undefined | null => {
     const selector = node.target[0];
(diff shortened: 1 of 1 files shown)


johnemau:
> LGTM. Would be good to have some metrics of the before/after.

I did not notice much of a difference in the unit tests (_jsdom_) while the improvement to the Edge DevTools integration (_utils-dom_) was noticeable.

Here are some anecdotal numbers from my machine:

#### jsdom via color unit test
|before|after|
|-------|-----|
|6.0492|5.5854|
|5.9948|5.8891|

#### utils-dom via Edge DevTools (webhint.io)
|before|after|
|-------|-----|
|1.23499|0.71500|
|1.01999|0.80000|
|1.18499|0.72499|

johnemau:
I pushed an update for the return typing.

This was needed because the `pageDOM?.querySelector` signature may also return `null`, failing TypeScript compilation locally. This failed after a testing a rebase onto the main branch. I am not sure why it passed before but did not investigate further.

PR #3861, opened 2020-07-10
Fix: Improve utils-dom querySelector perf by using cssSelect.selectOne

johnemau:
## Pull request checklist

Make sure you:

This change improves `querySelector` best case performance from O(N) to O(1).

Previously `querySelector` called `querySelectorAll` which had a best case of O(N) traversing all nodes in the DOM (passed in via `_document.children`).

The new implementation re-uses the css-select query cached (shared with `querySelectorAll`) and then leverages `cssSelect.selectOne` which internally leverages [domUtil's `findOne`][1] which has the improved performance (over `findAll`).

[1]: https://github.com/fb55/domutils/blob/d7d349c25f78a50cc3f29d9e83f86cbbcf93b2a9/src/querying.ts#L75

### DIFF
--- a/packages/utils-dom/src/htmldocument.ts
+++ b/packages/utils-dom/src/htmldocument.ts
@@ -165,8 +165,17 @@ export class HTMLDocument extends Node {
     /**
      * https://developer.mozilla.org/docs/Web/API/Document/querySelector
      */
-    public querySelector(selector: string): HTMLElement {
-        return this.querySelectorAll(selector)[0];
+    public querySelector(selector: string): HTMLElement | null {
+        if (!CACHED_CSS_SELECTORS.has(selector)) {
+            CACHED_CSS_SELECTORS.set(selector, cssSelect.compile(selector));
+        }
+
+        const data: any | undefined = cssSelect.selectOne(
+            CACHED_CSS_SELECTORS.get(selector) as cssSelect.CompiledQuery,
+            this._document.children
+        );
+
+        return data ? this.getNodeFromData(data) as HTMLElement : null;
     }
 
     /**
--- a/packages/utils-dom/src/htmldocument.ts
+++ b/packages/utils-dom/src/htmldocument.ts
@@ -170,7 +170,7 @@ export class HTMLDocument extends Node {
             CACHED_CSS_SELECTORS.set(selector, cssSelect.compile(selector));
         }
 
-        const data: any | undefined = cssSelect.selectOne(
+        const data = cssSelect.selectOne(
             CACHED_CSS_SELECTORS.get(selector) as cssSelect.CompiledQuery,
             this._document.children
         );
--- a/packages/utils-dom/tests/htmldocument.ts
+++ b/packages/utils-dom/tests/htmldocument.ts
@@ -11,3 +11,21 @@ test('title', (t) => {
     t.is(doc2.title, '');
     t.is(doc3.title, 'test');
 });
+
+test('querySelectorAll', (t) => {
+    const doc = createHTMLDocument('<div id="d1">div1</div><div id="d2">div2</div>', 'http://localhost/');
+
+    t.is(doc.querySelectorAll('div').length, 2);
+    t.is(doc.querySelectorAll('#d1').length, 1);
+    t.is(doc.querySelectorAll('#d2').length, 1);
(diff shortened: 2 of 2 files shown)


packages/utils-dom/src/htmldocument.ts
```diff
@@ -165,8 +165,17 @@ export class HTMLDocument extends Node {
...
      * https://developer.mozilla.org/docs/Web/API/Document/querySelector
      */
-    public querySelector(selector: string): HTMLElement {
-        return this.querySelectorAll(selector)[0];
+    public querySelector(selector: string): HTMLElement | null {
+        if (!CACHED_CSS_SELECTORS.has(selector)) {
+            CACHED_CSS_SELECTORS.set(selector, cssSelect.compile(selector));
+        }
+
+        const data: any | undefined = cssSelect.selectOne(
```
molant, line 173: Can we use `NodeData` in here or is that causing any types issue? If so, maybe we should cast it already here instead of pass `any` around?

packages/utils-dom/src/htmldocument.ts
```diff
@@ -165,8 +165,17 @@ export class HTMLDocument extends Node {
...
+            CACHED_CSS_SELECTORS.get(selector) as cssSelect.CompiledQuery,
```
molant, line 174: Can't the cache cause troubles if we switch to analyze continuously?

johnemau: Good question,

It will not cause any issues because "compiled selectors" are unique per css selector and have nothing to do with the DOM, so '#my-id div.my-class' will always _compile_ to the same function.

For more details see: https://github.com/fb55/css-select#how-does-it-work

Additionally in the current Edge DevTools implementation, we get a fresh webhint instance per DOM snapshot when running continuously.

packages/utils-dom/src/htmldocument.ts
johnemau: Good point, let me give it a try.

johnemau: I resolved this by removing the explicit. selectOne returns a complex union type that is filtered down by the arguments, in this case it returns `ChildData` which is compatible with `NodeData`.

By setting the typing to be implicit we let the TyepScript compiler do the work for us :)

johnemau:
> Do you have any before/after metrics for this change?
> I'm curious to see how it improves on a large website 😊

#### Performance Results (ms) (random sampling from multiple runs):
```
querySelector: 0.3227
quarySelectorAll: 1.844701
```

The `querySelectorAll` performance result is equivalent to _before_ this change (`return this.querySelectorAll(selector)[0]`).

#### Tests
```ts
test.serial.beforeEach(t => {
    const NEST_LEVEL = 1000;
    let docString = '<div></div>';

    for (let i = 0; i < NEST_LEVEL; i++) {
        docString = `<div>${docString}</div>`;
    }

    docString = `<div id="findme"></div>${docString}`;

    const doc = createHTMLDocument(docString, 'http://localhost/');

    (t.context as any).doc = doc;
    (t.context as any).done = new Promise((r) => {
        (t.context as any).resolve = r;
    });

   // https://nodejs.org/api/perf_hooks.html#perf_hooks_performance_measurement_apis
    const obv = new PerformanceObserver((items) => {
        const [{name, duration}] = items.getEntries();
        console.log(name, duration);
        performance.clearMarks();
        (t.context as any).resolve();
    });

    obv.observe({entryTypes: ['measure']});
});

test.serial.only('querySelector-perf-tests-1', async (t) => {
    performance.mark('start_qs');
    (t.context as any).doc.querySelector('#findme');
    performance.mark('stop_qs');
    performance.measure('querySelector("#findme")', 'start_qs', 'stop_qs');
    await (t.context as any).done;
    t.pass();
});

test.serial.only('querySelector-perf-tests-2', async (t) => {
    performance.mark('start_qsa');
    (t.context as any).doc.querySelectorAll('#findme');
    performance.mark('stop_qsa');
    performance.measure('querySelectorAll("#findme")', 'start_qsa', 'stop_qsa');
    await (t.context as any).done;
    t.pass();
});
```

#### How-to Run Tests
```CMD
yarn clean && yarn build:assets && yarn build:ts && npx ava -m querySelector-perf-tests-1
yarn clean && yarn build:assets && yarn build:ts && npx ava -m querySelector-perf-tests-2
```

PR #3862, opened 2020-07-10
Chore: Prefer querySelector over querySelectorAll

johnemau:
## Pull request checklist

Make sure you:

This change prefers `querySelector` over `querySelectorAll` in utils-dom.

This is semantically more correct (better expresses the intent of the code) and leverages the upcoming performance improvement in #3861

### DIFF
--- a/packages/utils-dom/src/htmldocument.ts
+++ b/packages/utils-dom/src/htmldocument.ts
@@ -48,7 +48,7 @@ export class HTMLDocument extends Node {
     }
 
     private getBaseUrl(finalHref: string): string {
-        const baseElement = this.querySelectorAll('base[href]')[0];
+        const baseElement = this.querySelector('base[href]');
         const baseHref = baseElement ? baseElement.getAttribute('href') : null;
 
         if (!baseHref) {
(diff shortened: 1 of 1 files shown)


packages/utils-dom/src/htmldocument.ts
```diff
@@ -77,7 +77,7 @@ export class HTMLDocument extends Node {
      * https://developer.mozilla.org/docs/Web/API/Document/body
      */
     public get body(): HTMLElement {
-        return this.querySelectorAll('body')[0];
+        return this.querySelector('body') as HTMLElement;
```
molant, line 80: Do we need `as HTMLElement` or can TypeScript infer it?

johnemau: Here we break type safety because `querySelector` might return `null` but the body getter must return a `HTMLElement`.

This is the best we can do, Before TypeScript did an implicit cast from `undefined` if the array was empty because it does not support `undefined` as a default return type for arrays: https://github.com/microsoft/TypeScript/issues/13778

PR #3903, opened 2020-07-20
Include links and text in problem reports for hint-compat-api

### LINKED ISSUE #3769: [Feature] Include links and text in problem reports
Related to #3045, we should generally include explicit links along with problem reports (along with suggested link text), whether we're linking to MDN, webhint.io, or somewhere else.

sarvaje:
## Pull request checklist

Make sure you:

This PR will allows to Webhint to report an specific url and text for `hint-compat-api`, so we can point to the user to the MDN website.

Fix #3769

### DIFF
--- a/packages/hint-compat-api/src/_locales/en/messages.json
+++ b/packages/hint-compat-api/src/_locales/en/messages.json
@@ -26,5 +26,9 @@
     "learnMoreCSS": {
         "description": "Text to show for the documentation link",
         "message": "Learn more about the CSS property"
+    },
+    "learnMoreHTML": {
+        "description": "Text to show for the documentation link",
+        "message": "Learn more about the HTML property"
     }
 }
--- a/packages/hint-compat-api/src/_locales/en/messages.json
+++ b/packages/hint-compat-api/src/_locales/en/messages.json
@@ -22,5 +22,9 @@
     "html_name": {
         "description": "html hint metadata name",
         "message": "Supported HTML features"
+    },
+    "learnMoreCSS": {
+        "description": "Text to show for the documentation link",
+        "message": "Learn more about the CSS property"
     }
 }
--- a/packages/hint-compat-api/src/_locales/en/messages.json
+++ b/packages/hint-compat-api/src/_locales/en/messages.json
@@ -29,6 +29,6 @@
     },
     "learnMoreHTML": {
         "description": "Text to show for the documentation link",
-        "message": "Learn more about the HTML property"
+        "message": "Learn more about this HTML feature on MDN"
     }
 }
--- a/packages/hint-compat-api/src/_locales/en/messages.json
+++ b/packages/hint-compat-api/src/_locales/en/messages.json
@@ -25,7 +25,7 @@
     },
     "learnMoreCSS": {
         "description": "Text to show for the documentation link",
-        "message": "Learn more about the CSS property"
+        "message": "Learn more about this CSS feature on MDN"
     },
     "learnMoreHTML": {
         "description": "Text to show for the documentation link",
--- a/packages/hint-compat-api/src/css.ts
+++ b/packages/hint-compat-api/src/css.ts
@@ -5,9 +5,9 @@
 import intersection = require('lodash/intersection');
 import { vendor, AtRule, Rule, Declaration, ChildNode, ContainerBase } from 'postcss';
 
-import { HintContext, HintDocumentation } from 'hint/dist/src/lib/hint-context';
+import { HintContext } from 'hint/dist/src/lib/hint-context';
 import { IHint } from 'hint/dist/src/lib/types';
-import { Severity } from '@hint/utils-types';
+import { ProblemDocumentation, Severity } from '@hint/utils-types';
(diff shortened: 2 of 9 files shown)


sarvaje:
@molant @antross @johnemau this is ready to review

packages/hint-compat-api/src/_locales/en/messages.json
```diff
@@ -22,5 +22,13 @@
     "html_name": {
         "description": "html hint metadata name",
         "message": "Supported HTML features"
+    },
+    "learnMoreCSS": {
+        "description": "Text to show for the documentation link",
+        "message": "Learn more about the CSS property"
```
antross, line 28: May need to tweak this text a bit. These could be properties, values, selectors, etc. Plus we should probably call out MDN somewhere in the link text. How about this?
```suggestion
        "message": "Learn more about this CSS feature on MDN"
```

packages/hint-compat-api/src/_locales/en/messages.json
```diff
@@ -22,5 +22,13 @@
...
+    },
+    "learnMoreHTML": {
+        "description": "Text to show for the documentation link",
+        "message": "Learn more about the HTML property"
```
antross, line 32: Same thing here
```suggestion
        "message": "Learn more about this HTML feature on MDN"
```

packages/hint-compat-api/src/css.ts
```diff
@@ -276,12 +277,22 @@ export default class CSSCompatHint implements IHint {
...
                 const severity = alternatives.length ? Severity.error : Severity.warning;
 
+                let documentation: ProblemDocumentation[] | undefined;
+
+                if (unsupported.mdnUrl) {
+                    documentation = [{
+                        link: unsupported.mdnUrl,
+                        text: getMessage('learnMoreCSS', context.language)
+                    }];
+                }
```
antross, line 287: Minor nit to reduce explicit types and allow this to be `const`. Feel free to ignore if you don't like the format.

```suggestion
                const documentation = unsupported.mdnUrl && [{
                    link: unsupported.mdnUrl,
                    text: getMessage('learnMoreCSS', context.language)
                }];
```

packages/hint-compat-api/src/html.ts
```diff
@@ -86,11 +86,20 @@ export default class HTMLCompatHint implements IHint {
 
             const report = ({ feature, unsupported }: ReportData) => {
                 const message = getMessage('featureNotSupported', context.language, [feature, joinBrowsers(unsupported)]);
+                let documentation: ProblemDocumentation[] | undefined;
+
+                if (unsupported.mdnUrl) {
```
antross, line 91: Nit: you can use the same inline initialization to make this `const` here as I suggested for the CSS case.

packages/utils-compat-data/src/browsers.ts
```diff
@@ -247,5 +248,15 @@ export const getUnsupportedBrowsers = (feature: Identifier | undefined, prefix:
...
     }
 
-    return unsupported.length ? { browsers: unsupported, details } : null;
+    if (unsupported.length === 0) {
+        return null;
+    }
+
+    let mdnUrl = feature.__compat.mdn_url;
+
+    if (!mdnUrl && parent && parent.__compat) {
```
antross, line 257: Not sure if there's a concrete example of this in the current data, but couldn't the documentation be more than one level up? 

If so, it may be more future-proof to pass a list of ancestors instead of a single parent and loop backwards through those hunting for an `mdn_url`. 

That said, if we can't find any current examples, I don't think it's worth significant additional work to support this now.

packages/utils-compat-data/tests/browsers.ts
```diff
@@ -219,3 +224,8 @@ test('Includes accurate details', (t) => {
         versionAdded: '15'
     }, 'Before -webkit- support');
 });
+
+test('Returns the mdn url', (t) => {
+    t.is(getUnsupportedBrowsers(keyframes, '', ['opera 12'], 'keyframes')!.mdnUrl, 'https://developer.mozilla.org/docs/Web/CSS/@keyframes', 'Right mdn url before first unprefixed support');
+    t.is(getUnsupportedBrowsers(maxContent, '', ['firefox 65'], 'max-content', width)!.mdnUrl, 'https://developer.mozilla.org/docs/Web/CSS/width');
```
antross, line 230: Nit: I think we should start using optional chaining for these situations now that it's available. Then if we're ever wrong we'll get `undefined` instead of an exception. Not as big of a deal in tests, but definitely in our other code. What do you think?
```suggestion
    t.is(getUnsupportedBrowsers(maxContent, '', ['firefox 65'], 'max-content', width)?.mdnUrl, 'https://developer.mozilla.org/docs/Web/CSS/width');
```

packages/hint-compat-api/src/css.ts
sarvaje: Looks good :)

packages/utils-compat-data/src/browsers.ts
sarvaje: As we talk before, in the current data, I didn't find any more than one level up

PR #3904, opened 2020-07-21
Chore: Improve utils-dom perf with matches() refactor

johnemau:
## Pull request checklist

Make sure you:

This change improves the performance of `utils-dom` by refactoring `Element.matches`.

#### webhint.io
| before (ms) | after (ms) |
| ------------  | ---------- |
|    0.7199      |   0.7000   |

#### cnn.com
| before (ms) | after (ms) |
| ------------  | ---------- |
|    +60000    |  16.4900  |

_Previously cnn.com could take minutes to complete._

The problem was matches relied on querySelectorAll which will always check every dom node O(N). Now we compile the selector  test function from `css-select` directly and invoke it on the current element O(1). This performance is not bottlenecked by `compile`.

This change also pulls `CACHED_CSS_SELECTORS` into a shared module used by `htmldocument.ts` and `htmlelement.ts`,

Close #3904

### DIFF
--- a/packages/utils-dom/src/htmlelement.ts
+++ b/packages/utils-dom/src/htmlelement.ts
@@ -1,4 +1,3 @@
-import { compile } from 'css-select';
 import * as parse5 from 'parse5';
 import * as htmlparser2Adapter from 'parse5-htmlparser2-tree-adapter';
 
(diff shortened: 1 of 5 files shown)


packages/utils-dom/src/htmlelement.ts
```diff
@@ -233,7 +235,11 @@ export class HTMLElement extends Node {
      * https://developer.mozilla.org/en-US/docs/Web/API/Element/matches
      */
     public matches(selector: string): boolean {
-        return this.ownerDocument.querySelectorAll(selector).includes(this);
+        if (!CACHED_CSS_SELECTORS.has(selector)) {
+            CACHED_CSS_SELECTORS.set(selector, compile(selector));
```
antross, line 239: Can we move the compilation step into the refactored module as well? Then we can just expose a single `getCompiledSelector` method and not have to deal with multiple consumers messing directly with the `CACHED_CSS_SELECTORS` collection.

johnemau: Thanks, done!

Issue #3900, opened 2020-07-20
[BUG] Issue running user guide sample via CLI or project dependency.


johnson-jesse:
https://webhint.io/docs/user-guide/index.html

When running the CLI command as instructed `npx hint https://example.com` the command fails as follows:

```
% npx hint https://example.com
Using the built-in configuration.
Visit https://webhint.io/docs/user-guide/ to learn how to create your own configuration.
✖ 
Error: No installation found for: ""
    at Object.exports.getInstallationPath (/usr/local/lib/node_modules/hint/node_modules/@hint/connector-puppeteer/dist/src/lib/chromium-finder.js:246:15)
    at PuppeteerConnector.toPuppeteerOptions (/usr/local/lib/node_modules/hint/node_modules/@hint/connector-puppeteer/dist/src/connector.js:55:35)
    at new PuppeteerConnector (/usr/local/lib/node_modules/hint/node_modules/@hint/connector-puppeteer/dist/src/connector.js:37:30)
    at new Engine (/usr/local/lib/node_modules/hint/dist/src/lib/engine.js:33:26)
    at Analyzer.analyze (/usr/local/lib/node_modules/hint/dist/src/lib/analyzer.js:140:27)
    at exports.default (/usr/local/lib/node_modules/hint/dist/src/lib/cli/analyze.js:316:23)
    at async /usr/local/lib/node_modules/hint/dist/src/lib/cli/actions.js:7:20
    at async Object.exports.execute (/usr/local/lib/node_modules/hint/dist/src/lib/cli.js:37:23)
    at async run (/usr/local/lib/node_modules/hint/dist/src/bin/hint.js:44:24)
```

Expected the command to produce some useful feedback on `example.com`. Instead, an error was shown.

Installed version:
```
% node --version
v12.18.0
% npx --version
6.14.4
```

johnson-jesse:
The error message is tracked back to this block of code:
```
const browsers = [Browser.Chrome, Browser.Chromium, Browser.Edge];
    let browserFound = '';
    while (browsers.length > 0 && !browserFound) {
        const br = browsers.shift();
        try {
            browserFound = findBrowserPath(br);
        }
        catch (e) {
        }
    }
    if (!browserFound) {
        throw new Error(ERRORS.NoInstallationFound `${browserFound}`);
    }
```

Is there additional configuration needed for this browser path? I'm relying on webhint defaults since I didn't provide a .hintrc. Do I need to download some browser driver?

[...]

PR #3907, opened 2020-07-22
Fix: Grep AWK path accounting for suffix

johnson-jesse:
## Pull request checklist

Make sure you:

Ref https://github.com/webhintio/hint/issues/3900 This affected Mac computers. No paths are found even though the correct browsers are installed.

The fix:
* For grep, I removed the $. As seen below, the path doesn't end in .app. So this entry will never get included if we use $.
* For awk, I updated the print to provide us only index 2 and 3

from this -> `path:     /Applications/Google Chrome.app (0x2134)`
to this     ->`/Applications/Google Chrome.app`

```
const lines = execSync(
        `${LSREGISTER} -dump` +
        ` | grep -i '\.app'` + // eslint-disable-line
        ` | awk '{print $2,$3}'`)
        .toString()
        .split(newLineRegex);
```

I am not familiar with your testing framework nor the proper setup to run such. But I am finally able to run hint via CLI now.

### DIFF
--- a/packages/connector-puppeteer/src/lib/chromium-finder.ts
+++ b/packages/connector-puppeteer/src/lib/chromium-finder.ts
@@ -116,10 +116,13 @@ const darwin = (browser: Browser) => {
 
     /**
      * The following generates a list of all the installed applications in macOS.
-     * The `grep` command makes sure we only get the lines that end with `.app`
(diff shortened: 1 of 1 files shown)


antross:
Contributing a fix in addition to reporting the problem is extra awesome 👍

Thanks @johnson-jesse!

PR #3918, opened 2020-07-25
Include Analyzer instance resources when calling formatters

### LINKED ISSUE #3910: [Bug] When running Analyzer.format(), the resources from Analyzer.create() are not referenced
# 🐛 Bug report

## Description

*Note: if the intended behavior is to always have all categories when using the programmatic API, then this could be closed & we could stick with our workaround. However, given that formatters have methods to only display categories for actual hints, it seems likely we'd want to "fix" this*

```js
// index.js
const hint = require('hint')
const fs = require('fs')
const userConfig = require('./hintConfig')

const webhint = hint.Analyzer.create(userConfig)
const results = await webhint.analyze('http://example.com')
await webhint.format(results[0].problems)
```

**Actual Result**: When running `webhint.format()`, the `resources` array that gets used during `webhint.analysis()` is not referenced. This means that while `analyze` checks the correct hints based on our configuration (in our case, those from the `accessibility` default), the generated HTML file does not know which hints were used. Meaning that the output file includes *all* categories, rather than just the categories where hints are actually targeted.

**Expected Result**: `webhint.format()` should generate a response that only includes the categories for targeted hints; i.e. if a user doesn't provide a custom list of `resources`, then the instance list of `Analyzer._resources` should be referenced

**Possible line to add:**
https://github.com/webhintio/hint/blob/main/packages/hint/src/lib/analyzer.ts#L269
```js
options = { ...options, resources: this.resources }
```

**Current Workaround:**
```js
await webhint.format(results[0].problems, { resources: webhint.resources })
```

Current:
Screen Shot 2020-07-23 at 1 19 31 PM

After Workaround:
Screen Shot 2020-07-23 at 1 19 06 PM

## Environment

* __`hint` version:__ v6.0.6
* __`Node.js` version:__ v12.16.1
* __`OS`:__ MacOS 10.15.6
* __URL for which webhint failed:__ any

## `webhint` configuration

<details>
<summary>webhint’s configuration</summary>

```js
// hintConfig.js
module.exports = {
	extends: ['accessibility']
}
```

</details>

## Debug output

n/a (Webhint completes successfully, see screenshots for issue)

jackkoppa:
## Pull request checklist

Make sure you:

- n/a Added/Updated related documentation.

Fixes #3910

Formatters need to know which hints were used during analyze, and so by including the `Analyzer.resources` when calling formatters, they can then display only the categories that were actually used.

### DIFF
--- a/packages/hint/src/lib/analyzer.ts
+++ b/packages/hint/src/lib/analyzer.ts
@@ -262,10 +262,9 @@ export class Analyzer {
      * @param {Problem[]} problems Problems to format.
      * @param {FormatterOptions} options Options for the formatters.
      */
-    public async format(problems: Problem[], options?: FormatterOptions): Promise<void> {
-        if (options) {
-            options.language = options.language || this.configuration.language;
-        }
+    public async format(problems: Problem[], options: FormatterOptions = {}): Promise<void> {
+        options.language = options.language || this.configuration.language;
+        options.resources = options.resources || this.resources;
 
         for (const formatter of this.formatters) {
             await formatter.format(problems, options);
--- a/packages/hint/tests/lib/analyzer.ts
+++ b/packages/hint/tests/lib/analyzer.ts
@@ -10,7 +10,8 @@ import {
     AnalyzerError,
     HintResources,
     IFetchOptions,
-    IFormatter
+    IFormatter,
+    FormatterOptions
 } from '../../src/lib/types';
 import { AnalyzerErrorStatus } from '../../src/lib/enums/error-status';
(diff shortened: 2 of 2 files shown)


jackkoppa:
Note - did my best w/ an added test to cover this case (failed before this change), but happy to remove or edit in any way

jackkoppa:
Ran `yarn build-release && npm pack`, then installed locally in our project, and it does indeed resolve #3910 👍

molant:
Thanks @jackkoppa!

PR #3957, opened 2020-08-17
Fix: Cookie name reported in lower case

### LINKED ISSUE #3535: [Bug] Cookie names get reported in lower case
# 🐛 Bug report

## Description

While working on #3534 I noticed that cookie names get transformed to lower case during the error reporting, which might lead to confusing reports as [cookie names are case-sensitive](https://stackoverflow.com/a/11312272) so you could have a valid cookie called `inValidCookie` and an invalid one called `InvalidCookie` and not be able to distinguish them in the report.

As I didn't experience this in a real use case, I'm not sure if this is a bug or intentional.

See https://github.com/webhintio/hint/blob/9eb0a7adf953d0d763d74c219ad54bbcfe5a2f57/packages/hint-validate-set-cookie-header/src/hint.ts#L62
https://github.com/webhintio/hint/blob/9eb0a7adf953d0d763d74c219ad54bbcfe5a2f57/packages/utils-string/src/normalize-string.ts#L6-L11

sarvaje:
Fix #3535

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/hint-validate-set-cookie-header/src/hint.ts
+++ b/packages/hint-validate-set-cookie-header/src/hint.ts
@@ -54,7 +54,7 @@ export default class ValidateSetCookieHeaderHint implements IHint {
         };
 
         /** Concat and unquote the strings after the first `=`. */
-        const splitByEqual = (splitResult: string[]): string[] => {
+        const unquoteAfterSplitByEqual = (splitResult: string[]): string[] => {
             const [key, ...value] = splitResult;
 
             return [key, unquote(value.join('='))];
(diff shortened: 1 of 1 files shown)


packages/hint-validate-set-cookie-header/src/hint.ts
```diff
@@ -53,11 +53,18 @@ export default class ValidateSetCookieHeaderHint implements IHint {
             return value.replace(/(^")|("$)/g, '');
         };
 
+        /** Concat and unquote the strings after the first `=`. */
+        const splitByEqual = (splitResult: string[]): string[] => {
```
antross, line 57: Naming nit (since the split by equal is done before calling the function - also makes in more consistent with `normalizeAfterSplitByEqual` naming):
```suggestion
        const unquoteAfterSplitByEqual = (splitResult: string[]): string[] => {
```

sarvaje: done!

PR #3958, opened 2020-08-17
New: Use mdn-data to derive CSS types

### LINKED ISSUE #3642: Use mdn/data to fix missing reports due to CSS value types
Currently compatibility of CSS values has gaps when the data is part of a CSS type instead of a CSS property (e.g. values for the `<color>` type applies to `color`, `background-color`, `border`, etc.)

Per https://github.com/mdn/browser-compat-data/issues/4309#issuecomment-602761482 we can associate the two using data from https://github.com/mdn/data, which should greatly increase coverage of CSS values for `@hint/hint-compat-api/css`.

antross:
Fix #3642


antross:
This is ready for initial review, though I'll probably add one new test case just to prove we're getting something new with this update (need to search for an appropriate property in the data to use...)

packages/utils-compat-data/scripts/mdn-browser-compat-data.js
```diff
@@ -35,7 +35,7 @@ const isUniversalSupportStatement = (browserName, supportStatement) => {
         return true; // Count unknown as universal support.
     }
 
-    if (version === '1') {
+    if (version === '1' || version === '1.0') {
```
antross, line 38: This saved ~10KB minified and gzipped, making up for the addition of new data so I didn't have to bump our bundle sizes 🎉

packages/utils-compat-data/scripts/mdn-data.js
```diff
@@ -0,0 +1,113 @@
...
+ * Helper for `[].flatMap` until we move to Node 11+.
+ * @param {any[]} arr
+ * @param {(item: any) => any} map
+ */
+const flatMap = (arr, map) => {
+    return [].concat(...arr.map(map));
+};
+
+/**
+ * Helper for `[].matchAll` until we move to Node 11+.
```
antross, line 27:
```suggestion
 * Helper for `string.matchAll` until we move to Node 11+.
```

packages/utils-compat-data/tests/support.ts
```diff
@@ -116,6 +116,26 @@ test('Property value query works (regex token match)', (t) => {
...
+test('Nested property value type query works', (t) => {
+    const unsupported = getUnsupported({
+        property: 'border-image',
+        value: 'conic-gradient(red, orange, yellow, green, blue)'
+    }, ['edge 12', 'edge 79']);
+
+    t.is(unsupported && unsupported.length, 1);
+    t.is(unsupported && unsupported[0], 'edge 12');
+});
+
```
antross, line 138: These are good examples of what we actually gain with this change: the ability to recognize sub-types (e.g. the existing alpha_hex_value which we used to special-case and no longer need to) and functions (e.g. `circle()`). 

These seem to be spaces with notable recent/current cross-browser gaps, so I expect this to prove useful to developers.

antross:
@molant @sarvaje This is ready for a full review when you have time. 

Looks like the Windows CI failure was infrastructure related and not due to this change (though I can re-run later if we want).

packages/utils-compat-data/src/css.ts
```diff
@@ -50,71 +56,126 @@ const getTokens = (nodes: any[]): [string, string][] => {
...
+            const result = typeContext && getPartialValueUnsupported(typeContext, parsedValue, browsers);
 
             if (result) {
                 return result;
             }
         }
     }
 
-    return getPartialValueUnsupported(context, value, browsers);
+    return context ? getPartialValueUnsupported(context, parsedValue, browsers) : null;
```
sarvaje, line 178: Maybe I'm missing something, but why don't do a `return getPartialValueUnsupported(context, parsedValue, browsers)` inside the `if (context)`?

antross: Good question. I was originally maintaining code order for precedence between searching for partial value matches in this entry versus searching for partial value matches in referenced types. More generally this ties back to this API not being designed to report multiple issues (e.g. when two separate parts of the value have different unsupported browsers we just report the first).

Looking at this again I don't think that order is particularly important so I'm going to make the change you suggest. It would only matter in the rare case when both paths had unsupported browsers - either way we'll display something to the user and once they fix one issue, we'll report the next.

captainbrosset (approved):
This looks good to me, though I don't have enough context from working on this before to make valuable review comments. I was able to easily follow the code, it seems clean to me, and has tests, so LGTM! Good work on this.

Issue #3951, opened 2020-08-16
[Bug] Do not show autocomplete warning (input[autocomplete]) if it's off


gianpaj:
# 🐛 Bug report

## Description

The webhint VSCode extensions shows an warning/problem even if the `autocomplete` attrib is `off`

![image](https://user-images.githubusercontent.com/899175/90337311-917ce480-dfd9-11ea-9d7f-d9104ca44377.png)

```html
<input
  type="text"
  name="groupname"
  id="groupname"
  value=""
  autocomplete="off"
  placeholder="Group Name"
/>
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

## `webhint` configuration

none

antross:
Thanks for the feedback @gianpaj! I agree this would be a good change, though I'm wondering if we should just ignore `autocomplete` entirely. I'm not sure if there's any particular value that needs to be flagged since this will gracefully degrade in IE 11 regardless.

@webhintio/contributors anyone else have thoughts/opinions on whether we should ignore this entire attribute, just the "off" value, or something else entirely?

As far as the change itself we'd just need to extend the built-in ignore list for `@hint/hint-compat-api/html` here: https://github.com/webhintio/hint/blob/main/packages/hint-compat-api/src/html.ts#L76.

molant:
Ignoring sounds good to me 😊

captainbrosset:
Ignoring sounds good to me too!

PR #3961, opened 2020-08-19
Ignore `autocomplete` html element

gianpaj:
## Pull request checklist

Make sure you:

As discussed in https://github.com/webhintio/hint/issues/3951

/cc @antross , @molant

### DIFF
--- a/packages/hint-compat-api/src/html.ts
+++ b/packages/hint-compat-api/src/html.ts
@@ -74,6 +74,7 @@ export default class HTMLCompatHint implements IHint {
 
     public constructor(context: HintContext) {
         const ignore = resolveIgnore([
+            'autocomplete',
             'crossorigin',
             'integrity',
             'link[rel]',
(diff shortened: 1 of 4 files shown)


gianpaj:
FYI: i signed it https://cla.js.foundation/webhintio/hint?pullRequest=3961

packages/hint-compat-api/src/html.ts
```diff
@@ -74,6 +74,7 @@ export default class HTMLCompatHint implements IHint {
 
     public constructor(context: HintContext) {
         const ignore = resolveIgnore([
+            'autocomplete',
```
antross, line 77: Slight tweak since we're trying to ignore the `autocomplete` attribute on input elements (still omitting the element name since this should also apply to `<textarea>` and `<select>` so we might as well be generic):
```suggestion
            '[autocomplete]',
```

antross: Actually ignore my suggestion. I just double-checked the code and realized we didn't use selector syntax for this when it's just a bare attribute name 🤦‍♂️ 

Your original change is correct, but I'd still like to add a test case so we're validating it.

PR #3973, opened 2020-08-27
New: Show 3rd party documentation links on the HTML formatter

### LINKED ISSUE #3930: [Feature] Add support for documentation links in the HTML formatter.
After one of these (or all of them) #3903, #3905 and #3909 are merged, we will need a way to show those links in the HTML Formatter.

sarvaje:
## Pull request checklist

Make sure you:

Fix #3930
image

### DIFF
--- a/packages/formatter-html/src/_locales/en/messages.json
+++ b/packages/formatter-html/src/_locales/en/messages.json
@@ -31,6 +31,10 @@
         "description": "Text for the expand all button",
         "message": "expand all"
     },
+    "furtherReading": {
+        "description": "Header for further reading in hints with external documentation",
+        "message": "Further Reading"
+    },
     "gatheringResults": {
         "description": "Text to show when the category is waiting for results",
         "message": "Gathering results"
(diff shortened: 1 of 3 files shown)


packages/formatter-html/src/_locales/en/messages.json
```diff
@@ -31,6 +31,10 @@
         "description": "Text for the expand all button",
         "message": "expand all"
     },
+    "furtherReading": {
+        "description": "Header for further reading in hints with external documentation",
+        "message": "Further Reading"
```
antross, line 36: I don't think we have to solve this now, but we may want to move this string to `@hint/utils-i18n` so we can share it between formatters and with the browser extension (and avoid localizing the string multiple times).

sarvaje: Do we support that?

antross: We're not explicitly exporting a helper to get it at the moment, but `import getMessage as getSharedMessage from '@hint/utils-i18n/dist/src/i18n.import';` then using `getSharedMessage("furtherReading")` should work. 

This would be better still if we just export a `getSharedMessage` helper from `@hint/utils-i18n`. Currently we only share category names so we exported a much more scoped `getCategoryName` helper.

PR #3977, opened 2020-08-27
Upgrade: Update dependencies

### LINKED ISSUE #3976: Update dependencies
Looks like dependabot is missing some updates and yarn sometimes miss to update some subdependencies.

We should catch up with the new versions of the packages and update yarn.lock

sarvaje:
Fix #3976

## Pull request checklist

Make sure you:


.dependabot/config.yml
```diff
@@ -17,7 +17,7 @@ update_configs:
           dependency_name: "jquery"
       - match:
           dependency_name: "@types/node"
-          version_requirement: "^10.12.18"
+          version_requirement: "^14.6.1"
```
antross, line 20: Should we just remove `@types/node` from the dependabot ignore list?

.eslintrc.json
```diff
@@ -249,19 +249,31 @@
...
+              "custom": {
+                "regex": "^I[A-Z]",
+                "match": true
+              }
+            },
+            {
+                "selector": "class",
+                "format": ["PascalCase"]
+            }
+          ]
```
antross, line 277: Just to clarify, does this replace `class-name-casing`, `inferface-name-prefix`, and `member-naming`?

.eslintrc.json
```diff
@@ -249,19 +249,31 @@
         "wrap-regex": "error",
         "yoda": ["error", "never"],
 
-        "@typescript-eslint/class-name-casing": "error",
         "@typescript-eslint/explicit-member-accessibility": "error",
-        "@typescript-eslint/interface-name-prefix": [ "error", "always" ],
         "@typescript-eslint/member-delimiter-style": "error",
-        "@typescript-eslint/member-naming": "error",
-        "@typescript-eslint/no-angle-bracket-type-assertion": "error",
```
antross, line 257: Was there a rename/replacement for this rule? I may be missing something, but I'm not seeing it anywhere...

packages/extension-vscode/src/utils/analyze.ts
```diff
@@ -8,7 +8,7 @@ import { loadWebhint } from './webhint-packages';
 import { problemToDiagnostic } from './problems';
 import { promptRetry } from './prompts';
 
-const analyze = async (textDocument: TextDocument, webhint: import('hint').Analyzer): Promise<PublishDiagnosticsParams> => {
+const analyze = async (textDocument: TextDocument, webhint: import('hint').Analyzer): Promise<PublishDiagnosticsParams> => { // eslint-disable-line no-use-before-define
```
antross, line 11: I'm surprised eslint flagged this line for no-use-before-define. Any idea what was going on here or am I missing a reference that isn't defined until later in the file?

packages/parser-javascript/src/parser.ts
```diff
@@ -30,7 +30,7 @@ export default class JavascriptParser extends WebhintParser<ScriptEvents> {
         try {
             await this.engine.emitAsync(`parse::start::javascript`, { resource });
 
-            const options: Options = { locations: true, ranges: true };
+            const options: Options = { ecmaVersion: 2020, locations: true, ranges: true };
```
antross, line 33: Should we consider using "latest" here per https://github.com/acornjs/acorn/tree/master/acorn#interface? Otherwise we're going to need some way to remember to bump this to the next version later.

```suggestion
            const options: Options = { ecmaVersion: "latest", locations: true, ranges: true };
```

antross (changes requested):
Thanks for taking the time to get these up-to-date @sarvaje! 
Just a few questions and small tweaks then this should be good-to-go.

.eslintrc.json
sarvaje: Yes, the new rule is `@typescript-eslint/consistent-type-assertions`. This one was replace in v2.
https://github.com/typescript-eslint/typescript-eslint/releases/tag/v2.0.0

.dependabot/config.yml
sarvaje: IMO, we should, I think that is there because in the pass, that package gave us a lot of issue, but looks like it is working fine now.

packages/extension-vscode/src/utils/analyze.ts
sarvaje: I think it is an issue in `eslint` or `@typescript-eslint`. I tried to change `Analyzer` for another property and it was ok. TBH, I don't know why it is complaining.

packages/parser-javascript/src/parser.ts
sarvaje: Sure, I use `2020`  because that was what the warning was saying hehehe.

.dependabot/config.yml
sarvaje: do you want me to remove that?

packages/extension-vscode/src/utils/analyze.ts
antross: Got it. I could see why eslint might choke on the `import('hint').Analyzer` syntax. Hopefully this will get fixed later, but for now I'm fine with the one-off comment to disable.

.dependabot/config.yml
antross: Yes please, thanks!

antross (approved):
LGTM. Thanks for the quick updates!

Issue #3962, opened 2020-08-20
[Feature] Markdown Formatter


EddyHaigh:
# 🚀 Feature request

## Description

 What do you want to be added?
A markdown formatter.

 Are you willing to work on this?
Yes

## What scenarios will this solve?

 Why do you want/need this?
This will solve issues where you can't display rich HTML formats such as in websites with particular Content Security Policies as well as environments such as Azure Pipelines where the pipeline summary page can only display markdown. [Webhint ADO Extension Issue](https://github.com/webhintio/webhint-ado-extension/issues/14).

### Example Markdown Report

``` md
# Webhint Report - {URL} - {Date}

[Table of Contents]

## Compatibility

### Supported CSS features - Errors 1, Warnings 1

[Learn why this is important and how to fix](https://webhint.io/docs/user-guide/hints/hint-compat-api/css/)

#### **ERROR** - 'position: sticky' is not supported by Safari, iOS Safari. Add 'position: -webkit-sticky' to support Safari 6.1+, iOS Safari 6.1+.'

https://webhint.io/static/styles/common-0788930e1c.css:1:11064

``` css
@media (max-width:63em) {
    .header {
        position: sticky;
    }
}
\```

#### **WARNING** - 'cursor: not-allowed' is not supported by Firefox Android.'

https://webhint.io/static/styles/common-0788930e1c.css:1:25382

``` css
button:disabled,
select:disabled {
    cursor: not-allowed;
}
\```

---

Powered By [Webhint](https://webhint.io/) v2.1.3
```

molant:
Hi @EddyHaigh 👋

I see that you are willing to work on this, is that still the case? If so, we will be more than happy to help you along the way 😊

You probably have seen this, but we have some documentation about [developing a formatter](https://webhint.io/docs/contributor-guide/how-to/formatter/).
Also, copy pasting the json formatter is a good starting point. Just make sure to update the names and the `package.json`.

Let us know if you run into any issues!

EddyHaigh:
@molant , yeah im still willing to work on this. I'm just trying to think of the best way to do this before I start. Initially, I was thinking a string builder style way but that could easily become complicated and hard to maintain. Then I took a look at the HTML formatter and thought about using the EJS templating engine to develop the Markdown. I know it might sound silly but it would be more maintainable and easier to read. However, I am way open to suggestions as I'm not a js/ts person.

molant:
[...]

PR #4040, opened 2020-09-27
New: Markdown Formatter

EddyHaigh:
## Pull request checklist

Make sure you:

This change adds a Markdown formatter related to issue: #3962 
The new formatter does add some duplication of code between the HTML and Markdown formatter and would benefit from creating a shared formatter utils package.


packages/formatter-markdown/package.json
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "@hint/formatter-markdown",
+  "version": "1.0.0",
```
molant, line 3: I believe the first version needs to be 0.0.1 or similar so it gest bumped automatically by our release script.

packages/formatter-markdown/src/formatter.ts
```diff
@@ -0,0 +1,224 @@
...
+        languagesToCheck.push(languageParts[0]);
+    }
+
+    // Default to 'en'.
+    let existingLanguage = 'en';
+
+    for (const lang of languagesToCheck) {
+        const file = path.join(rootPath, lang, messagesFileName);
+
+        if (fs.existsSync(file)) { // eslint-disable-line no-sync
```
molant, line 72: Can you please add a comment saying `fs.exists` is deprecated and that's why we use the sync version?

packages/formatter-markdown/src/formatter.ts
```diff
@@ -0,0 +1,224 @@
...
+            throw err;
+        }
+    }
+
+    /**
+     * Creates the markdown report of the webhint scan.
+     * @param result The webhint scan result.
+     */
+    /* istanbul ignore next [too hard to test] */
+    private createMarkdown(result: AnalysisResult) {
```
molant, line 166: Can you add a comment with an example of how the markdown should look like?

packages/formatter-markdown/src/formatter.ts
```diff
@@ -0,0 +1,224 @@
...
+        let markdown = '';
+
+        markdown += MarkdownHelpers.createHeader(`Webhint Report - ${result.date}`, HeaderCount.Title);
+        markdown += MarkdownHelpers.newLine;
+        markdown += `${this.getMessage('hints')}: ${result.hintsCount}`;
+        markdown += MarkdownHelpers.newLine;
+
+        result.categories.forEach((category) => {
```
molant, line 174: What do you think about refactoring this a bit to make it easier to read?
Maybe replace each `forEach` with a `map` so it can return the content for each one and then you can add them to `markdown` after a `[].join(MardownHelpers.newlLine)` or something similar?

packages/formatter-markdown/package.json
```diff
@@ -0,0 +1,86 @@
...
+    "url": "https://github.com/webhintio/hint.git"
+  },
+  "dependencies": {
+    "@hint/utils": "^7.0.6",
+    "@hint/utils-fs": "^1.0.5",
+    "@hint/utils-i18n": "^1.0.5",
+    "@hint/utils-types": "^1.1.0",
+    "lodash": "^4.17.20",
+    "fs-extra": "^9.0.1",
+    "moment": "^2.29.0"
```
molant, line 46: This is a note for the core team:
- We are using version 2.28.0 in another package (maybe that's why `yarn.lock` gets updated?)
- [moment is considered legacy now by their team](https://momentjs.com/docs/#/-project-status/), we should think about moving to other alternatives

molant (commented):
Took a quick look between meetings and it's looking great.
Thanks a lot @EddyHaigh !

I left a couple comments but you might want to hold until @sarvaje and @antross can spend sometime before doing anything.

packages/formatter-markdown/package.json
sarvaje: Maybe we should create an issue to replace moment?

molant: Done in https://github.com/webhintio/hint/issues/4044

sarvaje (changes requested):
Looking good, just a few things.

sarvaje (changes requested):
Thanks for the update!

Just a few more things :)

packages/formatter-markdown/src/formatter.ts
```diff
@@ -148,6 +148,8 @@ export default class MarkdownFormatter implements IFormatter {
                 const destination = path.join(destDir, `${name}.md`);
 
                 await fs.outputFile(destination, markdown);
+
+                logger.log(getMessageFormatter('youCanView', this.language, destination));
```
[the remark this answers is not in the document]
EddyHaigh: It's at the bottom of the [messages.json file](https://github.com/webhintio/hint/pull/4040/files#diff-d58af875a191c4335c51ce5e5c8376dc).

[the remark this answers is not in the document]
sarvaje: Sorry, I didn't saw it :s

packages/formatter-markdown/src/formatter.ts
EddyHaigh: @molant I've changed this to use map but I'm not sure if it is right or not.

packages/formatter-markdown/package.json
EddyHaigh: I've seen we'have removed it from the HTML formatted so I've removed it from this formatter.

sanmai-NL:
@EddyHaigh Looking forward to this.

PR #4046, opened 2020-10-01
New: Added classList.add/remove leading '.' warning

### LINKED ISSUE #3997: [Hint] Pitfalls - classList.add/remove with leading "."
I've hit this one a few times since I'm so used to writing selectors instead of plain class names. It actually adds/remove a className which includes the leading "." as part of the name which is almost never what you want. Seems like a warning/error here would be helpful as you can easily be left scratching your head about this for awhile.

E.g.
```html
<div id="foo" class="bar">
```

### Incorrect
```js
// Adds a class named ".foo" producing <div id="foo" class="bar .foo">
document.getElementById("#foo").classList.add(".foo");

// Tries to remove a class named ".bar" - silently doesn't do anything
document.getElementById("#foo").classList.remove(".bar");
```

### Correct 

```js
document.getElementById("#foo").classList.add("foo");
document.getElementById("#foo").classList.remove("bar");
```

May also be a good idea to check for a similar mistake in HTML markup too for completeness.

ruppysuppy:
## Pull request checklist

Make sure you:

Added classList.add/remove leading '.' warning hint

______________________________

Fixes #3997


ruppysuppy:
@antross, @sarvaje  the tests I added are failing I cannot figure out why and I was unable to use `walk.ancestors` to check if the add/remove is called on classList. Could you help with these issues please?

packages/hint-classlist-pitfall/src/meta.ts
```diff
@@ -0,0 +1,26 @@
...
+    },
+    /* istanbul ignore next */
+    getDescription(language: string) {
+        return getMessage('description', language);
+    },
+    /* istanbul ignore next */
+    getName(language: string) {
+        return getMessage('name', language);
+    },
+    id: 'classlist-add-remove-pitfall',
```
sarvaje, line 21: The id has to be the same as the folder without `hint-`, in this case `id: 'classlist-pitfall',`.

This is also why the test are failing.

packages/hint-classlist-pitfall/src/hint.ts
```diff
@@ -0,0 +1,76 @@
...
+                CallExpression(node, ancestor) {
+                    if (!('property' in node.callee && 'name' in node.callee.property)) {
+                        return;
+                    }
+
+                    if (node.callee.property.name !== 'add' && node.callee.property.name !== 'remove') {
+                        return;
+                    }
+
+                    const arg = node.arguments[0];
```
sarvaje, line 43: This is not checking for `classList` any `add` or `remove` method will be checked (and reported if it starts with `.`.

sarvaje (changes requested):
Thanks for the PR!!

There are a few request.

The tests are failing because the hint name has to be the same as the folder, but without the `hint-`  at the beginning.

Also, I'm not pretty sure about the word `pitfall` in the name of the hint.

ruppysuppy:
Thanks a lot @sarvaje for the review :)

I have implemented the changes you suggested, but I unable to come up with a name (in place of pitfall). Could you suggest something? Another issue is how do I check whether the ancestor is `classList`?

molant:
> Could you suggest something?

What about `leading-dot-classlist` or similar?

> Another issue is how do I check whether the ancestor is `classList`?

@antross can probably help with this when he has time.

sarvaje:
> What about leading-dot-classlist or similar?

I like that name better.

>Another issue is how do I check whether the ancestor is classList?

You can look into the property `object` in `node.callee` then check if the property name is `classList`, something like this:
`node.callee.object.property.name === 'classList'` (you need to make sure that `object` and `object.property` exist.

packages/hint-leading-dot-classlist/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "test-only": "nyc ava",
+    "test-release": "npm run i18n && npm run lint && npm run build-release && ava",
+    "init": "npm install && npm run build",
+    "watch": "npm run build && npm-run-all --parallel -c watch:*",
+    "watch:assets": "npm run build:assets -- -w --no-initial",
+    "watch:test": "ava --watch",
+    "watch:ts": "npm run build:ts -- --watch"
+  },
+  "version": "0.0.1"
+}
```
sarvaje, line 78: Add a new line at the end.
```suggestion
}

```

packages/hint-leading-dot-classlist/tests/fixtures/testPassing.js
```diff
@@ -0,0 +1 @@
+document.getElementById('container').classList.add('foo')
```
antross, line 1: Let's add a test to verify we pass for non-classList objects calling `add` with a leading `.`.
```ts
const set = new Set();
set.add('.foo');
```

packages/hint-leading-dot-classlist/tests/fixtures/testFailing.js
```diff
@@ -0,0 +1 @@
+document.getElementById('container').classList.add('.foo')
```
antross, line 1: Let's also verify this passes/fails as expected for `remove`.

packages/hint-leading-dot-classlist/tests/fixtures/testIrrelevant.js
```diff
@@ -0,0 +1,5 @@
+function add(str1, str2){
+    return str1 + str2
+}
+
+add(".1", ".2")
```
sarvaje, line 5:
```suggestion
add(".1", ".2");

```

packages/hint-leading-dot-classlist/tests/fixtures/testFailing.js
sarvaje, line 1:
```suggestion
document.getElementById('container').classList.add('.foo');

```

packages/hint-leading-dot-classlist/tests/fixtures/testPassing.js
sarvaje, line 1:
```suggestion
document.getElementById('container').classList.add('foo');

```

antross (commented):
Looking good so far, thanks for putting this together!

I think there are a few more test cases to add here to ensure we're triggering this only when we want. Also, I may make some minor tweaks to the README but I'll take care of that before merging.

sarvaje (commented):
Looking good, just a few nit's

ruppysuppy:
> You can look into the property `object` in `node.callee` then check if the property name is `classList`, something like this:
> `node.callee.object.property.name === 'classList'` (you need to make sure that `object` and `object.property` exist.

The test 
```javascript
set.add('.foo')
```
is failing. There was a mistake, I was checking for:
```javascript
node.callee.property.name !== 'add' && node.callee.property.name !== 'remove' && node.callee.property?.name !== 'classList'
```
inplace of
```javascript
node.callee.property.name !== 'add' && node.callee.property.name !== 'remove' && node.callee.object?.property?.name !== 'classList'
```

Failed

antross:
> Is node.callee.object private/protected? It cannot be accessed, even though its visible on logging it

@ruppysuppy I think this is just a bug in the types we have for ESTree (where the types are out-of-sync with the actual library). As a quick fix you can try `(node.callee as any).object` to see if that works. 

Ideally we'd also fix the types themselves at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/estree/index.d.ts so we can remove the `any` cast but I'm okay saving that for a follow-up PR.

packages/hint-leading-dot-classlist/src/hint.ts
```diff
@@ -36,7 +36,9 @@ export default class LeadingDotClasslist implements IHint {
                         return;
                     }
 
-                    if (node.callee.property.name !== 'add' && node.callee.property.name !== 'remove' && node.callee.property?.name !== 'classList') {
+                    if (!(node.callee.property.name === 'add' ||
+                    node.callee.property.name === 'remove')
+                    ) {
```
sarvaje, line 41: Nit: Move this back to one line or add some to the second line.
```suggestion
                    if (!(node.callee.property.name === 'add' || node.callee.property.name === 'remove')) {
```

packages/hint-leading-dot-classlist/tests/fixtures/testFailing01.js
```diff
@@ -0,0 +1 @@
+document.getElementById('container').classList.add('.foo');
```
sarvaje, line 1: nit: Change the name of the file to `test-failing-add.js`

packages/hint-leading-dot-classlist/tests/fixtures/testFailing02.js
```diff
@@ -0,0 +1 @@
+document.getElementById('container').classList.remove('.foo');
```
sarvaje, line 1: nit: Change the name of the file to `test-failing-remove.js`

sarvaje (commented):
Just a few nits.

Also, If I remember correctly, our convention for composed file name is to separate the names with a `-` instead of using camel case.

ruppysuppy:
> Ideally we'd also fix the types themselves at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/estree/index.d.ts so we can remove the `any` cast but I'm okay saving that for a follow-up PR.

Could you tell me what needs to be fixed, so I create an issue & pr for it before working on the follow up pr?

sarvaje (approved):
LGTM, thanks for you contribution @ruppysuppy

packages/hint-leading-dot-classlist/src/hint.ts
```diff
@@ -0,0 +1,76 @@
...
+
+            walk.simple(ast, {
+                CallExpression(node) {
+                    if (!('property' in node.callee && 'name' in node.callee.property)) {
+                        return;
+                    }
+
+                    if ((node.callee.property.name !== 'add' && node.callee.property.name !== 'remove') || (node.callee as any).object?.property?.name !== 'classList') {
+                        return;
+                    }
```
antross, line 41: Hey @ruppysuppy, sorry for my slow response. I dug into the type issue a bit more and realized the types are correct but we need a couple of additional checks to narrow them down. This is because there are multiple possible types for `callee`, `object`, etc. which don't all have the same properties.
```suggestion
                    if (node.callee.property.name !== 'add' && node.callee.property.name !== 'remove') {
                        return;
                    }
                    
                    if (!('object' in node.callee && 'property' in node.callee.object && 'name' in node.callee.object.property && node.callee.object.property.name === 'classList')) {
                        return;
                    }
```

antross (approved):
LGTM - Thanks for the update @ruppysuppy and for contributing a great new hint! 🎉

FYI I went ahead and made some tweaks to the README to avoid repeating content from MDN. I'll merge once tests pass. 

As a follow-up, do you want to look into enabling this hint by default in webhint? You can check https://github.com/webhintio/hint/commit/12afa7fcce1cb9d06b3ec4f7c6d63ca79d3ec2f9#diff-fbe1e25ee835847f4c2014117109c5dd46e26bae320290379599ca65d813902f for an example of which files need updated. Only now we also have `utils-worker` to update too https://github.com/webhintio/hint/blob/main/packages/utils-worker/package.json#L13.

Issue #4072, opened 2020-10-07
[Hint] IE Flexbox compat hint


captainbrosset:
IE10-11 have a lot of compatibility issues with other browsers when it comes to Flexbox.
Most of these are fixed in the new Edge, and will never be addressed in those older browsers.

When asked about what frustrates them the most about web dev, people tend to mention web compat a lot. And IE is almost always cited as a source of frustration.

Also, these IE-specific compat bugs are documented and workarounds exist: see https://github.com/philipwalton/flexbugs

I had an idea to create a series of hints, specifically for each of these bugs, that would check the computed styles and warn users against IE compat problems with flexbox, and offer workarounds.
I filed a few issues related to this today.

Thinking more about this, and discussing with @antross, we agreed that this might actually not be the right level of investment. Such a hint would be complex and costly to write.

IE is deprecated, slow and insecure. Using it today is really not a good idea. It is going away at some point for sure. Such a hint would, indirectly, encourage people to support and therefore use it. Not great for the web.

It's hard to verify, but sites that, still today, require IE10-11 to work are probably older intranets that aren't maintained anymore. They still need to run (and there's a way in Edge to run in IE mode) of course, but they're not likely to be actively worked on, and even less likely that someone would be adding new flexbox code to them.

Finally, in Edge, by default IE is excluded from the Issues panel webhint integration.

So, based on the above, the decision is create a much simpler hint that warns/informs you if you do configure IE as a target browser *and* you use flexbox.
The warning could be along the lines of "IE10 suffers from a number of compatibility bugs with Flexbox, find more information about them and potential workaround on this site ..."

antross:
Yeah, I like the more general detect flex usage + IE in target browsers and notify users with a link to known issues. I think setting the severity to `Hint` would be appropriate in this case.

molant:
I'm not sure about `Hint` as the level.
IE is not the default configuration so this is intentional by the user, especially if the code is not under a media query or similar.

antross:
@molant so are you thinking `Warning` instead? I think the challenge is we don't know for sure that they're hitting one of these issues, so I'd definitely avoid `Error`. 

Personally I feel `Warning` is a bit strong for something that won't go away even if you've addressed it already. We could go with `Information` since that's technically what we're doing here - we won't have decided whether or not you're affected, just that you should probably be aware of these issues if you aren't already.

molant:
[...]

PR #4073, opened 2020-10-08
New IE flexbox compat hint

captainbrosset:
This is an attempt to address #4072.

A hint that warns users when they do configure IE as a target browser and flexbox is used in one of the stylesheets.
This is mostly to get the ball rolling and gather feedback.

Let me know what you think.

### DIFF
--- a/packages/hint-ie-flexbox-compat/src/_locales/en/messages.json
+++ b/packages/hint-ie-flexbox-compat/src/_locales/en/messages.json
@@ -9,6 +9,6 @@
     },
     "usingFlexWithIE": {
         "description": "String warning users that flex is used and IE is a target browser",
-        "message": "The Internet Explorer brower versions 11 and older suffer from a number of CSS Flexbox compatibility bugs. The layout on this page may be different between Internet Explorer and other, more recent, browsers. The size or alignment of elements may be incorrect and content may overflow.\nSupport for older versions of Internet Explorer ended and Microsoft no longer provides security updates or technical support for these versions."
+        "message": "CSS Flexbox may render differently in Internet Explorer than more recent browsers. Check this page in IE then see the documentation for workarounds if needed."
     }
 }
--- a/packages/hint-ie-flexbox-compat/src/_locales/en/messages.json
+++ b/packages/hint-ie-flexbox-compat/src/_locales/en/messages.json
@@ -1,0 +1,14 @@
+{
+    "description": {
+        "description": "Metadata description",
+        "message": "Informs users that Internet Explorer suffers from Flexbox compatibility bugs."
+    },
+    "name": {
+        "description": "Metadata name",
+        "message": "IE Flexbox Compatibility"
+    },
+    "usingFlexWithIE": {
+        "description": "String warning users that flex is used and IE is a target browser",
+        "message": "TBD"
+    }
+}
--- a/packages/hint-ie-flexbox-compat/src/_locales/en/messages.json
+++ b/packages/hint-ie-flexbox-compat/src/_locales/en/messages.json
@@ -9,6 +9,6 @@
     },
     "usingFlexWithIE": {
         "description": "String warning users that flex is used and IE is a target browser",
-        "message": "TBD"
+        "message": "The Internet Explorer brower versions 11 and older suffer from a number of CSS Flexbox compatibility bugs. The layout on this page may be different between Internet Explorer and other, more recent, browsers. The size or alignment of elements may be incorrect and content may overflow.\nSupport for older versions of Internet Explorer ended and Microsoft no longer provides security updates or technical support for these versions."
     }
 }
--- a/packages/hint-ie-flexbox-compat/src/hint.ts
+++ b/packages/hint-ie-flexbox-compat/src/hint.ts
@@ -1,0 +1,84 @@
+import { Category } from '@hint/utils-types';
+import { HintContext } from 'hint/dist/src/lib/hint-context';
+import { HintScope } from 'hint/dist/src/lib/enums/hint-scope';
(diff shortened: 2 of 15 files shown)


packages/hint-ie-flexbox-compat/src/hint.ts
```diff
@@ -0,0 +1,84 @@
...
+import { getMessage } from './i18n.import';
+
+/*
+ * ------------------------------------------------------------------------------
+ * Public
+ * ------------------------------------------------------------------------------
+ */
+
+export default class IEFlexboxCompatHint implements IHint {
+    public static readonly meta: HintMetadata = {
```
sarvaje, line 19: Meta needs to be in its own file (`meta.ts`) and imported. (e.g. https://github.com/webhintio/hint/blob/main/packages/hint-manifest-is-valid/src/hint.ts).
We use the meta file to generate the documentation in the website.

captainbrosset: Thanks! Done.

packages/hint-ie-flexbox-compat/src/_locales/en/messages.json
```diff
@@ -0,0 +1,14 @@
...
+        "description": "Metadata description",
+        "message": "Informs users that Internet Explorer suffers from Flexbox compatibility bugs."
+    },
+    "name": {
+        "description": "Metadata name",
+        "message": "IE Flexbox Compatibility"
+    },
+    "usingFlexWithIE": {
+        "description": "String warning users that flex is used and IE is a target browser",
+        "message": "The Internet Explorer brower versions 11 and older suffer from a number of CSS Flexbox compatibility bugs. The layout on this page may be different between Internet Explorer and other, more recent, browsers. The size or alignment of elements may be incorrect and content may overflow.\nSupport for older versions of Internet Explorer ended and Microsoft no longer provides security updates or technical support for these versions."
```
antross, line 12: Let's keep the message a bit more concise and leave the details for the README.

```suggestion
        "message": "This page uses CSS Flexbox which may render differently in Internet Explorer than more recent browsers. Double-check this page in IE then see the documentation for workarounds if needed."
```

packages/hint-ie-flexbox-compat/src/hint.ts
```diff
@@ -0,0 +1,65 @@
...
+    public constructor(context: HintContext<StyleEvents>) {
+        if (!context.targetedBrowsers.some((browser) => {
+            return browser.toLowerCase().startsWith('ie ');
+        })) {
+            return;
+        }
+
+        let reported = false;
+
+        context.on('parse::end::css', ({ ast, element, resource }: StyleParse) => {
```
antross, line 28: Nit: I don't think you need to specify the type here as it should be inferred from the event name.
```suggestion
        context.on('parse::end::css', ({ ast, element, resource }) => {
```

packages/hint-ie-flexbox-compat/src/hint.ts
```diff
@@ -0,0 +1,65 @@
...
+    public static readonly meta = meta;
+
+    public constructor(context: HintContext<StyleEvents>) {
+        if (!context.targetedBrowsers.some((browser) => {
+            return browser.toLowerCase().startsWith('ie ');
+        })) {
+            return;
+        }
+
+        let reported = false;
```
antross, line 26: In the VS Code extension or CLI in watch mode this will never get reset between runs (causing the error message to disappear). Add a 'scan::end' listener to the context to reset this.

antross (commented):
Looks like a good start. We'll want a good README.md added as well (which will be published as the doc page on webhint.io). I'd expect to call out an example issue there then link to the full list of bugs + workarounds.

packages/hint-ie-flexbox-compat/src/hint.ts
captainbrosset: Thanks! Will do.

packages/hint-ie-flexbox-compat/src/hint.ts
captainbrosset: Sounds good, I will address this.

captainbrosset:
Oh you're right, I forgot to add a README file. I'll simplify the message and instead write a longer README.

antross:
@captainbrosset FYI just looks like this is hitting a linting failure on the updated README:
```

README.md: 3: MD013/line-length Line length [Expected: 80; Actual: 96]
README.md: 7: MD013/line-length Line length [Expected: 80; Actual: 87]
README.md: 8: MD013/line-length Line length [Expected: 80; Actual: 107]
README.md: 11: MD013/line-length Line length [Expected: 80; Actual: 109]
README.md: 15: MD013/line-length Line length [Expected: 80; Actual: 104]
README.md: 16: MD013/line-length Line length [Expected: 80; Actual: 105]
README.md: 20: MD013/line-length Line length [Expected: 80; Actual: 91]
```

antross (approved):
LGTM. Thanks @captainbrosset!

sarvaje (approved):
LGTM!

captainbrosset:
As a follow-up, should we add this hint to the default configuration? And if so, am I right in assuming this is the place to do so: `packages\configuration-web-recommended\index.json` ?

antross:
@captainbrosset that's one place, but there are also a few others to hit our default configurations in other contexts (browser extension, VS Code, etc.). There's probably a good TODO in here for establishing a common base config the others can extend from 😉

See this PR for a recent example: https://github.com/webhintio/hint/pull/4079

The online scanner also needs separately updated after a release. For now you can just log an issue like this one and @sarvaje will take care of it: https://github.com/webhintio/serverless-online-service/issues/199

PR #4082, opened 2020-10-22
New: Perf tests

sarvaje:
## Pull request checklist

Make sure you:

This PR adds some initial perf tests and configures azure pipelines to run them on main branch.

### DIFF
--- a/packages/utils-worker/tests/helpers/runner.ts
+++ b/packages/utils-worker/tests/helpers/runner.ts
@@ -1,0 +1,164 @@
+import { launch, Page } from 'puppeteer';
+
+import { FetchEnd } from 'hint';
+import { Server } from '@hint/utils-create-server';
+import { createHelpers, DocumentData } from '@hint/utils-dom';
+import { Problem } from '@hint/utils-types';
+
+import { WorkerEvents, HostEvents, Config } from '../../src/shared/types';
+import { Test, RunResult } from './types';
+import { readFile } from './fixtures';
+
+declare const __webhint: {
+    snapshotDocument(document: Document): DocumentData;
+};
+
+const runWorker = async (page: Page, config: Partial<Config>, test: Test) => {
+    return await page.evaluate((config: Partial<Config>, test: Test) => {
+        const mockFetchEnd = (): FetchEnd => {
+            return {
+                element: null,
+                request: {
+                    headers: {},
+                    url: location.href
+                },
+                resource: location.href,
+                response: {
+                    body: {
+                        content: test.html,
+                        rawContent: null as any,
+                        rawResponse: null as any
+                    },
+                    charset: '',
+                    headers: {
+                        'Cache-Control': 'no-cache',
+                        'Content-Type': 'text/html;charset=utf-8'
+                    },
+                    hops: [],
+                    mediaType: '',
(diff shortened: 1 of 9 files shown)


packages/utils-worker/tests/test-perf.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    return result;
+};
+
+const prefixOrderCSS = [generateCSS('./index.css')];
+const leadingDotClasslistJS = [generateJS('./index.js')];
+
+const tests: Test[] = [{
+    expectedHints: ['axe/text-alternatives'],
+    expectedTime: 26000,
```
antross, line 170: Can we set this as close as possible to the real expected time, then fail the test if we're +/- a certain amount (e.g. 10% faster or slower)? My thinking here is we should fail the test if we improve performance so we are reminded to update the values to catch future regressions. We should just make sure the output makes it clear that this was an improvement and the most likely resolution is to update the test.

packages/utils-worker/tests/test-perf.ts
```diff
@@ -0,0 +1,310 @@
...
+}, {
+    expectedHints: ['axe/text-alternatives'],
+    expectedTime: 40000,
+    html: generateDeepHTML(),
+    name: 'Axe deep perf test',
+    timeout: 60000
+}, {
+    expectedHints: ['button-type'],
+    expectedTime: 3000,
+    hints: {
```
antross, line 183: Nit: since we're disabling all the axe tests multiple times, let's re-use a shared object so we don't have to repeat this. Even if we need to make slight modifications on top of it we can use the object-spread `{...obj}` syntax to blend them.

packages/utils-worker/tests/helpers/runner.ts
```diff
@@ -0,0 +1,216 @@
...
+
+                if (message.requestConfig) {
+                    sendMessage({
+                        config: {
+                            resource: location.href,
+                            ...config
+                        }
+                    });
+                } else if (message.ready) {
+                    startTime = Date.now();
```
antross, line 109: Let's use the performance APIs to track the time instead of `Date`.

antross (commented):
Overall looks good, thanks @sarvaje! Some minor suggestions, but then this should be ready to merge.

antross (approved):
LGTM, thanks for the updates! Want to promote this to a full PR before we consider merging?

sarvaje:
@antross @molant this PR is ready to review!

packages/utils-worker/tests/test-perf.ts
```diff
@@ -0,0 +1,315 @@
...
+
+
+    copyResults.sort((a, b) => {
+        return a.totalTime - b.totalTime;
+    });
+
+    copyResults.splice(0, 1);
+    copyResults.splice(copyResults.length - 1, 1);
+
+    return copyResults;
```
antross, line 252: Nit: I think using `slice` to make a copy makes this a bit simpler.

```suggestion
    return copyResults.slice(1, -1);
```

antross (approved):
Still LGTM, but glad to see the updates 😉

@molant do you think you'll have time to review this or should I go ahead and merge?

PR #4088, opened 2020-10-29
Chore: Return real promises in no-broken-links#createReports

### LINKED ISSUE #2034: [Feature] `no-broken-links#createReports` should return only real promises in the array instead of `Promise.resolve()`
With the changes in #1943 there's a lot of code in `createReports` that returns `Promise.resolve()`. We should return only real Promises instead of doing that.
For that we could change `return urls.map((url) => {` to something like:

```ts
const requests: Promise<void>[] = [];
urls.forEach((url) => {
// magic here
requests.push(realPromiseToAdd);
}

return requests;
```

jnv:
Fix #2034

## Pull request checklist

Make sure you:

In `no-broken-links` hint use `for..of` loop for collecting requests in `createReports` function which return "real" promises instead of unrelated `Promise.resolve()`.

### DIFF
--- a/packages/hint-no-broken-links/src/hint.ts
+++ b/packages/hint-no-broken-links/src/hint.ts
@@ -216,13 +216,15 @@ export default class NoBrokenLinksHint implements IHint {
         };
 
         const createReports = (element: HTMLElement, urls: string[]): Promise<void>[] => {
-            return urls.map((url) => {
+            const requests: Promise<void>[] = [];
+
+            for (const url of urls) {
                 /*
                  * If the URL is not HTTP or HTTPS (e.g. `mailto:`),
                  * there is no need to validate.
                  */
                 if (!isRegularProtocol(url)) {
-                    return Promise.resolve();
+                    continue;
                 }
 
                 const fetched = getFetchedURL(url);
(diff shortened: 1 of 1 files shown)


sarvaje:
What are the benefits of this change (I like it, but I'm curious)

jnv:
> What are the benefits of this change? (I like it, but I'm curious)

Well, I have just followed #2034 so @molant can give better explanation. From my point of view there are at least two benefits:

- The returned promises are actually connected to the requests being made so awaiting them makes more sense,
- you are returning only promises for requests which were actually fired (while previously all incoming URLs were mapped to empty promises).

(Also despite the original recommendation I opted for `for..of` which tends to give better performance than `forEach`, but I don't insist on that.)

sarvaje:
Cool! Thanks for the PR. I also like the `for..of` more than `forEach`.

sarvaje (approved):

jnv:
Looks like the macOS CI got stuck? 🤔

sarvaje:
> Looks like the macOS CI got stuck?

Yep, it timed out, but for some reason, didn't report an error... It should be running now

PR #4091, opened 2020-10-30
Chore: Connector puppeteer use lock only in detached mode

### LINKED ISSUE #4089: [Feature] Allow connector puppeteer to disable writing to disk

Azure function consumption plan doesn't allow to write on any part of the disk, only in the tmp folder.

We have three options:
1. Modify the path where the connector save the files to the tmp folder.
2. Add a flag to NOT write in disk.
3. Add a flag to write in disk.

The writing is made to avoid lunching multiple browsers during the tests, so maybe options 3 is the better. 

@antross @molant thoughts?

sarvaje:
Fix #4089

## Pull request checklist

Make sure you:

Detach mode is used mostly for the test runner, to avoid opening multiple instances of the browser. To do so, it store in disk the browser information of the first instance opened, so the next calls can reuse that information.

Before this PR, the writing/reading of the disk were done even if it wasn't running in detach mode.

This PR disables the writings to disk if we are not in detach mode. This will save some time accessing to disk, but more important, it will solve a problem running `connector-puppeteer` in Azure Functions Consumption plan.

Azure Function Consumption plan doesn't allow writing in the disk, except for the tmp folder, so this change is needed if we want to complete https://github.com/webhintio/serverless-online-service/issues/143

### DIFF
--- a/packages/connector-puppeteer/src/lib/lifecycle.ts
+++ b/packages/connector-puppeteer/src/lib/lifecycle.ts
@@ -192,8 +192,6 @@ ${JSON.stringify(options, null, 2)}
 
 // TODO: Comments about the status
 export const launch = async (options: LifecycleLaunchOptions) => {
-    await lock();
-
     /**
      * Only try to connect to an existing browser when in detached mode,
      * otherwise the browser will be closed when one of the puppeteer
@@ -201,6 +199,7 @@ export const launch = async (options: LifecycleLaunchOptions) => {
      */
     /* istanbul ignore else */
     if (options.detached) {
+        await lock();
 
         const currentInfo = await getBrowserInfo();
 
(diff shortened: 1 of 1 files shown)


antross (commented):
The changes look fine, but can you add a bit more detail explaining why this won't work in the consumption plan without this change, and why using the lock isn't necessary in that context? Thanks!

sarvaje:
> The changes look fine, but can you add a bit more detail explaining why this won't work in the consumption plan without this change, and why using the lock isn't necessary in that context? Thanks!

Do you mean in the code or in the PR?

antross:
> > The changes look fine, but can you add a bit more detail explaining why this won't work in the consumption plan without this change, and why using the lock isn't necessary in that context? Thanks!
> 
> Do you mean in the code or in the PR?

I meant in the PR description. I figure we'll keep that around as part of the squashed commit comment too. Though now that you mention it, updating the existing code comment wouldn't hurt either 😉

sarvaje:
@antross please take another look.

antross (approved):
LGTM, thanks!

PR #4120, opened 2020-11-14
Fix: Check detach mode in close method

### LINKED ISSUE #4089: [Feature] Allow connector puppeteer to disable writing to disk

sarvaje:
Fix #4089

## Pull request checklist

Make sure you:

### DIFF
--- a/packages/connector-puppeteer/src/lib/lifecycle.ts
+++ b/packages/connector-puppeteer/src/lib/lifecycle.ts
@@ -247,7 +247,7 @@ export const launch = async (options: LifecycleLaunchOptions) => {
     return connection;
 };
 
-export const close = async (browser: puppeteer.Browser, page: puppeteer.Page) => {
+export const close = async (browser: puppeteer.Browser, page: puppeteer.Page, options: LifecycleLaunchOptions) => {
     debug(`Closing`);
 
     if (!browser) {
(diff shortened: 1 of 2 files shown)


packages/connector-puppeteer/src/lib/lifecycle.ts
```diff
@@ -256,7 +256,9 @@ export const close = async (browser: puppeteer.Browser, page: puppeteer.Page) =>
         return;
     }
 
-    await lock();
+    if (options && options.detached) {
```
molant, line 259: Wouldn't it be better to just lock when is really needed?
I think the way it is right now we will be delaying closing tabs for all of the even though it is not needed. Or is it for assurance and avoid racing conditions?

sarvaje: I don' t remember exactly why the lock here, but I think it is to avoid racing conditions.

antross (approved):

PR #4307, opened 2021-02-13
Docs: Fix npm short alias for install command

MauricioRobayo:
The `npm` short alias for the `install` command is `npm i`.


antross (approved):
Good catch, thanks @MauricioRobayo!

PR #4329, opened 2021-02-17
Build: Set 'importHelpers' to improve build size [WIP]

ruppysuppy:
## Pull request checklist

Make sure you:

Set 'importHelpers' to improve build size ([snyk-snapshot.json](https://github.com/ruppysuppy/hint/blob/f7a368411812e7ba2cc28fdb182379ae1f6a2c5d/packages/hint-no-vulnerable-javascript-libraries/src/snyk-snapshot.json) was modified by the build process)


ruppysuppy:
The build is failing due to the bundle size as well @antross

sarvaje:
> snyk-snapshot.json was modified by the build process

You don't need to update this file. you can restore it and remove it from the PR.

About the size, the reason to enable the option is to save size, but that is not happening hehehe.

Let see what @antross think about this hehehe.

molant:
Maybe we are not seeing any memory gains because we are targeting a modern JS version so there aren't that many things to polyfill. This is based on [this article](https://mariusschulz.com/blog/external-helpers-library-in-typescript), but we should probably look at the diff of the output with and without the flag.
If this is actually correct, we should probably make the hint smarter and check for the ES target.

ruppysuppy:
> we should probably make the hint smarter

@molant how should I get started on this?

molant:
We probably want to validate first that the size does indeed change depending on the ES version.
The easiest approach I can think of is to look at the [`target` property](https://www.typescriptlang.org/tsconfig#target) in the `tsconfig.json` file. Then maybe we can use different thresholds depending on that value?
* ES3-ES5 --> warning
* ES6-XXX --> hint

All of this is assuming the hypothesis is right!
@antross, @sarvaje what do you think?

antross:
I think @molant is on the right track here. 

I just tested this PR with our current build target of ES2017 and not much changed. I bumped that back to ES5 and was able to see a few places we used the tslib import, but even that didn't appear to affect a lot of the files I tested.

The ideal fix may need to be more nuanced around the actual features being used by a project (and number of files using those features) to determine the practical savings involved. At least from this test case it's looking like a lot of scenarios won't result in much savings so bringing in the entire tslib without tree-shaking can definitely result in a net size increase.

antross:
@webhintio/contributors until this check can be improved, do we still feel comfortable with having this warning or do folks feel we should disable it in the interim (I'm personally leaning toward the latter)?

molant:
Disable sgtm

hxlnt:
Per the conversation above, we're going to close this PR and pursue disabling the hint by default. See https://github.com/webhintio/hint/issues/4526.

PR #4429, opened 2021-03-24
Update issue templates

hxlnt:
This change:
 - simplifies issue templates to make it easier for users to file issues
 - migrates issue templates to GitHub's recommended new format
 - combines "request a feature" and "request new hint" templates

The browser extension bug report remains unchanged since the extension reporting system requires it, I think. We should still test it and make sure the change to the new GitHub issue template format didn't break anything.


antross:
Retriggering CI. Looks like our link validator failed on a bunch of links in files unrelated to this change. I think it was probably a temporary network issue and should be resolved on re-run.

antross:
No luck, @sarvaje any ideas as to why the link validator is suddenly failing on this PR?

antross (approved):
LGTM once we figure out what's going on with CI. Also, I confirmed the browser extension is just replacing everything between `---` and `---` with the empty string so the slight change in format within shouldn't be a problem.

antross:
Confirmed the markdown link validator is also having issues on `main`. We hadn't been seeing the issues because it was only running for *.md files editing outside of a sub-package. Going to merge this update and we'll work on fixing (or disabling) the validator separately.

PR #4872, opened 2021-11-02
New: Disable typescript-config/no-comments by default

### LINKED ISSUE #4839: [Bug] removeComments is inappropriate to suggest
# 🐞 Bug report

## Description

webhint currently has a suggestion that TypeScript users turn on `removeComments` in their project because this will help reduce payload size over the wire. This is inappropriate because users who care about this should either be using a minifier, or the comment emit doesn't matter enough in the first place. Ideally, TypeScript wouldn't even have supported `removeComments` if we did it all over again.

## Details

The TypeScript team received a suggestion about this behavior and it wasn't clear that this was a hint from webhint, not TypeScript.

https://github.com/microsoft/TypeScript/issues/46596

Our team doesn't believe this is an appropriate best practice to suggest though.

antross:
## Pull request checklist

Make sure you:

Disable the hint `typescript-config/no-comments` by default per
feedback from the TypeScript team in #4839. Also update docs
to indicate the hint is disabled.

Fix #4839

### DIFF
--- a/packages/configuration-development/index.json
+++ b/packages/configuration-development/index.json
@@ -29,7 +29,6 @@
         "typescript-config/consistent-casing",
         "typescript-config/import-helpers",
         "typescript-config/is-valid",
-        "typescript-config/no-comments",
         "typescript-config/strict",
         "typescript-config/target"
     ],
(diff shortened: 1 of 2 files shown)


antross:
@molant @sarvaje I suspect we'll also eventually want to remove this hint, but since that's likely a breaking change I think we should start with disabling it then follow-up with a full removal later. What do you think?

molant:
I think disabling is fine, not every body uses a bundler so I think there's still some value in it😊

sarvaje (approved):

PR #4882, opened 2021-11-09
Fix: Assume expected parents in template fragments

### LINKED ISSUE #4868: HTML fragments starting with <li> incorrectly report requiring a <ul> parent
Webhint has logic to handle HTML fragments such that expected ancestors should be assumed to be present in the final result, but that doesn't appear to be happening when `li` is the root node of a fragment. 

I can repro with just a basic HTML file containing an empty `li`:
```
<li></li>
```

When opening this in the VS Code extension, webhint reports:
> `<li> elements must be contained in a <ul> or <ol> webhint(axe/structure)`

_Originally posted by @antross in https://github.com/webhintio/hint/issues/4840#issuecomment-956414676_

antross:
## Pull request checklist

Make sure you:

Prevent false-positives when processing root elements like `<li>` in
template fragments by assuming and creating expected parent elements
so hints see the most likely fully-realized structure.

Fix #4868

### DIFF
--- a/packages/parser-jsx/src/parser.ts
+++ b/packages/parser-jsx/src/parser.ts
@@ -237,12 +237,8 @@ const allowsTextChildren = (node: JSXElement): boolean => {
  * provided roots derived from the specified resource.
  */
 const createHTMLFragment = (roots: RootMap, resource: string) => {
-    const dom = parse5.parse(
-        `<!doctype html><html data-webhint-fragment></html>`,
-        { treeAdapter: htmlparser2Adapter }
-    ) as DocumentData;
-
-    const body = (dom.children[1] as ElementData).children[1] as ElementData;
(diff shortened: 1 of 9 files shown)


sarvaje (approved):
LGTM but some tests are failing, so maybe this is introducing some issue.

antross:
@sarvaje looks like those are for `hint-doctype`. I originally set this up to go down the new fragment path for anything that doesn't have an `<html>` tag in source, but I may tweak that to also treat anything with `<!doctype html>` (or similar) in source as a full document instead of a fragment, even if no `<html>` tag is present. I think that'll match author intent for more of these cases.

antross:
@sarvaje the test failures were useful (related to expectations around `<!doctype>`) so I ended up making some changes to treat content with a `DOCTYPE` as a full document too. Also refactored the `isFragment` logic to be specified by the parser to ensure the answer is consistent.

sarvaje (approved):

PR #5027, opened 2022-02-04
Introducing config for the stale app

captainbrosset:
This configuration file can be used by the GitHub Stale app, which can help us maintain the repository by adding a label to issues that became inactive.


antross:
@captainbrosset I think this will be a good addition. Let me know if you need any help with the CLA so we can get this landed.

Issue #4715, opened 2021-08-29
[Bug] tsconfig "extends" error


RebeccaStevens:
# 🐞 Bug report

## Description

In vscode, whenever I set the `"extends"` property in my `tsconfig.json` file, webhint gives the following error message:

`Unexpected token } in JSON at position 705 (typescript-config/is-valid)`

## Details

I'm using the latest vscode extention.
I don't have any custom config (I don't have a `.hintrc` file).
I'm using TypeScript 4.4.2

antross:
Thanks for the report @RebeccaStevens!

I was able to hit the same error in a test project, but I'm not sure if it's for the same reason. In my case the config I was extending had a trailing comma at the end of a list of properties:

```ts
    "compilerOptions": {
        "target": "esnext",
    }
```

Removing the trailing comma fixed the issue.

```ts
    "compilerOptions": {
        "target": "esnext"
    }
```

However this is definitely a bug as it *should* work. Digging in it looks like the problem is that webhint is using a JSONC parser for the top-level tsconfig file, but is handing off to a regular `JSON.parse` for handling extended configs. The latter is much more restrictive on what's allowed (hence the error).

This should probably be fixed by updating [`loadJSONFile`](https://github.com/webhintio/hint/blob/main/packages/utils-fs/src/load-json-file.ts#L7) to use [`parseJSON`](https://github.com/webhintio/hint/blob/9253978337228a09a57b08eb6722c37d93dbb865/packages/utils-json/src/parse-json.ts#L167) (which uses JSONC under the covers) instead of `JSON.parse`.

antross:
@sarvaje just checking, do you see any issues with switching `loadJSONFile` to use the JSONC-backed `parseJSON` helper instead of `JSON.parse`?

sarvaje:
No, if we are already using `loadJSONFile` for the top-level tsconfig, I think it makes sense to use it for the extended configs too.

jamclema:
PR #5031 has made this change to the JSON parsing behavior and there should now be a useful error message when extending from an invalid JSON file.

PR #5031, opened 2022-02-10
Fix: Make JSON parsing behavior mode consistent

jamclema:
## Pull request checklist

Make sure you:

Switch the parser used in loadJSONFile to use jsonc parse instead of JSON.parse to match the behavior of parseJSON as discussed in #4715 to fix inconsistent behavior. parseJSON cannot be used directly as it would introduce a circular dependency.
This switch resolves the issue by making parsing more consistent.

Will also remove the need for #4742 by removing the dependency on strip-json-comments from the project.


packages/hint-babel-config/tests/is-valid.ts
```diff
@@ -70,15 +70,6 @@ const tests: HintLocalTest[] = [
             position: { match: '"../circular-2/.babelrc"' },
             severity: Severity.error
         }]
-    },
-    {
-        name: 'If .babelrc contains an invalid extends, it should fail',
-        path: path.join(__dirname, 'fixtures', 'invalid-extends'),
-        reports: [{
-            message: `Unexpected token ' in JSON at position 191`,
-            position: { match: '"../invalid/.babelrc"' },
```
antross, line 79: Can we keep this test by modifying the content of `packages/hint-babel-config/tests/fixtures/invalid/.babelrc` similarly to what you did for the `tsconfig.invalid.json` in this PR?

antross: Thanks!

PR #5046, opened 2022-02-17
Fix: Disable typescript-config/import-helpers by default

### LINKED ISSUE #4526: Disable ts-import-helpers by default
## Description

This can be disabled as it doesn't result in significant bundle size reduction in modern JS targets. See https://github.com/webhintio/hint/pull/4329

antross:
## Pull request checklist

Make sure you:

Testing has shown this guidance does not always result in
reduced bundle sizes. Keeping the hint so folks can opt-in
but disabling by default to avoid providing guidance that
does not produce the expected result.

Fix #4526

### DIFF
--- a/packages/configuration-development/index.json
+++ b/packages/configuration-development/index.json
@@ -27,7 +27,6 @@
         "scoped-svg-styles",
         "sri",
         "typescript-config/consistent-casing",
-        "typescript-config/import-helpers",
         "typescript-config/is-valid",
         "typescript-config/strict",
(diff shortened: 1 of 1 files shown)


antross:
@flynnolivia FYI, I'll plan to merge this without re-running the Windows CI leg since it just timed out (we've had problems with that in the past). Both Mac and Linux are passing so this should be good-to-go.

flynnolivia (approved):
LGTM!

PR #5087, opened 2022-03-25
Fix: Ensure development configuration is installed

antross:
The `hint` package treats `@hint/configuration-development` as an
optional dependency which means npm can choose to ignore it if there
are issues during install. However, when using the VS Code extension
this configuration is expected to exist, leading to errors in the
extension if it is not present.

Fixed by explicitly requesting the package for this configuration
in addition to `hint` when installing the shared webhint instance.

### DIFF
--- a/packages/extension-vscode/src/utils/webhint-packages.ts
+++ b/packages/extension-vscode/src/utils/webhint-packages.ts
@@ -6,7 +6,7 @@ const updateWebhintTimeout = 120000;
 
 /* istanbul ignore next */
 const installWebhint = (options: InstallOptions) => {
-    return installPackages(['hint@latest', 'typescript@latest'], options);
+    return installPackages(['@hint/configuration-development@latest', 'hint@latest', 'typescript@latest'], options);
 };
 
 /**


packages/extension-vscode/src/utils/webhint-packages.ts
```diff
@@ -6,7 +6,7 @@ const updateWebhintTimeout = 120000;
 
 /* istanbul ignore next */
 const installWebhint = (options: InstallOptions) => {
-    return installPackages(['hint@latest', 'typescript@latest'], options);
+    return installPackages(['@hint/configuration-development@latest', 'hint@latest', 'typescript@latest'], options);
```
molant, line 9: Isn't redundant to have `hint@latest` and `typescript@latest` if you are adding this configuration?
Wonder if this can cause issues if different versions are downloaded.

antross: It's redundant only if everything succeeds. Since it's an optional dependency we've seen some folks who didn't get `@hint/configuration-development` causing `hint` to fail on startup. 

My theory as to how they got into this state is that their network connection was interrupted during the install, but `npm` happily moved on without the package since it was optional. What's worse is this state appears to get cached so subsequent attempts to re-install don't end up restoring the missing optional dependency.

I was able to confirm on an affected machine that explicitly listing the configuration file for install "unsticks" the broken state (in addition to hopefully helping prevent it the state from happening again). I think it's still correct for this package to be optional relative to `hint` which won't necessarily use it in all contexts, but not the extension (plus if we wanted to make it non-optional we'd have to figure out how to break the circular dependency cycle that'd create in the release script).

antross: > Wonder if this can cause issues if different versions are downloaded.

Possibly, but we should be able to get out of that state on the next check for updates. In theory this would only happen if someone was installing while we were releasing a new version since we can't flip all new packages live simultaneously (and that could still cause problems even if we were only depending on `hint` since we don't currently control what order the packages go live in).

molant: I thought we had hint as a direct dependency of the development config and thus why I thought it was redundant having it.

After thinking a bit I believe we remove it to prevent dependency cycles 😅

seafoodmengthao (commented):
Meng Thao here what is reason for the issue here.

flynnolivia (approved):

PR #5131, opened 2022-04-21
Fix: Requirements for form-field-multiple-labels rule from axe-core

vidorteg:
## Pull request checklist

Make sure you:


packages/utils-dom/tests/axe.ts
```diff
@@ -71,6 +93,13 @@ test.serial('aria-hidden-focus', async (t) => {
     });
 });
 
+test.serial('form-field-multiple-labels', async (t) => {
+    await testAxe(t, {
+        incomplete: '<label for="test">Hi</label><label for="test">Foo</label><input type="text" id="test" />',
```
antross, line 98: Generally, an incomplete rule suggests a bug. Is there a reason we expect this case to be incomplete versus explicitly passing or failing?

packages/utils-dom/src/htmlelement.ts
```diff
@@ -260,6 +260,13 @@ export class HTMLElement extends Node {
         return parse5.serialize(this._element as htmlparser2Adapter.Node, { treeAdapter: htmlparser2Adapter });
     }
 
+    /**
+     * https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll
+     */
+    public querySelectorAll(selectors: string): HTMLElement[] {
+        return this.ownerDocument.querySelectorAll(selectors);
```
antross, line 267: We need to scope this query to only include results representing nodes under the element it was called on. To do that, I suspect we'll need to call `cssSelect` directly so we can specify the proper scope.

packages/utils-dom/src/htmlelement.ts
```diff
@@ -260,6 +260,13 @@ export class HTMLElement extends Node {
         return parse5.serialize(this._element as htmlparser2Adapter.Node, { treeAdapter: htmlparser2Adapter });
     }
 
+    /**
+     * https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll
+     */
+    public querySelectorAll(selectors: string): HTMLElement[] {
```
antross, line 266: While we're here, can we include `querySelector` as well for completeness?

antross (changes requested):

packages/utils-dom/src/htmlelement.ts
vidorteg: good catch, done.

packages/utils-dom/tests/axe.ts
vidorteg: As axe-core is reporting this as incomplete I created some logic to pass if the right 'incomplete' was reported (per offline discussion)

antross (approved):

PR #5161, opened 2022-05-11
New: quick-fixes for vscode-extension package

vidorteg:
## Pull request checklist

Make sure you:

New: Quick-fixes for vscode-extension package

This PR introduces quick-fixes in the webhint overlay inside visual studio.
<Placeholder for link to vscode-devtools-extension PR>


packages/extension-vscode/src/quickFixProvider.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    private quickFixActionsTemplates = new Map([
+        ['vscode-webhint/ignore-category-project', (category: string) => {
+            return `Ignore '${category}' in this project`;
+        }],
+        ['vscode-webhint/ignore-category-global', (category: string) => {
+            return `Ignore '${category}' in all projects`;
+        }],
+        ['vscode-webhint/ignore-hint-project', (name: string) => {
+            return `Ignore rule'${name}' in this project`;
```
antross, line 16:
```suggestion
            return `Ignore rule '${name}' in this project`;
```

packages/extension-vscode/src/quickFixProvider.ts
```diff
@@ -0,0 +1,60 @@
...
+        if (!textDocument) {
+            return null;
+        }
+
+        const results: CodeAction[] = [];
+
+        params.context.diagnostics.forEach((currentDiagnostic) => {
+            this.quickFixActionsTemplates.forEach((value, key) => {
+
+                // only respond to Microsoft Edge Tools requests.
```
antross, line 40:
```suggestion
                // only respond to requests for specified source.
```

packages/extension-vscode/src/server.ts
```diff
@@ -28,6 +55,12 @@ connection.onDidChangeWatchedFiles(async () => {
     }));
 });
 
+connection.onCodeAction(quickFixActionProvider.provideCodeActions.bind(quickFixActionProvider));
```
antross, line 58: After reading through this, I think we can move the CodeActionProvider to the extension.ts side which would allow us to reduce message passing and modifications to the language server (since we can also read the current list of diagnostics from that side).

See this vscode sample for an example of registering a CodeActionProvider from the extension host:
https://github.com/microsoft/vscode-extension-samples/blob/30c1740717acdc7e4aabc783d07bd83e21857320/code-actions-sample/src/extension.ts#L22

packages/extension-vscode/src/quickFixProvider.ts
vidorteg: missed this one, will address it in the next round.

packages/extension-vscode/src/quickFixProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    private quickFixActionsTemplates = new Map([
+        ['vscode-webhint/ignore-hint-project', (category: string) => {
+            return `Ignore hint '${category}' in this project`;
+        }],
+        ['vscode-webhint/ignore-hint-global', (category: string) => {
+            return `Ignore hint '${category}' in all projects`;
+        }],
+        ['vscode-webhint/ignore-problem-project', (name: string) => {
+            return `Ignore problem '${name}' in this project`;
```
antross, line 17: These will need to be scoped to only the `compat-api` hints (`compat-api/css` and `compat-api/html`). Currently it matches other hint messages so things like `css-prefix-order` end up with an ignore entry, but the generated `.hintrc` doesn't work (as that hint doesn't accept an ignore list).

packages/extension-vscode/src/quickFixProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+                // only respond to Microsoft Edge Tools requests.
+                if (!currentDiagnostic.source || currentDiagnostic.source !== this.sourceName) {
+                    return;
+                }
+
+                // default title
+                let title = currentDiagnostic.code?.toString() || '';
+
+                // if it is a problem (not a hint) use the problem as a title instead
+                if (key.includes('-problem-')) {
```
antross, line 50: These will need to be scoped to only the `compat-api` hints (`compat-api/css` and `compat-api/html`). Currently it matches other hint messages so things like `css-prefix-order` end up with an ignore entry, but the generated `.hintrc` doesn't work (as that hint doesn't accept an ignore list).

antross: This also leads to some interesting duplicate entries for our a11y hints:
image

packages/extension-vscode/src/quickFixProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+            });
+        });
+
+        const editCurrentProjectConfigTitle = 'Edit .hintrc for current project';
+        const editGlobalProjectConfigTitle = 'Edit .hintrc for all projects';
+        const editCurrentProjectConfig: Command = { command: 'vscode-webhint/edit-hintrc-project', title: editCurrentProjectConfigTitle };
+        const editGlobalProjectConfig: Command = { command: 'vscode-webhint/edit-hintrc-global', title: editGlobalProjectConfigTitle };
+
+        results.push(CodeAction.create(editCurrentProjectConfigTitle, editCurrentProjectConfig, CodeActionKind.QuickFix));
+        results.push(CodeAction.create(editGlobalProjectConfigTitle, editGlobalProjectConfig, CodeActionKind.QuickFix));
```
antross, line 73: The addition of these entries needs to be conditional based on whether we've actually reported a problem at a given location. Currently they cause the lightbulb to show up no matter where in the document I've put the cursor, even if there's no issue at that location:
image

packages/extension-vscode/src/server.ts
```diff
@@ -28,6 +56,12 @@ connection.onDidChangeWatchedFiles(async () => {
...
+
+connection.onExecuteCommand(() => {
+    return;
```
antross, line 62: Is this left-over code or does having this registration cause some sort of side-effect?

packages/extension-vscode/src/server.ts
antross: Revisiting since I've been looking at workspace edits. Apparently CodeActions from the language server can include an `edit` instead of a `command`. This could potentially allow us to specify the changes we want VS Code to apply to a file from the language server without having to add custom handling in the extension host.

I think either this or my above suggestion would help reduce the contract between the host and the language server.

packages/extension-vscode/src/server.ts
```diff
@@ -17,7 +21,31 @@ connection.onInitialize((params) => {
...
+    return resultObject;
+});
+
+connection.onNotification(notifications.reloadAllProjectsConfig, async () => {
+    const message = `Reloading .hintrc file from:\n ${globalStoragePath}`;
+
+    console.log(message);
+    analyzer.onConfigurationChanged();
+    await Promise.all(documents.all().map((doc) => {
+        return analyzer.validateTextDocument(doc, globalStoragePath);
```
antross, line 47: Passing `globalStoragePath` here forces loading webhint itself from the shared location (not just using the global config). We still want to use the project's version of webhint if one is present.

PR #5168, opened 2022-05-17
Fix: Remove reports for Via HTTP Header

### LINKED ISSUE #5159: Remove or reduce severity of reports for `Via` HTTP Header
Taking the liberty to comment on this issue, especially regarding the `Via` HTTP header. More often that not, it appears that this header is added by proxies, which makes it difficult or impossible to fix. For instance, [Google Cloud Load Balancing](https://cloud.google.com/load-balancing/docs/https) adds `Via: 1.1 google` to incoming requests and responses. Any CDN sitting behind GCLB, such as [Sanity](https://sanity.io)’s, will display this header. 

The article linked in the original message from this issue also mentions this. Although it does say it’s most likely safe to remove (which I’m not arguing here), it also states that proxies have to set it up, and most people don’t have control over their proxy.

> More debatable perhaps is Via, which is required ([by RFC7230](https://httpwg.org/specs/rfc7230.html#header.via)) to be added to the request by any proxy through which it passes to identify the proxy. This can be the proxy’s hostname, but is more likely to be a generic identifier like “vegur”, “varnish”, or “squid”. Removing (or not setting) this header on a request can cause [proxy forwarding loops](http://www.icir.org/vern/papers/cdn-loops.NDSS16.pdf). However, interestingly it is also copied into the response on the way back to the browser, and here it's just informational and no browsers do anything with it, so it’s reasonably safe to get rid of it if you want to.

I was wondering if you’d be considering dropping the `Via` warning.

_Originally posted by @KittyGiraudel in https://github.com/webhintio/hint/issues/1096#issuecomment-1119311751_

flynnolivia:
## Pull request checklist

Make sure you:

This PR removes the Via header from the list of disallowed headers in the `hint-no-disallowed-headers` package. The tests are updated to ensure that including this header does **not** produce any report. The list of disallowed headers is also updated in the README file.

Fixes #5159

### DIFF
--- a/packages/hint-no-disallowed-headers/src/hint.ts
+++ b/packages/hint-no-disallowed-headers/src/hint.ts
@@ -106,7 +106,6 @@ export default class NoDisallowedHeadersHint implements IHint {
             host: 'disallowedHostHeader',
             p3p: 'disallowedP3PHeader',
             pragma: 'disallowedPragmaHeader',
-            via: 'disallowedViaHeader',
             'x-frame-options': 'disallowedXFrameOptionsHeader'
         };
 
--- a/packages/hint-no-disallowed-headers/README.md
+++ b/packages/hint-no-disallowed-headers/README.md
@@ -39,7 +39,6 @@ HTTP headers:
 * `Pragma`
 * `Public-Key-Pins`
 * `Public-Key-Pins-Report-Only`
-* `Via`
 * `X-AspNet-Version`
 * `X-AspNetMvc-version`
 * `X-Frame-Options`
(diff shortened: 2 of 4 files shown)


packages/hint-no-disallowed-headers/src/hint.ts
```diff
@@ -106,7 +106,6 @@ export default class NoDisallowedHeadersHint implements IHint {
             host: 'disallowedHostHeader',
             p3p: 'disallowedP3PHeader',
             pragma: 'disallowedPragmaHeader',
-            via: 'disallowedViaHeader',
```
antross, line 109: We should also clean up the string from the corresponding `messages.json`: https://github.com/webhintio/hint/blob/5ff878455533a5f44b0652154f39760c1b81be4d/packages/hint-no-disallowed-headers/src/_locales/en/messages.json#L30

antross (changes requested):
Overall LGTM - just one extra piece of cleanup needed before this is ready to land.

antross (approved):

PR #5215, opened 2022-06-14
New: Add getAttributeLocation to elements

antross:
## Pull request checklist

Make sure you:

Supports finer-grained positions for hints which will enable
placing errors more accurately and facilitate auto-fixes.

Also add end positions and tests for getLocation.

### DIFF
--- a/packages/utils-dom/src/htmlelement.ts
+++ b/packages/utils-dom/src/htmlelement.ts
@@ -208,6 +208,26 @@ export class HTMLElement extends Node {
 
     /**
      * Non-standard.
+     * Zero-based location of the specified attribute in original source code.
+     *
+     * @param name The name of the attribute to get the location for.
+     */
+    public getAttributeLocation(name: string): ProblemLocation {
+        const elementLocation = this._getOriginalLocation();
+        const location = elementLocation?.attrs?.[name];
+
+        return {
+            column: location ? location.startCol : -1,
+            endColumn: location ? location.endCol : -1,
+            endLine: location ? location.endLine - 1 : -1,
+            endOffset: location ? location.endOffset : -1,
+            line: location ? location.startLine - 1 : -1,
+            startOffset: location ? location.startOffset : -1
+        };
+    }
+
+    /**
+     * Non-standard.
      * Zero-based location of the element in original source code.
      */
     public getLocation(): ProblemLocation {
@@ -217,6 +237,8 @@ export class HTMLElement extends Node {
         return {
             column: location ? location.startCol : -1,
             elementId: this._element.id,
+            endColumn: location ? location.endCol : -1,
+            endLine: location ? location.endLine - 1 : -1,
             endOffset: location ? location.endOffset : -1,
             line: location ? location.startLine - 1 : -1,
             startOffset: location ? location.startOffset : -1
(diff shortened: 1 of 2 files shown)


packages/utils-dom/src/htmlelement.ts
```diff
@@ -206,6 +206,26 @@ export class HTMLElement extends Node {
         return null;
     }
 
+    /**
+     * Non-standard.
+     * Zero-based location of the specified attribute in original source code.
+     *
+     * @param name The name of the attribute to get the location for.
+     */
+    public getAttributeLocation(name: string): ProblemLocation {
```
hkal, line 215: What is an attribute in this context? An element's attribute?

antross: Yes, an HTML attribute on an element.

antross:
I'm not yet sure if I want to include it in this PR or a separate follow-up, but we'll likely want to enhance `context.report` to allow specifying an attribute name in addition to an element reference. Then we can generalize extracting the attribute location on behalf of the hint in the same way we currently do for elements.

hkal:
> I'm not yet sure if I want to include it in this PR or a separate follow-up, but we'll likely want to enhance `context.report` to allow specifying an attribute name in addition to an element reference. Then we can generalize extracting the attribute location on behalf of the hint in the same way we currently do for elements.

I think it's fine to include the `context.report` work in this PR - seems appropriately scoped.

antross:
> I think it's fine to include the `context.report` work in this PR - seems appropriately scoped.

I did both that and updated `hint-no-protocol-relative-urls` so we have a concrete example of this end-to-end.

packages/utils-dom/src/htmlelement.ts
```diff
@@ -217,6 +237,8 @@ export class HTMLElement extends Node {
         return {
             column: location ? location.startCol : -1,
             elementId: this._element.id,
+            endColumn: location ? location.endCol - 1 : -1,
+            endLine: location ? location.endLine - 1 : -1,
```
antross, line 241: Looks like I can't add an end position to elements without unintended side-effects. Here's an example of a report on the `<html>` element. With this change we now underline the entirety of the tag and all its contents:
image

The reason - our VS Code extension looks for an omitted `endColumn` before applying logic to underline only the tag name:
https://github.com/webhintio/hint/blob/8162e2a7a04d31bb7e79e09b1e0b9a2b9365148a/packages/extension-vscode/src/utils/problems.ts#L33

This seems reasonable for underlining, but different than what we want if we need to apply an edit to replace the entire element. I may need to introduce a separate method to get the full tag range and keep this one as just the position of the tag name.

antross: After further review, I realized this is really only an issue for derived locations within `context.report`. So instead of splitting into separate APIs, I kept the change to include the full range in `getLocation` and `getAttributeLocation`, then updated `context.report` to use only the start of the location when extracting from a provided element/attribute.

This keeps the underlying locations more consistent with how we're using them elsewhere. For example, the `endOffset` was already present and used by `hint-meta-charset-utf8` and CSS-related locations already included `endLine` and `endColumn`.

The end result (one issue on the `img` for a missing title and another on `src` for the protocol-relative-url):
image

hkal (approved):
LGTM

mliao95 (approved):
LGTM

PR #5220, opened 2022-06-22
[Feature]: Adding code fix for no-protocol-relative-urls hint + Adding test infra for code fixes

### LINKED ISSUE #5214: [Feature] Create code action for "No Protocol-Relative URLs" hint
See documentation on hint here: https://webhint.io/docs/user-guide/hints/hint-no-protocol-relative-urls/

Triggers + Code Action solutions:
- File contains protocol-relative URLs like `"//example1.com/style.css"`
    - Code Action: Prefix all instances of `//` with `https://`

mliao95:
## Pull request checklist

Make sure you:

Closes #5214

### DIFF
--- a/packages/hint-no-protocol-relative-urls/src/hint.ts
+++ b/packages/hint-no-protocol-relative-urls/src/hint.ts
@@ -54,6 +54,22 @@ export default class NoProtocolRelativeUrlsHint implements IHint {
                 debug('Protocol relative URL found');
 
                 const message = getMessage('noProtocolRelativeUrl', context.language);
+                const tagString = src ? 'src' : 'href';
+                const attributePosition = element.getAttributeLocation(tagString);
+                const fixedUrl = url.replace('//', 'https://');
+                const replacementText = `${tagString}="${fixedUrl}"`;
+
+                const codeFix = [
+                    {
+                        location: {
+                            column: attributePosition.column,
+                            endColumn: attributePosition.endColumn,
+                            endLine: attributePosition.endLine,
+                            line: attributePosition.line
+                        },
+                        text: replacementText
+                    }
+                ];
 
                 const severity = isHTTPS(resource) ?
                     Severity.hint :
@@ -63,9 +79,10 @@ export default class NoProtocolRelativeUrlsHint implements IHint {
                     resource,
                     message,
                     {
-                        attribute: src ? 'src' : 'href',
+                        attribute: tagString,
                         content: url,
                         element,
+                        fixes: codeFix,
                         severity
                     });
             }
--- a/packages/hint-no-protocol-relative-urls/src/hint.ts
+++ b/packages/hint-no-protocol-relative-urls/src/hint.ts
@@ -54,19 +54,14 @@ export default class NoProtocolRelativeUrlsHint implements IHint {
                 debug('Protocol relative URL found');
 
                 const message = getMessage('noProtocolRelativeUrl', context.language);
-                const tagString = src ? 'src' : 'href';
-                const attributePosition = element.getAttributeLocation(tagString);
+                const attribute = src ? 'src' : 'href';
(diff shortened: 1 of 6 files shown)


packages/hint-no-protocol-relative-urls/src/hint.ts
```diff
@@ -54,6 +54,17 @@ export default class NoProtocolRelativeUrlsHint implements IHint {
                 debug('Protocol relative URL found');
 
                 const message = getMessage('noProtocolRelativeUrl', context.language);
+                const attribute = src ? 'src' : 'href';
+                const attributeLocation = element.getAttributeLocation(attribute);
+                const fixedUrl = url.replace('//', 'https://');
+                const replacementText = `${attribute}="${fixedUrl}"`;
+
+                const codeFix = [
```
antross, line 62: @mliao95 want to take a stab at adding a test to the suite for this hint to verify the proposed fix matches what we expect? 

At a minimum we can just compare the returned location and text to hard-coded values, but even better if we can enhance our test helpers to apply the replacement so we can verify the resulting text matches our expectations instead.

packages/utils-tests-helpers/src/hint-runner.ts
```diff
@@ -79,6 +79,57 @@ const findPosition = (source: string, position: MatchProblemLocation): ProblemLo
...
+
+    if (!endLine || !endColumn) {
+        return [-1, -1];
+    }
+
+    const regex = /(\r\n|\r|\n)/gm;
+
+    for (let i = 0; i < endLine; i++) {
+        // We want to add 1 to the resulting index to account for the newline char. If the resulting index is undefined, we set curLineOffset to 0.
+        const curLineOffset = (regex.exec(document)?.index || -1) + 1;
```
antross, line 119: Per your comment, I suspect this only works with `\n` or `\r`, but not `\r\n`. Probably better to use `lastIndex` instead which automatically places you at the start of the next match so you don't have to manually account for which type of newline was used.

```suggestion
        const curLineOffset = regex.exec(document)?.index || -1;
```

Might also be worth exporting this function and writing a couple of basic unit tests for it to ensure we handle each newline case correctly.

packages/utils-tests-helpers/src/hint-runner.ts
```diff
@@ -330,7 +381,77 @@ const validateResults = (t: ExecutionContext<HintRunnerContext>, sources: Map<st
             return;
         }
 
-        const filteredBySeverity = filteredByPosition.filter((report) => {
+        const filteredByFixes = filteredByPosition.filter((report) => {
+            /*
+             * If the report from the test doesn't ask for fixes,
+             * we don't need to macth it.
```
antross, line 387:
```suggestion
             * we don't need to match it.
```

antross (changes requested):
Looks good overall and works great with the current test cases. I found one potential bug we should consider addressing, but after that this should be ready to merge.

packages/utils-tests-helpers/tests/hint-runner-helpers.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+test('positionToOffset - properly translates offset of file with LF line endings', (t) => {
+    const document = 'one\ntwo\nthree\nfour';
+
+    const actual = positionToOffset({column: 1, endColumn: 3, endLine: 2, line: 2}, document);
+    const expected = [9, 11];
+
+    t.true(actual[0] === expected[0] && actual[1] === expected[1], `Offset differs from expected.
+                Actual: [${actual[0]}, ${actual[1]}], Expected: [${expected[0]}, ${expected[1]}]`);
```
antross, line 18: Not blocking, but `t.deepEqual(actual, expected)` would also work here and provide the diff output for you.

PR #4886, opened 2021-11-15
Breaking: Raise min required node version to 14

antross:
## Pull request checklist

Make sure you:

- [x] Signed the Contributor License Agreement (after creating PR)
- [x] Followed the [commit message guidelines](https://webhint.io/docs/contributor-guide/getting-started/pull-requests/#commit-messages)

For non-trivial changes, please make sure you also:

- [ ] Added/Updated related documentation.
- [ ] Added/Updated related tests.

## Short description of the change(s)

Align specified minimum node version with that required by recent dependency updates.

PR #5243, opened 2022-07-01
Breaking : Remove 'request' dependency

vidorteg:
## Pull request checklist

Make sure you:

This PR removes 'request' as a (direct) dependencies for the packages/* folder, this is due
it being deprecated. node-fetch is used as a substitution. 

How to understand this change:
- `Direct dependency` changes are those who are directly affected by request removal.
- `Indirect dependency` changes are those who are affected by `Direct dependency` changes.
- `Other dependency` changes are those who are affected in any other way.
 
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>
</head>

<body link="#0563C1" vlink="#954F72">

**Change Type** | **Package name** | Impact on package | Notes
-- | -- | -- | --
Direct-dependency | `download-dist` | Medium | Changes due to being dependent on  a `request` type as a method signature
Direct-dependency | `Hint-html-checker` | Small |  Updating  caller to match new signature |
Direct-dependency | `Hint-no-broken-links` | Medium | Some tests were using https and   other http. As they share a default configuration they all need to be the   same. Previously ‘request’ allowed this mixing, but node-fetch requires that   all connections should share protocol.
Direct-dependency | `utils-connector-tools` | Large | Core change, I tried to keep as much of the previous logic but a lot   changed mostly due: <br /><br />1. Handling https: request   previously handled it internally but now we need to do it ourselves with the   help of another package `https`  <br /> 2. Request was used previously to   strong type most of the packages, so removing request was a breaking change   for a lot of dependencies. To minimize impact in the future I Introduced a   new interface `IRequestOptions`
Direct-dependency | `utils-network` | Large | Had to be fully refactored,   please see `utils-connector-tools` for more details
Indirect-Dependency | `hint-sri` | Small | changes in headers
Indirect-Dependency | `Connector-jsdom` | Small | Removal of unsupported options
Indirect-Dependency | `Connector-puppeteer` | Small | Removal of unsupported options
Indirect-Dependency | `utils-tests-helpers` | Small | Changes on a method signature
Other | `hint-https-only` | Large | This hint has to be refactored   to address the two facts: <br /> 1.- that `request` used to work   with a callback and node-fetch does not. Previous behavior completed the   callback and then threw the error, while the new behavior uses try {} and   catch{}. <br />2. `request` used to handle   https internally but now we need to handle it. To keep most of the logic in   place I had to change how exceptions were handled.

</body>
</html>


packages/utils-connector-tools/src/requester.ts
```diff
@@ -51,33 +52,29 @@ const identity = (buff: Buffer): Promise<Buffer> => {
...
+const defaults: RequestInit = {
+    compress: false,
     headers: {
         'Accept-Encoding': 'gzip, deflate, br',
         'Accept-Language': 'en-US,en;q=0.8,es;q=0.6,fr;q=0.4',
         'Cache-Control': 'no-cache',
-        DNT: 1,
+        DNT: '1',
         Pragma: 'no-cache',
         'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.131 Safari/537.36'
```
molant, line 63: Maybe a good opportunity to update the UA? It's almost 30 versions behind.
Maybe this should be updated as part of the release although if it's worked this far maybe a `good first issue`?

vidorteg: good call, will do.

vidorteg: I gave a sewcond thought to this one, as we discussed releasing this change as a single (and easily reversible change) I think is better if I update the UA in a different PR, so that, if any reason we need to revert this one that change does not also get reverted.

scripts/dist-management/download-dist.js
```diff
@@ -1,7 +1,7 @@
 const path = require('path');
 const fs = require('fs');
 
-const request = require('request');
+const fetch = require('node-fetch');
```
vidorteg, line 4: @molant do you know what this script is used for?

scripts/update-3rd-party.ts
```diff
@@ -38,13 +34,15 @@ const cleanSchemaObject = (obj: any) => {
...
-                throw new Error(res.body.message);
-            }
-
-            const refJson = JSON.parse(res.body);
+            /**
+             * [vidorteg] validate if this is needed
+             * if (res.body.message) {
+             *      throw new Error(res.body.message);
+             * }
+             */
```
vidorteg, line 44: @anton do you know what this script is used for or how to test this change?

johnemau:
Just a thought here:

Node v18.6.0 has an [experimental _undocumented_ `fetch()` API][1]. Which could be an alternative to adding a new dependency on `node-fetch`.

To use we should set `engines.node` to `>= 18.6.0` in all the `package.json`s and check if there are any compatibility issues with other NodeJS API changes. 

[1]: https://github.com/nodejs/node/commit/4944ad0b9ed93354f3a6c7fd063ff204e016f75f#diff-7349c17bce4d97526eefcb6f8291d617e7edb92ed9927fd3b9264e8014fc4e7c

yarn.lock
```diff
@@ -8711,7 +8716,7 @@ release-zalgo@^1.0.0:
   dependencies:
     es6-error "^4.0.1"
 
-request@2.88.2, request@^2.88.2:
+request@2.88.2:
```
vidorteg, line 8719: This is an indirect dev dependency, so it will not block any new npmjs packages as its code is not released. It is used by `web-ext` which is used to build firefox extension. Mozilla team has plans to remove this dependency too:

https://github.com/mozilla/web-ext/issues/2130#issuecomment-1168408414

johnemau:
In future to make this kind of dependency rollouts easier to test we can do it in phases:

1. Add the new alternative dependency (use it in 1 package), [add a deprecation warning to old dependency usages][1].
2. Incrementally move all other packages to the new dependency.
3. Remove old dependency.

[1]: https://www.geeksforgeeks.org/node-js-util-deprecate-method/

vidorteg:
> Just a thought here:
> 
> Node v18.6.0 has an [experimental _undocumented_ `fetch()` API](https://github.com/nodejs/node/commit/4944ad0b9ed93354f3a6c7fd063ff204e016f75f#diff-7349c17bce4d97526eefcb6f8291d617e7edb92ed9927fd3b9264e8014fc4e7c). Which could be an alternative to adding a new dependency on `node-fetch`.
> 
> To use we should set `engines.node` to `>= 18.6.0` in all the `package.json`s and check if there are any compatibility issues with other NodeJS API changes.



> Just a thought here:
> 
> Node v18.6.0 has an [experimental _undocumented_ `fetch()` API](https://github.com/nodejs/node/commit/4944ad0b9ed93354f3a6c7fd063ff204e016f75f#diff-7349c17bce4d97526eefcb6f8291d617e7edb92ed9927fd3b9264e8014fc4e7c). Which could be an alternative to adding a new dependency on `node-fetch`.
> 
> To use we should set `engines.node` to `>= 18.6.0` in all the `package.json`s and check if there are any compatibility issues with other NodeJS API changes.

Great catch, we discussed this during the initial talks, at that point we chose that it was best to favor the current schedule (specially given it is still experimental). You can find some context on the discussion and the full schedule here:

https://github.com/webhintio/hint/pull/4886

vidorteg:
> In future to make this kind of dependency rollouts easier to test we can do it in phases:
> 
> 1. Add the new alternative dependency (use it in 1 package), [add a deprecation warning to old dependency usages](https://www.geeksforgeeks.org/node-js-util-deprecate-method/).
> 2. Incrementally move all other packages to the new dependency.
> 3. Remove old dependency.

Another good point, it was also brought up during the last webhint sync. TL;DR as the change is risky and the dependency chain is handled by the scripts, we chose to do it atomically as it is easier to revert (if needed). More details in the meeting notes:

https://docs.google.com/document/d/12c_1bkQZvf5zTjVxyATWLgfIYjUUembtdEhTMr4gCoI/edit#heading=h.59cpjbaabqwe

scripts/dist-management/download-dist.js
molant: IIRC, this is the script used to download the generated build artifacts when doing a fresh cloned. This should speed up things under certain circumstances, especially when cloning the project.

scripts/update-3rd-party.ts
molant: I believe this is for when we are pulling schema definitions from different places (can't remember exactly which ones).
This seems to be a safeguard for when the server is down or similar so I'd keep it because it has probably happened in the past.

molant:
@vidorteg do you think you can get rid of `got` in [this hint](https://github.com/webhintio/hint/blob/6dac6c694ba2fd575eb2177349beaae4311600d3/packages/hint-ssllabs/src/api.ts#L3) in this PR as well?

Can't remember why it wasn't using `request` there, but that way everything should be using the same thing and get rid of #5260

vidorteg:
> @vidorteg do you think you can get rid of `got` in [this hint](https://github.com/webhintio/hint/blob/6dac6c694ba2fd575eb2177349beaae4311600d3/packages/hint-ssllabs/src/api.ts#L3) in this PR as well?
> 
> Can't remember why it wasn't using `request` there, but that way everything should be using the same thing and get rid of #5260

I prefer to address that one in a different PR. This PR is already big enough so I prefer to keep it scoped to remove `request`  only ATM, I'll follow up with a PR to remove got once this get's checked in 😃, thoughts?

hkal (approved):
LGTM

PR #5286, opened 2022-08-30
Fix: Axe-core throws errors with local iframes

### LINKED ISSUE #5285: [Bug] allowedOrigins value "null" is not a valid origin
# 🐞 Bug report

## Description

When opening a local html with an iframe that has the default CSP webhint throws an error:
`allowedOrigins value "null" is not a valid origin`

image

## Details

- Create a file `test.html` with the following content:
```
<!DOCTYPE html>
<html lang='en'>
<body>
    <div>
        <div>
            <span>iframe: default-src CSP</span>
            <iframe marginwidth="0"
            title="iframe: default-src CSP"
            marginheight="0"
            hspace="0"
            vspace="0"
            frameborder="10"
            scrolling="no" src="http://bing.com"></iframe>
        </div>
    </div>
</body>
</html>
```
- Open webhint and open `test.html`

vidorteg:
## Pull request checklist

Make sure you:

Fixes an issue where axe-core throws:
`allowedOrigins value "null" is not a valid origin`

This is due to a axe-core throwing an error instead of finishing the analysis, this is currently an open bug in dequelabs:
`dequelabs/axe-core`
`issue: 3002`

This only manifests under the following conditions:
- local `html` files only
- file has iframes
- the iframes have the `Default` CSP
- The exception gets bubbled and eventually crashes `webhint`. To avoid this the proposed fix is to not scan `iframes` for local `html` files.
-----
Fix #5285

### DIFF
--- a/packages/hint-axe/src/util/axe.ts
+++ b/packages/hint-axe/src/util/axe.ts
@@ -145,12 +145,8 @@ const evaluateAxe = async (context: HintContext, event: CanEvaluateScript, rules
      * This is caused by an axe-core bug which is currently tracked here:
      * https://github.com/dequelabs/axe-core/issues/3002
      */
-    let shouldScanIframes = true;
     const uri = getAsUri(resource);
-
-    if (uri && uri.protocol.includes('file')) {
-        shouldScanIframes = false;
-    }
+    const shouldScanIframes = !(uri && uri.protocol.includes('file'));
 
     /* istanbul ignore next */
     try {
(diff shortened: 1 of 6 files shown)


flynnolivia (approved):

packages/hint-axe/src/util/axe.ts
```diff
@@ -137,6 +138,20 @@ const withQuotes = (ruleId: string) => {
...
+     * 'allowedOrigins value "null" is not a valid origin'
+     *
+     * This is caused by an axe-core bug which is currently tracked here:
+     * https://github.com/dequelabs/axe-core/issues/3002
+     */
+    let shouldScanIframes = true;
+    const uri = getAsUri(resource);
+
+    if (uri && uri.protocol.includes('file')) {
+        shouldScanIframes = false;
```
hkal, line 152: nit: The following would allow us to have `shouldScanIframes` as a const:
```
const uri = getAsUri(resource);
const shouldScanIframe = !(uri && uri.protocol.includes('file');
```

hkal (approved):
Small nit, but otherwise looks good.

PR #5615, opened 2023-06-21
Fix: Upgrade of utils-network to 1.0.25 breaks hint

berkon:
util-network was upgraded to a new major version of node-fetch which is now an ESM module. This breaks hint.

## Pull request checklist

Make sure you:


vidorteg:
thank you for putting this PR out definitely on the right direction, I added some missing parts in here:
https://github.com/webhintio/hint/pull/5618

Usually I would just made some comments on the original PR (this) but with so many people hitting this I had to create a new one so that I could release a new version with the fix ASAP. Thanks again for the contribution.

PR #5718, opened 2023-11-08
stale-while-revalidate and stale-if-error caching directives are valid with value

Psychpsyo:
## Pull request checklist

Make sure you:

``parseCacheControlHeader()`` now treats ``stale-while-revalidate`` and ``stale-if-error`` directives with value as valid directives.
They should be, according to [rfc5861](https://datatracker.ietf.org/doc/html/rfc5861).


vidorteg (approved):
Thank you for the contribution! LGTM

